PR #6, opened 2023-06-11
chore: add comments to code

huntabyte:
Some of these might be too verbose for your liking, so no offense will be taken if you remove some bits!


TGlide (commented):
This is great!!! Thank you 😄

PR #7, opened 2023-06-11
feat: popover

huntabyte:
WIP - adding this as a draft so @TGlide can see what's going on


TGlide:
Thanks for this!

Noticed something I assume is unexpected:
https://github.com/melt-ui/melt-ui/assets/26071571/348eba11-b390-46ce-b66f-90cbe7426f64

src/lib/builders/popover/index.ts
```diff
@@ -0,0 +1,147 @@
...
+		placement: 'bottom',
+	},
+	arrowSize: 8,
+	open: false,
+} satisfies CreatePopoverArgs;
+
+export function createPopover(args?: CreatePopoverArgs) {
+	const options = { ...defaults, ...args } as CreatePopoverArgs;
+	const positioning = writable(options.positioning);
+	const arrowSize = writable(options.arrowSize);
```
TGlide, line 34: Unless we want to return these so that they are modifiable post-creation, they don't need to be writables.

The thing is, so far for controlled creations, I've been using reactive statements:
image

But tbh, that runs the whole thing again, and may break something. So maybe we should return the options as stores? Maybe even a single `options` store. This is a pattern I have yet to define tbh.

src/lib/builders/popover/index.ts
```diff
@@ -0,0 +1,147 @@
...
+			return {
+				role: 'button' as const,
+				'aria-haspopup': 'dialog' as const,
+				'aria-expanded': $open,
+				'data-state': $open ? 'open' : 'closed',
+			};
+		};
+	});
+
+	let popoverCleanup: ((() => void) | undefined)[] = [];
```
TGlide, line 66: This is a clever idea, but I think it shows a shortcoming of `elementDerived`. What if we added an addUnsubscriber method to it?

TGlide: I took the liberty of doing that 😄

TGlide (commented):
Other notes:
- I think we can move actions to `internal/actions`. Also, quite nice that you just applied the action to the element ref! Never thought of that.
- Would be nice to have the popover close on esc!
- We should add a portal action as well probably.

huntabyte:
> Noticed something I assume is unexpected: [melt-ui/melt-ui/assets/26071571/348eba11-b390-46ce-b66f-90cbe7426f64](https://github.com/melt-ui/melt-ui/assets/26071571/348eba11-b390-46ce-b66f-90cbe7426f64)

This is actually expected and I believe we should keep it unless we want to provide an option for opting out somehow. This is accomplished using [autoUpdate](https://floating-ui.com/docs/autoUpdate).

The way I think about popovers is that if it's open, you expect it to be visible, regardless of the screen size or viewport, right? This lets the popover adapt to the screen on scrolling or resizing.

It's also how Radix handles it which I feel is like the standard for good UI design practices at this point.

Let me know your thoughts on that and if we want to provide the ability to "opt-out", I can look into adding that later on once the core functionality is complete :)

TGlide:
> > Noticed something I assume is unexpected: [melt-ui/melt-ui/assets/26071571/348eba11-b390-46ce-b66f-90cbe7426f64](https://github.com/melt-ui/melt-ui/assets/26071571/348eba11-b390-46ce-b66f-90cbe7426f64)
> 
> This is actually expected and I believe we should keep it unless we want to provide an option for opting out somehow. This is accomplished using [autoUpdate](https://floating-ui.com/docs/autoUpdate).
> 
> The way I think about popovers is that if it's open, you expect it to be visible, regardless of the screen size or viewport, right? This lets the popover adapt to the screen on scrolling or resizing.
> 
> It's also how Radix handles it which I feel is like the standard for good UI design practices at this point.
> 
> Let me know your thoughts on that and if we want to provide the ability to "opt-out", I can look into adding that later on once the core functionality is complete :)
> 
> 

Oh, the autoupdate is great! I just think the positioning is off when it goes to the top 😁

src/lib/builders/popover/index.ts
huntabyte: I was going to say that's something that I felt was missing that may end up being useful down the line! Awesome!

huntabyte:
@TGlide Oh whoops! Just noticed that, yeah definitely not intentional/expected and will resolve :)

huntabyte:
Still have some things to touch up on - especially regarding the docs and such, but I'll take this out of draft incase you want to merge the construction state.

PR #10, opened 2023-06-11
add keyboard constant object

huntabyte:
I added a `kbd` object which maps commonly used keyboard keys to their corresponding string values.

This prevents us from hard-coding strings all over the place & also guarantees typesafety/accuracy considering `KeyboardEvent.key` is just typed as a `string`.


TGlide (commented):
Great idea!

PR #14, opened 2023-06-12
Adjust isBrowser detection

pawelblaszczyk5:
This PR adjusts isBrowser detection to not rely on the window variable, which can be declared in various environments, such as e.g. Deno. 

Prior art: https://github.com/TanStack/virtual/pull/516 https://github.com/framer/motion/pull/1522 https://remix.run/docs/en/1.16.1/pages/gotchas#typeof-window-checks

I've also updated the packageManager field in package.json. It looks like the pnpm-lock was generated using newer version than this field suggested. It didn't work properly using corepack for automatically managing pnpm version

### DIFF
--- a/.changeset/forty-hounds-beg.md
+++ b/.changeset/forty-hounds-beg.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+Adjust browser detection function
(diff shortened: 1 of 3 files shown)


TGlide (approved):
Nicely noted!

PR #16, opened 2023-06-12
fix: select preview

huntabyte:
Having the arrow on the preview by default for a select feels odd. We can definitely keep it as an option for those who would want that and can possibly add an example of using the arrow within the `Select` docs!

I updated the styles of the preview as well to get rid of the bug and also position the popper appropriately for a select. I think it's best to avoid using `translate` css properties to move it around, and instead rely on the `positioning` options.

I'm not sure whether they would conflict with one another, but the CSS property definitely overrides any default positioning props. This is something we can investigate further.


TGlide:
This is great. Definitely agree on the arrow!

PR #18, opened 2023-06-13
feat(select): updates & touchups

huntabyte:
Things complete with this PR:
- [x] Update select preview code snippet
- [x] Add ChevronDown Icon to preview
- [x] Move longer tailwind classes to `<style>` block for clarity
- [x] Add `disabled` and `required` props to builder function
- [x] Add additional aria & data attributes 

Some additional thoughts:

### Accessibility 
Some things I'm starting to realize is that if we want to truly provide an "out of the box" accessible component, with little thought from the user required to do so, we're going to have to take some more props and possibly add some additional elements to the builder.

For example, with a select, the "dropdown icon" should never be part of the accessibility tree, so by providing some default `icon` attributes that can be applied, more specifically `aria-hidden`. Of course, the beauty with builders is that a lot of things can be "opt-out"/"opt-in", by simply not using/spreading the element, but I think it would be nice to provide it.

### Typeahead Support
If we want to align with [WAI-ARIA guidlines](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/) for listbox, then we'll need typeahead support. I've thought of a few different ways we might approach this:

#### Option A: Using the values
Since the user has to pass a value into the `...$option`, we could have a builder-root level `values` store which would be an array of values and use that to search against. 

The issue with this is that sometimes non user-friendly values are used, think something like languages, which might use the locale ISO code as the value, but display the language's name for the label. So this is probably not an ideal approach.

#### Option B:
Forcing the user to pass a `label`/`name` into the option like so `...$option({ value: "pt", label: "Portuguese"})`

#### Option C:
Instead of forcing the user to pass in a label/name, if they provide one, we use that for typeahead, if not, we use the value.

### DIFF
--- a/src/lib/builders/select/index.ts
+++ b/src/lib/builders/select/index.ts
@@ -26,10 +26,14 @@ import { derived, writable } from 'svelte/store';
 type CreateSelectArgs = {
 	positioning?: FloatingConfig;
 	arrowSize?: number;
+	required?: boolean;
+	disabled?: boolean;
 };
 
 const defaults = {
 	arrowSize: 8,
+	required: false,
+	disabled: false,
 	positioning: {
 		placement: 'bottom',
 		sameWidth: true,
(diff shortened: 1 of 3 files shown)


src/lib/builders/select/index.ts
```diff
@@ -96,6 +77,38 @@ export function createSelect(args?: CreateSelectArgs) {
 		}
 	);
 
+	const trigger = elementMultiDerived(
```
TGlide, line 80: Now that I think about it, does it make sense for trigger to be multi derived? I'd say no, right?

TGlide:
For type-ahead, we can also read the trigger's innerText 👀

PR #35, opened 2023-06-15
selectedText should hold the correct value and support numbers

geerew:
This update is 2 fold:
1. Update selectedText store to take in a number
2. Option now sets 2 additional data-* attributes (value and type) which is used when setting selectedText
    - Value holds the option value
    - Type holds the value type (string, number

Testing:
- Passed in an array containing both strings and numbers
- Asserted type-ahead works with numbers
- Mornitored the reactivitey of selectedText to ensure it's correct

### DIFF
--- a/.changeset/brave-mirrors-appear.md
+++ b/.changeset/brave-mirrors-appear.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+refactor: update selectedText to support numbers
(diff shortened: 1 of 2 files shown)


TGlide (approved):
Thank you for the PR!! LGTM 😄

huntabyte:
I'm thinking that the `value` should be of type `any` to align with Svelte's element types for the `HTMLSelectElement` / `HTMLInputElement`. I assume a common practice will be binding the selected value to one of those native elements and when data is coming from a database which could possible be `null` / `undefined` we may run into some issues.

TGlide:
> I'm thinking that the `value` should be of type `any` to align with Svelte's element types for the `HTMLSelectElement` / `HTMLInputElement`. I assume a common practice will be binding the selected value to one of those native elements and when data is coming from a database which could possible be `null` / `undefined` we may run into some issues.

Sorry, merged before I saw it! that makes sense.... Maybe we can use JSON.stringify and JSON.parse to handle it.

PR #41, opened 2023-06-15
Remove md:max-h-none class from select menu example

Manuel-DaSilva:
This small PR fixes a style issue in the menu class for the select example, so it keeps its max size and show the scrollable section when a long list of elements are provided


TGlide:
Thanks for the PR!

Issue #39, opened 2023-06-15
Select value type should be `any`


huntabyte:
I'm thinking that the `value` of the Select component should be of type `any` to align with Svelte's element types for the `HTMLSelectElement` / `HTMLInputElement`.

I assume a common practice will be binding the selected value to one of those native elements, and when data is coming from a database which could possible be `null` / `undefined` we may run into some issues.

TGlide:
Resolved by #44

PR #44, opened 2023-06-16
refactor(select): label attr & unknown value

TGlide:
## Description
- Adds optional label attribute to select
	- When label is not present, uses textContent
- Sets type of selected to `unknown`

Motivations for this are: It's more intuitive to not pass in a label, coming from native select. But sometimes you'd want to override this, so we allow the label attribute to be set. This will be documented accordingly.

@huntabyte @geerew would love your thoughts on this!

Closes #35 & #39

### DIFF
--- a/.changeset/quick-carpets-arrive.md
+++ b/.changeset/quick-carpets-arrive.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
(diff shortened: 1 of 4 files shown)


huntabyte:
Yes - I support this idea 100%!

geerew (approved):

PR #46, opened 2023-06-16
fix: prevent focus on mousedown

### LINKED ISSUE #43: 🐛Bug: Select focus causes flicker
**Describe the bug**
When a select has a `focus:`style, it will result in a _flicker_. The select box will gain focus for a very short amount of time and then the popup gets the focus. This results in a flicker

An example of how it should work is -> https://captaincodeman.github.io/svelte-headlessui/listbox/

**To Reproduce**
1. go to https://www.melt-ui.com/docs/builders/select
2. ensure the select component is not in focus
3. click the select component

there wil be a slight orange flicker while the select component gains and looses focus

**Expected behavior**
Focus should be ignored when the popup is about to happen

**Screenshots**
[video example](https://github.com/melt-ui/melt-ui/assets/494875/f9e714f3-0f3f-4573-9c3c-5f15f9948303)

 - OS: MacOS
 - Brower 

n/a

**Additional context**
n/a

marcelgundermann:
Closes #43

### DIFF
--- a/.changeset/slimy-laws-float.md
+++ b/.changeset/slimy-laws-float.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+fix(select): prevent focus on mousedown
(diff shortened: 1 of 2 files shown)


TGlide:
Thank you for the PR! LGTM :)

PR #51, opened 2023-06-17
feat: add color transition for switch component

marcelgundermann:
This pull request adds just a small enhancement to the switch component by including the `transition-colors` class, which enables a smooth transition effect for the background color change.


TGlide (approved):
Thank you! LGTM 😄

PR #53, opened 2023-06-17
fix: add transition-colors class to code.ignore-svelte

marcelgundermann:
Added the `transition-colors` class to `code.ignore-svelte` in the switch component.


TGlide:
LGTM, thanks!

PR #54, opened 2023-06-17
improve: popper options/config

huntabyte:
This PR gives us the ability to disable the `useClickOutside` and `usePortal` actions for the internal `usePopper` action. It also enables the popper to accept options for the portal, defaulting to `'body'` if an element or other string isn't passed.


TGlide (approved):
LGTM, good idea!

PR #60, opened 2023-06-18
Update copy in +page.svelte

geerew:
Fixes the copy button on the home page to npm install @melt-ui/svelte


TGlide:
LGTM, thanks!!

PR #69, opened 2023-06-19
feat: require user interaction to close alert dialogs

marcelgundermann:
This pull request introduces a new feature where alert dialogs now require user interaction to be closed.


TGlide:
Thank you for the PR! But I don't think we should introduce this change, as it goes against the [WAI-ARIA guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/examples/alertdialog/)

image

huntabyte:
What's the specific use case for this @marcelgundermann ? If there was an event fired when the user closed that you could listen to would that help with your issue? Something like "on:dialogclose"?

marcelgundermann:
@TGlide Thank you for the hint, i didn't know this.

@huntabyte I previously thought that alert dialogs couldn't be closed by pressing the escape key either.

PR #76, opened 2023-06-21
Add Separator component

kaevupoiss:
I have tried to implement the separator component. It doesn't seem like I have to return anything else from the `createSeparator()` builder.

### DIFF
--- a/src/lib/builders/separator/index.ts
+++ b/src/lib/builders/separator/index.ts
@@ -1,0 +1,23 @@
+import { elementDerived } from "@melt-ui/svelte/internal/helpers";
+import { writable } from "svelte/store";
+
+type SeparatorArgs = {
+    orientation?: 'horizontal' | 'vertical';
+    decorative?: boolean;
+}
+
+export function createSeparator(args: SeparatorArgs = {}) {
+    const orientation = writable(args.orientation ?? 'horizontal');
+    const decorative = writable(args.decorative ?? 'false');
+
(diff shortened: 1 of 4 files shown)


src/lib/builders/separator/index.ts
```diff
@@ -0,0 +1,23 @@
...
+import { writable } from "svelte/store";
+
+type SeparatorArgs = {
+    orientation?: 'horizontal' | 'vertical';
+    decorative?: boolean;
+}
+
+export function createSeparator(args: SeparatorArgs = {}) {
+    const orientation = writable(args.orientation ?? 'horizontal');
+    const decorative = writable(args.decorative ?? 'false');
```
TGlide, line 11: These should be returned so they can be modified after creation

src/routes/docs/builders/separator/preview.svelte
```diff
@@ -0,0 +1,35 @@
+<script lang="ts">
+	import { Docs } from '$routes/(components)';
+	import { createSeparator } from '@melt-ui/svelte/builders/separator';
```
TGlide, line 3: Should be $lib. Also, don't forget to export it from '$lib/builders/index.ts' as well

TGlide (changes requested):
Thank you for the PR! Just some small changes needed 😄

huntabyte:
Closing this in favor of #88 - thank you!

PR #77, opened 2023-06-21
Fix pagination issue when fewer than 7 pages

geerew:
There is a bug when there are fewer than 7 pages. This PR fixes this

When there are 1 - 6 pages, all the pages will show for example 

`1 2 3 4 5 6` if there are 6 pages 
`1` is there is only 1 page

when there are 7 pages or more, the original functionality takes over


TGlide:
Thank you for the PR! Though I'm closing this in favor of https://github.com/melt-ui/melt-ui/releases/tag/v0.12.0. 🙂

PR #82, opened 2023-06-22
Documentation for toggle and allow it disabled state to be changed

### LINKED ISSUE #81: 🐛Bug: Toggle's disabled state cannot be changed after creation
I was looking into writing docs for [Toggle](https://www.melt-ui.com/docs/builders/toggle) but noticed it seems to be impossible to disable a toggle or enable a toggle after creation.

I assume this is a bug since setting `disabled={true}` on the button you turn into a toggle also doesn't disable it

Since  the `disabled` store inside it isn't returned like the `pressed` store is, which can be seen in return statement in the bottom of the [builder](https://github.com/melt-ui/melt-ui/blob/develop/src/lib/builders/toggle/index.ts)

NikiSkaarup:
Updates documentation for [toggle](https://www.melt-ui.com/docs/builders/toggle), so it is up the same standard as `Accordion` and `Checkbox`

Also fixes #81 so it possible to disable the toggle after creation

### DIFF
--- a/.changeset/thirty-numbers-drive.md
+++ b/.changeset/thirty-numbers-drive.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+feat(toggle): allow disabled state to be changed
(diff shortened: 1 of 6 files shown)


TGlide (approved):
Hey, this is amazing! Thank you so much 😄

PR #84, opened 2023-06-22
Documentation for switch

NikiSkaarup:
Updates documentation for [switch](https://www.melt-ui.com/docs/builders/switch), so it is up the same standard as other builders


TGlide (approved):
LGTM! 🎉

PR #87, opened 2023-06-24
Feat: Label builder

FranckPoingt:
Introducing the `createLabel` builder.

This includes: 
- builder
- documentation
- tests


huntabyte:
I removed the `asterisk` as although it's a nice touch, I think it adds unnecessary complexity to the headless component and feels more like something that should be up to the consumer to implement. Especially since we don't want to delve into adding classes & be truly "headless".

TGlide (approved):
Awesome work!!! 😄 🎉

Issue #90, opened 2023-06-24
⚙Tech Debt: Svelte 4 support


TGlide:

softwarecurator:
what needs to be done to support this?

PR #91, opened 2023-06-25
Doc overhaul with action builders & new layout

huntabyte:
This PR updates the docs to the refreshed layout, adds initial API References/doc pages for all existing components, and includes the action builder updates.

Also adds support for Svelte 4, which will close: #90


Suyashtnt:
How can I use this while I wait for a merge? Currently making a new project that uses svelte 4 and I would like to use melt-ui

Suyashtnt:
Welp my answer is next release

PR #97, opened 2023-06-26
Fix: focus trap not being applied to drawer example in dialog docu

bennymi:
Hi, just a quick fix for the drawer example not having a focus trap.


TGlide:
Thank you, nice catch!

PR #103, opened 2023-06-27
Fix spelling error in label docs

jannisrichter:
A control that add a label to an input element. 
->
A control that adds a label to an input element.


TGlide (approved):
LGTM, thanks 😄

PR #109, opened 2023-06-28
fix: docs scrollbar styling

kaevupoiss:
This fixes a theming issue where the scrollbars appeared too light compared to everything else and felt distracting.
This also fixes a minor problem where the example box corners were not rounded on the right side.

The background is left transparent. The scrollbar-thumb uses a lighter shade of the same color palette as the body.

### DIFF
--- a/src/app.postcss
+++ b/src/app.postcss
@@ -36,3 +36,30 @@ code:not(pre > code) {
 		color: theme('colors.zinc.300');
 	}
 }
+
+/* Firefox */
+* {
+	scrollbar-width: thin;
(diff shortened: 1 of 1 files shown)


TGlide:
Thank you for the PR! We'll eventually use scroll-area once it's implemented, but it's great to have a workaround until then.

It seems it adds a small white circle at the bottom right of code-panels. Any idea why?
Screenshot 2023-06-28 at 09 17 32

kaevupoiss:
Yeah that's my oversight, sorry. I didn't test it with both horizontal and vertical scrollbars showing together.

TGlide:
Thank you!

PR #110, opened 2023-06-28
feat: css example for separator

kaevupoiss:
If this PR is alright I will continue porting the other components over to CSS as well.

Should I make a PR for every component separately or should I just make a single large PR for this?


TGlide:
Thank you so much, this is great! 🎉 

For the other PRs, I'd prefer having separate PRs, makes it easier to review. But if you want to group some together, e.g. adding 3 builders at a time, that'd be okay too 😄

PR #112, opened 2023-06-28
Initial PR for tag-inputs

geerew:
Features: 

- Type in the input and press enter to add tags
- Prepopulate with tags
- Delete tags
- Disable everything or disable specific tags
- Option to only allow unique tags
- Keyboard navigation

Would love a full review and to know if there is anything optimisation that could be done, or rework needed

I have additional features I would like to add such as 

- tag min/max -> number of allowed tags
- tag char min/max -> the min/max length of a single tag
- on:blur -> do nothing, clear input, add input as tag
- support for validation callback

Some of these features would need a discussion.. 

Another point is that each tag get a unique ID. So a tag is `{id: string, value:string}`. I did it this way so we can uniquely identify tags when navigating, deleting, etc. However I noticed Zag just use the `index`. So when Zag render a tag they pass in the tag value and the tag index. This would match up with an internal store and for the most part should be ok. 

I wasn't happy doing it this way, as there could be a discrepency between the tag index displayed on screen and then position of the tag in store. So I opted for uuid's. Im opening to reworking to use index though..


geerew:
- Made the tags-input wider 
- Added suport to select last tag when caret is at the start of the input and user types backspace

TGlide (approved):
LGTM, excellent work! 🎉

Issue #117, opened 2023-06-28
🐛Bug: Shouldn't scroll be locked when the dropdown menu is open?


buhodev:
**Describe the bug**
Is scroll supposed to work when an instance of the dropdown menu is open?

**To Reproduce**
1. Go to [melt-ui.com/docs/builders/dropdown-menu](https://www.melt-ui.com/docs/builders/dropdown-menu)
2. Click on the demo dropdown menu button
3. Scroll down or up
4. See bug (scroll is allowed)

**Expected behavior**
If I follow the same steps in [radix-ui.com/docs/primitives/components/dropdown-menu](https://www.radix-ui.com/docs/primitives/components/dropdown-menu), scroll is locked.

**Screenshots**

In this screenshot, I opened the dropdown menu, and then the submenu. I can scroll to the end of the page. Is this expected behaviour?
![image](https://github.com/melt-ui/melt-ui/assets/105121122/4ee43684-621a-494b-9766-c58cf15327cb)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome

huntabyte:
You are right about that one! Thanks for catching it!

PR #119, opened 2023-06-28
fix: prevent scroll on dropdown option

huntabyte:
Exposes a `preventScroll` option to the `createDropdownMenu` builder which is set to `true` by default to prevent scrolling while the dropdown is open.

Closes: #117


TGlide (approved):
Awesome! 🎉

PR #121, opened 2023-06-29
feat: Context Menu

huntabyte:
This PR adds the `Context Menu` builder function, which has the same functionality as the `Dropdown Menu`, but displays the menu at the pointer position and is triggered by a right click or long-press in the trigger inside the trigger element. 

**TODO:**
- [x] Figure out how to best share code between the various menus


src/lib/builders/context-menu/index.ts
```diff
@@ -0,0 +1,1346 @@
...
+
+	const checkboxItemDefaults = {
+		checked: writable(false),
+	};
+
+	const checkboxItem = hiddenAction({
+		role: 'menuitemcheckbox',
+		tabindex: -1,
+		'data-orientation': 'vertical',
+		'data-melt-part': 'item',
```
TGlide, line 409: Shouldn't it be checkbox?

TGlide (approved):
Awesome, you're on fire! I do wonder about two behaviours:
- Do we need an arrow for context menu? Maybe it's a nice to have, but for the preview I feel we don't need it.
- I feel the default positioning should be to the bottom right of wjere we click the mouse, instead of the center: Screenshot 2023-06-28 at 21 52 33 (imaigne that the cursor is at the top-left corner of this box)

src/lib/builders/context-menu/index.ts
huntabyte: Working on fixing this :)

huntabyte:
> Awesome, you're on fire! I do wonder about two behaviours:
> 
> * Do we need an arrow for context menu? Maybe it's a nice to have, but for the preview I feel we don't need it.
> * I feel the default positioning should be to the bottom right of wjere we click the mouse, instead of the center: Screenshot 2023-06-28 at 21 52 33 (imaigne that the cursor is at the top-left corner of this box)

You are right about both of those!

I also just consolidated a lot of the `menu` code :)

huntabyte:
@TGlide if you wouldn't mind taking a look at how I'm exporting/handling the types. It feels messy/hectic but I'm unsure right now what the most elegant way to handle it is so any opinions are appreciated!

TGlide:
> @TGlide if you wouldn't mind taking a look at how I'm exporting/handling the types. It feels messy/hectic but I'm unsure right now what the most elegant way to handle it is so any opinions are appreciated!

They look fine to me! We could separate them into a types.ts, since the file is so big, but no big deal if not.

TGlide:
Also, @huntabyte some merge conflicts :(

PR #124, opened 2023-06-29
[Tags Input] Doc improvements and additional features/fixes

geerew:
- Improves the docs dramatically
- Adds support for on:blur -> do nothing, add tag, clear input
- Adds support for on:paste -> do nothing, add tag
- Adds data-invalid attribute to $root and $input when the tag to add is deemed invalid (not unique if unique === true, not in allowed list, in denied list, etc)
- Adds support for allowed/denied lists
- Support for add and remove functions
- Fixes small issues


src/lib/builders/tags-input/index.ts
```diff
@@ -22,15 +32,30 @@ export type CreateTagsInputArgs = {
 	// Whether the tags should be unique
 	unique?: boolean;
 	// What to do on blur
-	// blur?: 'nothing' | 'add' | 'clear';
+	blur?: 'nothing' | 'add' | 'clear';
+	// What to do on paste
+	addOnPaste?: boolean;
+	// Maximum number of tags
```
mikroskeem, line 38:
```suggestion
	// Maximum number of tags (undefined || 0 or lower === no max)
```

src/lib/builders/tags-input/index.ts
```diff
@@ -22,15 +32,30 @@ export type CreateTagsInputArgs = {
...
+	maxTags?: number;
+	// Allowed tags (undefined || 0 or lower === no max)
```
mikroskeem, line 40:
```suggestion
	// Allowed tags
```

src/lib/builders/tags-input/index.ts
```diff
@@ -22,15 +32,30 @@ export type CreateTagsInputArgs = {
 	// Whether the tags should be unique
 	unique?: boolean;
 	// What to do on blur
-	// blur?: 'nothing' | 'add' | 'clear';
+	blur?: 'nothing' | 'add' | 'clear';
+	// What to do on paste
```
mikroskeem, line 36:
```suggestion
	// Whether the tag should be added on paste
```

Perhaps?

src/routes/docs/builders/tags-input/(sections)/Anatomy.svelte
```diff
@@ -0,0 +1,27 @@
...
+	<Docs.Li><b>root:</b> The root container for the tags input</Docs.Li>
+	<Docs.Li><b>tag:</b> A tag element</Docs.Li>
+	<Docs.Li><b>deleteTrigger:</b> A button element which deletes a tag</Docs.Li>
+	<Docs.Li><b>input:</b> The input textbox for adding new Tags</Docs.Li>
+</Docs.Ul>
+
+<u>Helpers</u>
+<Docs.Ul>
+	<Docs.Li><b>isSelected:</b> Returns true if a tag is selected</Docs.Li>
+</Docs.Ul>
```
TGlide, line 27: These shouldn't be part of Anatomy IMO, as the anatomy section should only be concerned with the markup structure. Stores and Helpers should instead be mentioned in an API reference section.

src/routes/docs/builders/tags-input/(sections)/Description.svelte
```diff
@@ -0,0 +1,351 @@
...
+<Docs.H2>Builder Description</Docs.H2>
+
+<Docs.P class="mt-4">
+	The following is an in-depth description of the return values from
+	<Docs.Code>createTagsInput()</Docs.Code>.
+</Docs.P>
+
+<Docs.H3><u>Stores</u></Docs.H3>
+<Docs.H4>options</Docs.H4>
+<Docs.P class="mt-4">
```
TGlide, line 14: We should avoid adding classes to the Docs components, to keep a consistent styling throughout the docs.

src/routes/docs/builders/tags-input/(sections)/Description.svelte
```diff
@@ -0,0 +1,351 @@
...
+</script>
+
+<Docs.H2>Builder Description</Docs.H2>
+
+<Docs.P class="mt-4">
+	The following is an in-depth description of the return values from
+	<Docs.Code>createTagsInput()</Docs.Code>.
+</Docs.P>
+
+<Docs.H3><u>Stores</u></Docs.H3>
```
TGlide, line 12: We should avoid adding underline to H3, to keep a consistent style.

src/routes/docs/builders/tags-input/(sections)/Description.svelte
```diff
@@ -0,0 +1,351 @@
+<script lang="ts">
+	import { Docs } from '$routes/(components)';
+</script>
+
+<Docs.H2>Builder Description</Docs.H2>
+
+<Docs.P class="mt-4">
```
TGlide, line 7: This section is quite elaborate, and I appreciate all the effort that's gone into it! I think it's easy to get lost in it a bit though.

In general, Melt UI's philosophy for approaching docs is to first show concise usage exampoles for common use cases, and show an API reference for anything more specific. Skimming through this section, I think it can be repurposed so that certain parts are moved to Usage, and others are moved to API Reference. Wdyt?

TGlide: Otherwise, we need to think of a clear way to show information like this across builders.

src/lib/builders/tags-input/index.ts
```diff
@@ -22,15 +31,34 @@ export type CreateTagsInputArgs = {
...
-	// blur?: 'nothing' | 'add' | 'clear';
+	blur?: 'nothing' | 'add' | 'clear';
+	// Whether the tag should be added on paste
+	addOnPaste?: boolean;
+	// Maximum number of tags (undefined || 0 or lower === no max)
+	maxTags?: number;
+	// Allowed tags
+	allowedTags?: string[];
+	// Denied tags
+	deniedTags?: string[];
```
TGlide, line 42:
```suggestion
	// Allowed tags
	allowed?: string[];
	// Prohibited tags
	prohibited?: string[];
```

We can find another word if prohibited sounds too verbose.

src/lib/builders/tags-input/index.ts
```diff
@@ -22,15 +31,34 @@ export type CreateTagsInputArgs = {
...
+	// A custom validator function. A response of true means the validator passed
+	validator?: (tag: string) => boolean;
```
TGlide, line 44: Instead of having a validator, we should probably expose a hiddenInput that holds the value of the tagsInput. Then you can use HTML5 validation, JS validation with something like superforms, etc, otherwise it'll probably be weird to interface Tags Input validation with other libs.

src/lib/builders/tags-input/index.ts
```diff
@@ -22,15 +31,34 @@ export type CreateTagsInputArgs = {
 	// Whether the tags should be unique
 	unique?: boolean;
 	// What to do on blur
-	// blur?: 'nothing' | 'add' | 'clear';
+	blur?: 'nothing' | 'add' | 'clear';
+	// Whether the tag should be added on paste
+	addOnPaste?: boolean;
```
TGlide, line 36:
```suggestion
	allowPaste?: boolean;
```

TGlide (commented):
Awesome PR, thank you so much! I commented some things I'd like to discuss :)

src/lib/builders/tags-input/index.ts
geerew: There is the $value store that holds the value of tags input, so validation can be done on that. However I’m not sure how you would block adding the tag then if the validation is done external to the builder?

src/lib/builders/tags-input/index.ts
geerew: allowPaste is too vague imo. 

I copied addOnPaste from zag js. The idea being that it will automatically add a tag on paste (if valid)

I’ll change it though if you want

src/routes/docs/builders/tags-input/(sections)/Description.svelte
geerew: Yeah. I guess I went a little overboard with the docs.

Tags input has a lot of moving parts to it so I tried to covey how to use it clearly. 

I’ll strip it all back though

src/routes/docs/builders/tags-input/(sections)/Description.svelte
geerew: Hunta does this same mt-4 with a lot of his P tags in the drop-down menu docs.

That is where I based this off of..

TGlide: Oh, I didn't notice that. Nonetheless, we should not be styling these. We can do certain rules instead, e.g. if a p follows an H3, then we add mt-4, something like that.

Pinging @huntabyte to hear his thoughts

src/lib/builders/tags-input/index.ts
TGlide: I still think we should have the add function! You could block it through there, correct?

src/lib/builders/tags-input/index.ts
TGlide: > The idea being that it will automatically add a tag on paste (if valid)
Oh okay! We can leave it as is then.

src/lib/builders/tags-input/index.ts
geerew: Ok so remove the validator? You can use the add to validate and if you aren’t happy with the tag then you can throw an error or return a reject, both of which will block the add.

Happy with this? Can also monitor the $value store and do html validation there but it wouldn’t block the add unless we monitor for an attribute of some sort? HTML validation could set an attribute and we could look for this before adding and block? Or are you happy with just doing it through the add?

src/lib/builders/tags-input/index.ts
geerew: Disallowed?

src/routes/docs/builders/tags-input/(sections)/Description.svelte
geerew: It would be great if the api section supported a description column to add a little blurb..

src/routes/docs/builders/tags-input/(sections)/Description.svelte
geerew: Agreed. Sorry

src/routes/docs/builders/tags-input/(sections)/Description.svelte
geerew: I’ll remove for now and yous can handle backend then

huntabyte: I agree with you on this @TGlide - it was like your your first cigarette, once I started I couldn't stop! I'm going to clean this up in a separate PR!

geerew: Sweet! I’ll update accordingly now your stuff is done

src/lib/builders/tags-input/index.ts
TGlide: That could work!

src/lib/builders/tags-input/index.ts
TGlide: Yeah, I'd remove the validator and keep the add function.

geerew:
I have made all those changes @TGlide Hopefully it's cleaner now

src/routes/docs/builders/tags-input/+page.svelte
```diff
@@ -22,20 +27,109 @@
...
+	<Docs.P>
+		On <Docs.Code>reject</Docs.Code> or <Docs.Code>error</Docs.Code>, the input is invalidated and
+		not added to the store.
+	</Docs.P>
+
+	<Docs.P>
+		The following example sets the id via a third-party API call and forces the tag to always be
+		uppercase.
+	</Docs.P>
+	<Docs.CodeBlock code={snippets.usage3} />
```
TGlide, line 87: The example being used here is showcasing `remove`, not add

src/routes/docs/builders/tags-input/+page.svelte
```diff
@@ -22,20 +27,109 @@
...
+	</Docs.Callout>
+	<Docs.CodeBlock code={snippets.usage1} />
+</Docs.P>
+
+<Docs.P>
+	Use the return values to construct a <Docs.Code>tags-input</Docs.Code>.
+</Docs.P>
+<Docs.CodeBlock code={snippets.usage2} />
+
+<Docs.H4>add</Docs.H4>
```
TGlide, line 51: This can be title-case. Maybe "Add function"? Just Add could be fine as well.

src/routes/docs/builders/tags-input/+page.svelte
```diff
@@ -22,20 +27,109 @@
...
-		<b>root:</b> The root container for the tags input</Docs.Li
-	>
+	<Docs.Li><b>root:</b> The root container for the tags input</Docs.Li>
+	<Docs.Li><b>tag:</b> A tag element</Docs.Li>
+	<Docs.Li><b>deleteTrigger:</b> A button element which deletes a tag</Docs.Li>
 	<Docs.Li><b>input:</b> The input textbox for adding new Tags</Docs.Li>
-	<Docs.Li><b>tag:</b> The element that represents a tag</Docs.Li>
-	<Docs.Li><b>deleteTrigger:</b> The element that represents the delete button for a tag</Docs.Li>
-	<Docs.Li><b>options:</b> The current options</Docs.Li>
-	<Docs.Li><b>tags:</b> The current tags</Docs.Li>
```
TGlide, line 32: These elements can also be title-case

TGlide (commented):
Looks great! Just some small docs changes, and there are some merge conflicts.

geerew:
Changes made 😄

TGlide:
Awesome!! Thank you for the amazing contribution, and your patience during the PR review. 🎉

PR #126, opened 2023-06-29
[New Builder] Menubar

huntabyte:
### Description

This PR adds a `Menubar` component.

#### Approach

Since the `DropdownMenu`, `Menubar`, and `ContextMenu` all share a lot of the same logic, we were able to introduce an internal _Menu builder_, which is used as a base for the other menu builders.

The `createMenubar` builder function only returns a `menubar` attribute object, and a `createMenu` builder function, which works the exact same as `createDropdownMenu` from the `Dropdown Menu` component. The only differences are internal scoping logic as well as moving between menubar menus using the left & right arrow keys.

Here's an example of a simple menubar component:
```svelte
<script lang="ts">
    import { createMenubar } from "@melt-ui/svelte";

    const { menubar, createMenu } = createMenubar()
    const { menu, item, trigger} = createMenu()
    const { menu: menuA, item: itemA, trigger: triggerA } = createMenu()
    const { menu: menuB, item: itemB, trigger: triggerB } = createMenu()
</script>

<div {...menubar}>
    <button {...$trigger} use:trigger.action>File</button>
    <div {...$menu} use:menu.action>
        <div {...item} use:item.action>...</div>
        <div {...item} use:item.action>...</div>
        <div {...item} use:item.action>...</div>
    </div>

    <button {...$triggerA} use:triggerA.action>Edit</button>
    <div {...$menuA} use:menuA.action>
        <div {...itemA} use:itemA.action>...</div>
        <div {...itemA} use:itemA.action>...</div>
        <div {...itemA} use:itemA.action>...</div>
    </div>

    <button {...$triggerB} use:triggerB.action>Help</button>
    <div {...$menuB} use:menuB.action>
        <div {...itemB} use:itemB.action>...</div>
        <div {...itemB} use:itemB.action>...</div>
        <div {...itemB} use:itemB.action>...</div>
    </div>
</div>
```

As mentioned previously, the individual menus are just enhanced dropdown menus, so you could add submenus, checkbox items, radio groups, etc. in the same way you could with the dropdown menu. And of course, you'd probably want to "componentize" the menus to avoid having a ton of renames, but you get the idea.

- [x] Add roving tabindex behavior - #127


TGlide:
Small conflict @huntabyte :)

huntabyte:
> Small conflict @huntabyte :)

Done ✅

Issue #92, opened 2023-06-25
⚙Tech Debt: Define a consistent `data-melt-[name]` convention/standard


huntabyte:
The project is still early in development & we're moving fast, but I think defining a convention/standard for the `data-melt-part` attributes will not only improve our lives as maintainers/contributors but also those who wish to style their components using those attributes.

I know that we use the `data-melt-part` attributes internally to handle certain logic/functionality, but it'd be nice to advertise these to the users as well. I think [Zag](https://zagjs.com/overview/styling) handles this beautifully.

Ideally, we'd add that attribute to all the elements that we provide attributes/actions. Something that might come up that I've also considered is overlap. For example, if two different components within the same page have `data-melt-part="root"`. I think if you're running into that issue, it's probably best to break the component off into its own file and use scoped styles. The alternative is of course using some sort of prefix, which could get messy and makes the API less predictable.

This also brings to light that we're being a bit inconsistent with the names we're exposing. I understand it's nice to expose just `slider` or `progress`, but if we're going to do that for a few then we should do it for all or none. If the user wants to rename `root` to `slider`, they are capable of doing so themselves. It feels odd to have some of the "wrappers" named as their component name, and others just "root".

huntabyte:
After thinking this through a bit more and digging into multiple different components, I think it makes more sense to replace `root` with the top-level component name wherever we're having to use `root`.

For example, if we look at the `Accordion` component, rather than have `root` be the wrapper, doesn't `accordion` make more sense?

There will of course still be conflicts between the different components for things like `item`, `trigger`, etc., but as mentioned previously it'll be up to the consumer to rename those if they're using multiple within the same component/page.

TGlide:
I agree with everything here. We *should* expose data-melt-part for all components, and maybe we *should* change root to be the component name.

Leaving this open so we can work on it!

huntabyte:
Okay after shipping some more complex components - I think a better approach is similar to how Radix handles it compared to Zag. Prevents overlap/conflicts and makes life a bit easier within the builders themselves.

### Accordion
Root: `data-melt-accordion`
List: `data-melt-accordion-list`
Trigger: `data-melt-accordion-list`

It might seem like a lot, but it's more explicit and improves the follow flow I find myself needing internally:


```ts

  const isAccordionTrigger = e.target.hasAttribute('[data-melt-accordion-trigger]')

```

vs something like:

```
[...]

PR #138, opened 2023-07-01
Refactor: builder wrapper

TGlide:
## Description
This PR refactors the internals of Melt UI to use a function called `builder` to create its element-related derived stores and actions, to achieve consistency throughout the project.

I'm opening this as a draft, so it can be reviewed while its being worked upon, as to be able to raise discussions over its implementation and avoid unnecessary rewrites.

As this also addresses #92, we need to be extremely careful as to not introduce bugs with this change. @huntabyte would really appreciate your review, especially in the `Menu` related builders.


huntabyte:
Made some adjustments with the menu selectors but all seems to be working as expected (functionality-wise).

The actions aren't hidden for the items, which I assume is because we aren't filtering it out when just using the `lightable` for the cases where there aren't any store dependencies for the element like we are with the `hiddenAction` for those that do. I didn't touch that part as I don't want to step on what you already have going, just fixed up the menus :)

Other than that it seems to work well and I do like the builder-builder better than the previous one-offs we had everywhere, incredible work so far dude!

src/lib/internal/helpers/builder.ts
```diff
@@ -103,228 +93,134 @@ export function effect<S extends Stores>(
...
+						};
+					};
+					fn.action = action ?? noop;
+					return fn;
+				}
+
+				return hiddenAction({
+					...result,
+					[`data-melt-${name}`]: '',
+					action: action ?? noop,
```
huntabyte, line 164: Here we're returning a hidden action if there are `stores`, which is great, as expected

src/lib/internal/helpers/builder.ts
```diff
@@ -103,228 +93,134 @@ export function effect<S extends Stores>(
...
+			if (isFunctionWithParams(result)) {
+				const resultFn = (...args: Parameters<typeof result>) => {
+					return {
+						...result(...args),
+						[`data-melt-${name}`]: '',
+					};
+				};
+				resultFn.action = action ?? noop;
+
+				return lightable(resultFn);
```
huntabyte, line 179: But here we're just returning a `lightable`, which doesn't appear to filter out the action so it still shows the entire action string.

src/lib/internal/helpers/builder.ts
```diff
@@ -103,228 +93,134 @@ export function effect<S extends Stores>(
...
 
-		return { attach, getElement, addAction, index: index++ };
-	};
-
-	const addUnsubscriber: AddUnsubscriber = (cb) => {
-		if (Array.isArray(cb)) {
-			unsubscribers.push(...(cb.filter((a) => a !== undefined) as (() => void)[]));
-		} else {
-			unsubscribers.push(cb);
+			return lightable({
```
huntabyte, line 182: And here as well. So possibly checking if there is an action and applying the same logic as `hiddenAction` to the lightable would be ideal?

huntabyte (commented):
Great work man! Loving the new syntax... much cleaner than `.action`! Just the issue with the action strings not being filtered but other than that everything appears to work as expected and I believe it's more intuitive than the previous way we were handling things.

TGlide:
> Made some adjustments with the menu selectors but all seems to be working as expected (functionality-wise).
> 
> The actions aren't hidden for the items, which I assume is because we aren't filtering it out when just using the `lightable` for the cases where there aren't any store dependencies for the element like we are with the `hiddenAction` for those that do. I didn't touch that part as I don't want to step on what you already have going, just fixed up the menus :)
> 
> Other than that it seems to work well and I do like the builder-builder better than the previous one-offs we had everywhere, incredible work so far dude!

Oops, that's my bad!

src/lib/internal/helpers/builder.ts
TGlide: Yuo!

src/lib/internal/helpers/builder.ts
TGlide: It's not needed here, as the fn doesn't return an action. Same as on the derived above.

PR #140, opened 2023-07-02
[New Builder] Hover Card

huntabyte:
This PR introduces a new builder: **Hover Card**

The Hover Card behaves very similarly to the tooltip, with some of the following extra enhancements:
- Remains open if anything is selected inside the hover card
- Customize the `clickOutside` behavior using the `closeOnOutsideClick` option.

A basic example of an *uncontrolled* hover card looks like this:
```svelte
<script lang="ts">
	import { createHoverCard } from '@melt-ui/svelte';
    const { trigger, content, arrow } = createHoverCard()
</script>

<button {...$trigger} use:trigger.action>Hover Me</button>

<div {...$content} use:content.action>
    <div>I am content inside the hover card</div>
    <div {...$arrow} />
</div>
```


TGlide:
🔥

TGlide (approved):

PR #150, opened 2023-07-04
fix(dialog): missing flyAndScale

sunsreng:
export flyAndScale from internal to allow dialog to use it.

### DIFF
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -1,1 +1,2 @@
 export * from './builders';
+export * from './internal/helpers/style';


huntabyte:
Rather than exporting everything from our `internal/styles`, we should just include the function itself in the preview code snippet if peeps want to use the same transition effect.

Anything inside `internal` should be capable of being changed without needing to semver the package, as those APIs shouldn't be exposed to the users of the package.

huntabyte (changes requested):
We should copy the function from internal over to the preview snippet rather than exporting from internal.

sunsreng:
We should place animation to be able use, as it's also required to use.
ex: in the dialog example.

TGlide:
> @sunsreng Why was this closed?

Per @huntabyte's reply:

> Rather than exporting everything from our internal/styles, we should just include the function itself in the preview code snippet if peeps want to use the same transition effect.

> Anything inside internal should be capable of being changed without needing to semver the package, as those APIs shouldn't be exposed to the users of the package.

voxpelli:
@TGlide Yes, but that was never done?

> include the function itself in the preview code snippet if peeps want to use the same transition effect

Should I open an issue to track that then?

TGlide:
> @TGlide Yes, but that was never done?
> 
> > include the function itself in the preview code snippet if peeps want to use the same transition effect
> 
> Should I open an issue to track that then?

Feel free!

PR #151, opened 2023-07-04
[Docs] Refactor to Markdown

huntabyte:
To prepare for a 1.0 release, this PR attempts to reduce the friction in writing, updating, and maintaining documentation for Melt UI.


src/docs/components/install-tabs.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+	import { Tabs } from '$docs/components';
+</script>
+
+<Tabs tabs={['npm', 'yarn', 'pnpm']} let:tab>
```
TGlide, line 5: Is a separate component necessary for this? It's fine tho

P.S. Gonna add dynamic slot names to that Dominik Twitter thread

src/docs/components/markdown/blockquote.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+	import { cn } from '$docs/utils';
+
+	let className: string | undefined | null = undefined;
+	export { className as class };
```
TGlide, line 5: Is this React? /s

src/docs/components/markdown/pre.svelte
```diff
@@ -0,0 +1,48 @@
...
+		copied = true;
+		clearTimeout(copytimeout);
+		copytimeout = setTimeout(() => {
+			copied = false;
+		}, 2500);
+	}
+
+	function copyCodeToClipboard(node: HTMLPreElement) {
+		codeString = node.innerText.trim() ?? '';
+	}
```
TGlide, line 26: We use this in some places. Guess it'd make sense to abstract it maybe, same with the copy icon below.

> Copy builder incoming? 👀

src/docs/components/nav/mobile-nav.svelte
```diff
@@ -0,0 +1,85 @@
+<script lang="ts">
```
TGlide, line 1: Would love to have the nav come from below as we discussed, but it can come on a future interaction 😄

src/docs/components/preview.svelte
```diff
@@ -0,0 +1,120 @@
...
+	$: files = codingStyleObj !== null ? Object.keys(codingStyleObj) : [];
+
+	const { value } = createSelect({
+		value: codingStyle,
+	});
+	value.subscribe((v) => {
+		if (v === 'tailwind' || v === 'css') {
+			codingStyle = v;
+		}
+	});
```
TGlide, line 58: I'm confused why we have this and pass it to PreviewStyleSelect. Why not just create a writable and pass it instead?

src/docs/components/site-header.svelte
```diff
@@ -0,0 +1,53 @@
...
+	<div class="container flex h-14 items-center rounded-md bg-neutral-800">
+		<div class="flex">
+			<a href="/" class="mr-6 flex items-center">
+				<img src="/logo.svg" alt="Melt UI Logo" class="h-9 w-auto" />
+			</a>
+		</div>
+
+		<div class="flex flex-1 items-center justify-between space-x-2 sm:space-x-4 md:justify-end">
+			<div class="w-full flex-1 md:w-auto md:flex-none">
+				<!-- Search/CMDK here eventually -->
```
TGlide, line 16: hype.gif

src/docs/components/tabs/tab.svelte
```diff
@@ -0,0 +1,27 @@
...
+
+<button
+	{...$trigger(tab)}
+	use:trigger
+	class="rounded-md border border-transparent bg-neutral-800 px-3 py-2 text-neutral-400 transition
+        hover:opacity-100 focus:!border-magnum-400 focus:!text-magnum-400
+        data-[state=active]:border-magnum-700 data-[state=active]:py-2 data-[state=active]:text-magnum-600 data-[state=active]:opacity-100"
+>
+	<div class="flex items-center gap-2 px-1">
+		{#if tab === 'npm'}
```
TGlide, line 18: Didn't we have a Tabs component with NPM stuff already?

src/docs/content/getting-started.md
```diff
@@ -0,0 +1,117 @@
+---
+title: Getting Started
```
TGlide, line 2: Should we rename this to usage? Or at least link Installation to it

TGlide (changes requested):
Breaking a Sweat

other/globalcss.html
```diff
@@ -0,0 +1,526 @@
+<div data-rehype-pretty-code-fragment="">
```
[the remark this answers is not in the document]
huntabyte: We could name it something else but the syntax highlighter that does the highlighting and such adds that fragment. It's then used to scope the specific styles like line highlights, etc. It could possibly be added to the `pre` tag, but I'd need to look into it!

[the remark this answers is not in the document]
huntabyte: Actually I just remembered the real power. If we have a `title` for the codeblock, let's say a filename, or something of that sort. It uses this to place it in the correct location and also apply the right styles/themes to the code blocks (should we ever do a light/dark mode).

package.json
```diff
@@ -57,12 +57,13 @@
 		"@floating-ui/core": "^1.3.1",
 		"@floating-ui/dom": "^1.4.2",
 		"focus-trap": "^7.4.3",
-		"nanoid": "^4.0.2"
+		"hast-util-from-html": "^1.0.2",
+		"nanoid": "^4.0.2",
+		"shiki-es": "^0.14.0"
```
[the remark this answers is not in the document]
huntabyte: Good catch, we don't actually need `hast-util-from-html`, but we do need `hast-util-to-html` which we use to convert the highlighted code blocks into renderable html :)

package.json
```diff
@@ -57,12 +57,13 @@
...
 	},
 	"devDependencies": {
-		"svelte-wrap-balancer": "^0.0.3",
-		"@types/prettier": "^2.7.3",
 		"@changesets/cli": "^2.26.1",
+		"@huntabyte/mdsvex": "0.16.5",
```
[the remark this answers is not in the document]
huntabyte: Some custom things that make the syntax highlighting better :)

src/docs/components/tables/api-table.svelte
```diff
@@ -0,0 +1,31 @@
...
+	} from '$docs/components';
+
+	export let data: APISchema;
+</script>
+
+<div class="relative overflow-x-visible rounded-md border border-neutral-600/80 px-3 pt-16">
+	<APISectionHeading description={transformAPIString(data.description, true)} title={data.title} />
+	<!-- 
+	<H3 class="mt-4 text-magnum-400">{data.title}</H3>
+	<P>{@html transformAPIString(data.description, true)}</P> -->
```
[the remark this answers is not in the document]
huntabyte: We sure can :)

src/docs/components/tabs/tab.svelte
huntabyte: We did, it's just moved to a different directory outside of the routes!

src/docs/components/code-block.svelte
```diff
@@ -0,0 +1,38 @@
...
+			copied = false;
+		}, 2500);
+	}
+
+	function copyCodeToClipboard(node: HTMLElement) {
+		codeString = node.innerText.trim() ?? '';
+	}
+</script>
+
+<div use:copyCodeToClipboard {...$$restProps} data-rehype-pretty-code-fragment>
```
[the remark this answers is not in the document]
huntabyte: 👍

src/docs/components/info-popover.svelte
```diff
@@ -0,0 +1,28 @@
...
+	import Info from '~icons/lucide/info';
+
+	const { trigger, content, open, arrow, close } = createPopover();
+
+	export let iconClasses = '';
+</script>
+
+<button {...$trigger} use:trigger aria-label="More info">
+	<Info class={cn('h-4 w-4 text-white', iconClasses)} />
+	<span class="sr-only">Open tooltip</span>
```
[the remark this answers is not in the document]
huntabyte: We should!

src/docs/components/info-tooltip.svelte
```diff
@@ -0,0 +1,34 @@
...
+		openDelay: 400,
+		closeDelay: 250,
+	});
+
+	export let infoClasses = '';
+</script>
+
+<button {...$trigger} use:trigger aria-label="More info">
+	<Info class={cn('h-4 w-4 text-white', infoClasses)} />
+	<span class="sr-only">Open tooltip</span>
```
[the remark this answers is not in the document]
huntabyte: TBH - probably never, will delete!

src/docs/components/markdown/blockquote.svelte
huntabyte: 😂

src/docs/components/install-tabs.svelte
huntabyte: I don't think it's _necessary_, but the syntax highlighting in the markdown gets a bit funky when you start adding `let`s and slots and such!

src/docs/components/markdown/pre.svelte
huntabyte: Yeah I don't think that's a bad idea actually! Could be more of a "utility"!

huntabyte: Added an initial internal 'copy button builder' to the doc utils, seems like something we could def consider opening to the public as well!

src/docs/components/nav/mobile-nav.svelte
huntabyte: Let me see what I can whip up, might as well take care of it while it's fresh!

src/docs/components/preview-style-select.svelte
```diff
@@ -21,29 +21,38 @@
...
 	aria-label="Select"
 >
-	{$label || 'Select an option'}
+	{#if $label === 'css'}
+		CSS
+	{:else if $label === 'tailwind'}
+		Tailwind
+	{:else}
+		Select an option
+	{/if}
```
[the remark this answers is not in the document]
huntabyte: So the object it's pulling these labels from sets its props based on the filenames, in this case `tailwind.svelte` and `css.svelte`, I suppose I could take it a step further since we'll only have `tailwind` or `css` and just handle that in the util function that builds the object to get rid of this mess!

I only did it this way because CSS is all caps and tailwind is first letter caps 😆

src/docs/content/getting-started.md
huntabyte: Yeah I think usage is a better name, I was thinking we add a more intuitive "quickstart" to this page that goes through a journey of starting with a basic uncontrolled component, and then moving onto some control, customization, etc.!

src/app.d.ts
```diff
@@ -2,7 +2,7 @@
 /// <reference types="@sveltejs/kit" />
 /// <reference types="unplugin-icons/types/svelte" />
 
-import type { TextDirection } from '$lib/internal/types';
+import type { TextDirection } from '@melt-ui/svelte/internal/types';
```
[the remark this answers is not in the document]
huntabyte: I think we should be able to, will look into that!

src/docs/previews/tags-input/main/tailwind.svelte
```diff
@@ -0,0 +1,51 @@
+<script lang="ts">
```
geerew, line 1: This is an old version of the tags unit example. The newer version supports edit..

huntabyte: Fixed!

TGlide (approved):

PR #155, opened 2023-07-05
tags input: edit support

geerew:
- Adds support for editing a tag inline
- Adds update function that is called before updating a tag
- several fixes
- code cleanup

I consider this feature complete, bar any bugs that come up!


src/routes/docs/builders/tags-input/(examples)/(preview)/tailwind.svelte
```diff
@@ -30,6 +31,13 @@
 					<X class="h-3 w-3" />
 				</button>
 			</div>
+			<div
```
TGlide, line 34: Why not an `input`? Weird that we pass in `{t.value}` down at all tbh 🤔

TGlide (commented):
Amazing work, thanks! Just a small question.

src/routes/docs/builders/tags-input/(examples)/(preview)/tailwind.svelte
geerew: yeah so I was debating this, given zag uses an input. The thing is, input is 'messy'. 

Say we use input like zag does.. They are constantly calculating the `width` of the input, which is not pretty and leads to some weird text getting cut behaviour. They also have to handle other things too such as font size, etc. I've done this before in another project and it's truely annoying. I really didn't want the headache just to support an input edit field.

With a div + contenteditable, it is so much clear. The div grows automatically and cleanly. There are no parts of a letter disappearing while the new width is calculated. There is no 'jitter'. Styles are automatcially take into consideration.

As for passing in the {t.value}. I could make a change so this value is set when you start editing... so you don't need to set it ..?

TGlide: I see. That works then! I'd probably change the action type so that it *needs* to be a div, or *cannot* be an input (and other certain types).

Setting the value when you start editing would also be ideal!

geerew: Ah I wasnt aware we could specific the action type .. 

I'll look into it. Thanks for the review. 

If down the road we find there are issues with using a div + contenteditable, then we can change to an input and just deal with the complexity.

geerew:
https://github.com/melt-ui/melt-ui/assets/494875/b46ba8ba-fedd-4560-9ab4-215e049ef8ad

Just for reference, this is the zag implementation on their site ..

src/routes/docs/builders/tags-input/(examples)/(preview)/tailwind.svelte
geerew: Ignore me saying about setting the specific element on the action. I though it was something more to do with the builder(...) but it's just the typescript type being set :)

Issue #148, opened 2023-07-02
✅Pre-1.0 checklist


TGlide:
## ✅ Add new Builders
- [X] Combobox
- [X] Pin Input
- [x] Toast
- [x] Table of Contents - #264 

## Builder tasks
- [x] Accordion
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
  - [x] Tests
  - [X] CSS examples
- [x] Avatar
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
  - [X] Tests
  - [x] CSS examples
- [x] Checkbox
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
  - [x] Tests
  - [X] CSS examples
- [x] Collapsible
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
  - [X] Tests
  - [x] CSS examples
- [ ] Context Menu
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
  - [x] Tests
  - [ ] CSS examples
- [ ] Combobox
  - [ ] Builder Wrapper
  - [ ] Export `Args` and `Returned` types
  - [ ] Tests
  - [ ] CSS examples
- [x] Dialog
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
  - [x] Tests
  - [x] CSS examples
- [ ] Dropdown Menu
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
  - [x] Tests
  - [ ] CSS examples
- [ ] Hover Card
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
  - [x] Tests
  - [ ] CSS examples
- [ ] Label
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
  - [x] Tests
  - [ ] CSS examples
- [ ] Menubar
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
  - [ ] Tests
  - [ ] CSS examples
- [x] Pagination
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
  - [x] Tests
  - [x] CSS examples
- [ ] Pin Input
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
  - [ ] Tests
  - [ ] CSS examples
- [ ] Popover
  - [X] Builder Wrapper
  - [X] Export `Args` and `Returned` types
[...]

PR #157, opened 2023-07-05
[Combobox] Add Combobox builder

Vpr99:
## Summary

This PR creates a Combobox builder which is needed for the 1.0 release (#148)

## Changes

- Added an ARIA-1.2 compatible combobox + corresponding documentation 🎉 
- There were a few shared consts/utilities between the Select and Combobox builders. I centralized them in helpers. Specifically:
  - `FIRST_LAST_KEYS` and `SELECTION_KEYS` were replicated in several components. Those have been moved to `$lib/internal/helpers/keyboard`.
  - `getOptions` and `getFirstOption` were moved into `$lib/internal/helpers/list`
- Keyboard entries for F-keys, Caps Lock, and Alt were added to the `kbd` map.
- The `prev` array helper (`$lib/internal/helpers/array`) condition has been changed to return the last item in the list if the given index is less than 0, as well as exactly 0. This is helpful in this Combobox condition:
	- The `loop` configuration is set to `true`
	- The user expands the menu but doesn't highlight an item (the highlighted index is initially set to -1)
	- They press the Up Arrow and the last item in the list should now be highlighted.
	- ...without this change, the Up Arrow doesn't do anything.

### Out of scope

- **Setting a default selected item.** This should be a quick & easy fast-follow and shouldn't change the underlying architecture at all.
- **Page Up/Page Down navigation** should also probably be added to the `select` builder. It'll be a case of adding some more array utilities like `prev` and `next` that can traverse a variable amount.
-  "Non-looping items." Right now, interaction with items outside the main loop lack keyboard controls. This is definitely a nice-to-have but IMO not critical to the initial release since it shouldn't introduce breaking changes to the API. Eg:
```svelte
<script lang="ts">
	const { open, menu, item, inputValue, isSelected, filteredItems } = createCombobox({
	});
</script>

<ul class="menu" {...$menu} use:menu.action>
  {#if $open}
    {#if $filteredItems.length !== 0}
      {#each $filteredItems as book, index (index)}
        <li {...$item({ index, item: book })} use:item.action>
          {book.title}
        </li>
      {/each}
    {:else}
      <!-- 👎 This item can't be controlled via the keyboard. -->
      <li>No results found</li>
    {/if}
  {/if}
</ul>
```
- Add some more examples & evaluate how we're going to approach documentation for return types.
 
### Callouts

- We're not 100% sold in the `updateItems` callback function that we introduced to the builder. It seems like a pretty common use case to update the list of items passed to the combobox on the fly, which doesn't play nicely with the internal `items` store. We decided to explicitly provide an updater instead of:
  - Requiring `args.items` to be a writable store instead of `T[]`. 
  - Just telling people to write a function to update $options. This felt too hacky for something that we suspect will be a frequent use case.
- @davidwoolf is the best. Thanks for jamming on this with me 🤜💥🤛


geerew:
@TGlide, Should this be updated to use the new builder(…) when defining the stores and actions? so it gets the data-melt-* attributes? 

It also makes for consistent reading when spreading and setting the action on an element

Tags-inputs did this update recently for reference.

geerew:
It’s also something that can easily be done after merge though 👌

it would be good to create a list of which builders are using the new approach and which are using the older approach so we can update for a consistent API

TGlide (approved):
A W E S O M E 🎉

PR #158, opened 2023-07-06
Updated builders to use builder() and standardise preview example

geerew:
So this reworks all (with exceptions) the builders to use the builder() function and tidies the preview examples to support this change.

The builders not updated are:

- `label` - uses `hiddenAction(...)`
- `popover` -  uses `hiddenAction(...)`
- `toolbar` - uses `hiddenAction(...)`
`slider` -  uses `typedDerived(...)`

I didnt want to break these so left them for someone smarter than me to update.


TGlide (approved):
Awesome!!!1 🎉

PR #161, opened 2023-07-06
feat: Adds the PP

### LINKED ISSUE #100: ⚙Tech Debt: Melt Preprocessor
There's a bit of inconsistency among the builder props when it comes to those that are meant to be spread onto elements. Many of the spreadable props are provided as stores, while others are simple objects. 

For example, [`createRadioGroup`](https://www.melt-ui.com/docs/builders/radio-group) returns `{ root, item, isChecked }`. All 3 of them are stores. No problem here.

While other builders, such as [`createAccordion`](https://www.melt-ui.com/docs/builders/accordion), returns `{ content, item, trigger, isSelected, root }`, which are a mix of stores and objects (`root` in this case is the outlier that's not a store).

It would be great if every prop that needs to be spread onto an element is a store so that it's consistent for implementers.

If you're concerned about the potential overhead of making those props into stores, we can create a lightweight version of a readable store which doesn't have to track a list of subscribers nor does it require unsubscribing:

```ts
export function lightable<T>(value: T): Readable<T> {
    function subscribe(run: Subscriber<T>): Unsubscriber {
        run(value);
        return () => {
            // don't need to unsub from anything
        };
    }
    return { subscribe };
}
```

Taking the `root` from `createAccordion` as an example, we can simply modify it like this: 
```diff
-const root = {
-	'data-melt-id': generateId(),
-};

+const root = lightable({ 'data-melt-id': generateId() });
```

Or a prop that has a bit more going on:
```diff
// for `createToolbar`
const button = hiddenAction({
-	role: 'button',
-	type: 'button',
-	'data-melt-part': 'toolbar-item',
+	...lightable({
+		role: 'button',
+		type: 'button',
+		'data-melt-part': 'toolbar-item',
+	}),
	action: (node: HTMLElement) => {
		const unsub = addEventListener(node, 'keydown', getKeydownHandler(options));

		return {
			destroy: unsub,
		};
	},
} as const);


```

AdrianGonz97:
Closes #100


TGlide (approved):
All hail the PP!

PR #162, opened 2023-07-07
Datepicker builder

isaacfink:
Added date picker builder, there is still some housekeeping to do but the structure is there, I am opening this now because I am taking a break and might not get back to it for a while


isaacfink:
I still need to do the following

- [ ] aria labels
- [ ] respect earliest and latest date and disabled
- [ ] confirm choice (should not automatically change value on select)
- [ ] add range support
- [ ] write docs
- [ ] clean up and write tests (at least for utils)

package-lock.json
TGlide, line 1: You should be usin pnpm, not npm! Please be sure to remove this file, then use pnpm install instead.

isaacfink: @TGlide I will be sure to do that before it's completed, I was actually wondering how come there is no lockfile

ollema:
this should be closed now I think?

Issue #183, opened 2023-07-11
🐛Bug: Incorrect link for Grail UI


ahmad1702:
**Describe the bug**
The link for Grail UI in the "Inspiration & Credits" section leads to a 404.

**To Reproduce**
1. Go to [https://www.melt-ui.com/docs/introduction](https://www.melt-ui.com/docs/introduction)
2. Scroll to the "Inspiration & Credits" section
3. Click on the Grail UI link
4. See how the page gets redirected to a 404 link

**Expected behavior**
Should be taking users to the Grail UI site using the link that can be found in the [grail ui docs](https://grail-ui.vercel.app/).

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - Version: 114.0.5735.198

huntabyte:
Good catch, thank you!

TGlide:
It's not yet in main, but closing as it soon will be.

PR #184, opened 2023-07-11
#183 - fix grail-ui link according to grail-ui docs

ahmad1702:
Swapped out the incorrect grail-ui.com link for [grail-ui.vercel.app](https://grail-ui.vercel.app/) according to the [grail-ui-docs](https://github.com/grail-ui/grail-ui).


TGlide:
I did this in another commit! Thanks though 😄

PR #187, opened 2023-07-12
Table of Contents builder

### LINKED ISSUE #129: ✨ Builder: Table of Contents
Would there be an interest in a ToC component? I created one for my library ([example](https://ato-ui.vercel.app/docs/components/toc-headless)) and since I'm planning to move all dependencies to Melt UI it would make sense to transfer the ToC here albeit with some changes to match the other builders implementations. I also feel like my component is a bit overly complicated so we could probably strip away some of the options.

Let me know, if this is something you would like to add to Melt UI and I could set up a PR for it and we can discuss what the options for it should look like.

bennymi:
I added the ToC builder. The structure is very similar to how it was in my library but I adapted it a bit to how the table of contents was set up for the Melt UI documentation. Also I'm not sure if I'm properly following the builder pattern. Let me know.

- Closes #129 

**TODOs**
- [x] Add documentation
- [x] Add scroll behaviour? This would be an action that would be added to the individual anchor tags
- [x] 🐞  Accordion page generates table of contents which includes the different item headings...

image


TGlide:
Sorry for the later reply, thank you for this! I've been busy so haven't had the time to take a proper look yet. I'll probably merge this into its own internal branch to add documentation, examples and what not. Looks amazing though!

bennymi:
@TGlide great thanks! I can start with the documentation after your initial review just so that we're both on the same page about whether the current options are okay for you.

src/lib/builders/table-of-contents/types.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Arguments to be passed to the Table of Contents builder.
+ * 
+ * @param exclude - A list of headings that should be excluded in the table.
+ * @param scrollOffset - The pixel offset that should be added when scrolling to a heading.
+ * @param scrollBehaviour - Describes whether the scroll behaviour should be smooth or not.
+ * @param tocType - Describes which headings should be considered active.
+ */
+export type CreateTableOfContentsArgs = {
```
TGlide, line 17: We should also export a `CreateTableOfContentsReturn`

src/lib/builders/table-of-contents/types.ts
```diff
@@ -0,0 +1,46 @@
...
+    selector: string,
```
TGlide, line 18: We should probably also accept HTMLElement, in case they want to pass a ref

src/lib/builders/table-of-contents/types.ts
```diff
@@ -0,0 +1,46 @@
...
+    exclude?: Heading[];
+    scrollOffset?: number;
+    scrollBehaviour?: 'smooth' | 'jump';
+    tocType?: TOCType;
```
TGlide, line 22: Maybe just type?

src/lib/builders/table-of-contents/types.ts
```diff
@@ -0,0 +1,46 @@
+
+export type Heading = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+
+/**
+ * The ToC type describes which headings should be considered active.
+ */
+export type TOCType = 'lowest' | 'highest' | 'all-active' | 'lowest-parents' | 'highest-parents';
```
TGlide, line 7: I'm confused by what this does tbh 🤔

src/lib/builders/table-of-contents/create.ts
```diff
@@ -0,0 +1,259 @@
+import { addEventListener, builder, createElHelpers, executeCallbacks, generateId, isBrowser } from "$lib/internal/helpers";
+import type { Defaults } from "$lib/internal/types";
+
+import { onMount } from "svelte";
+import { get, writable, type Writable } from "svelte/store";
+
+import type { Heading, CreateTableOfContentsArgs, ElementHeadingLU, HeadingParentsLU, TableOfContentsItem, NavItemProps } from "./types";
+
+const defaults = {
+    exclude: ['h1'],
```
TGlide, line 10: Why exclude h1 by default? Just curious

src/lib/builders/table-of-contents/create.ts
```diff
@@ -0,0 +1,259 @@
...
+    scrollOffset: 0,
+    scrollBehaviour: 'smooth',
+    tocType: 'lowest',
+} satisfies Defaults<CreateTableOfContentsArgs>;
+
+// Do we need this?
+// const { name } = createElHelpers('table-of-contents');
```
TGlide, line 17: Always, if we're exposing stores that contain element props. In this case you'd use it in item

src/lib/builders/table-of-contents/create.ts
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    const item = builder('', {
+        returned: () => {
+            return {
+                role: 'link',
+                tabindex: 0
+            }
+        },
+        action: (node: HTMLElement, params: NavItemProps) => {
```
TGlide, line 221: I definitely think we should enforce to make this an A tag. Also, I'd change this so the params don't come from the action, but from the store (e.g. `melt={$item(id)}` or `{...$item(id)} use:item`. This way we can use the melt attribute. 

To get this value from within the action, you can store the id in a data attr and get it later

src/lib/builders/table-of-contents/create.ts
```diff
@@ -0,0 +1,259 @@
...
+
+        return () => {
+            observer?.disconnect();
+        };
+    });
+
+    return {
+        activeHeadingIdxs,
+        headingsTree,
+        navItem: item
```
TGlide, line 257: I think it can be just item tbh

src/lib/builders/table-of-contents/types.ts
TGlide: Okay, I think I kinda understand, but only partially. This needs to be well documented, maybe adding more info to the JSDoc comment would be stellar as well!

TGlide (commented):
This is pretty exciting stuff, thanks again! I have a few questions about it 🙂

src/lib/builders/table-of-contents/types.ts
bennymi: Yes makes sense. I'll update it.

src/lib/builders/table-of-contents/types.ts
bennymi: Yeah I'll add them to the docs, I wasn't sure if you would even want this option so I didn't expand too much on it for now. Essentially the type defines which headings are considered active, based on the content that is visible. The ones with `-parents` also mark the parent headings of the current active headings as active, and `all-active` marks all headings with visible content as active. Maybe there's a better way to name them?

I was also thinking it would be good to add a `none` option, for those that just want the ToC but don't want to add a bunch of intersection observer APIs to all the elements. What do you think?

src/lib/builders/table-of-contents/create.ts
bennymi: @TGlide The H1 tags are always the article titles so they usually don't appear in a table of contents. Actually that's why I was thinking whether I should remove it all together from the possible values. What do you think?

src/lib/builders/table-of-contents/create.ts
bennymi: I agree. Then I'll rename the `items` in the `TableOfContentsItem` type to `children` just so we don't have the same name twice

src/lib/builders/table-of-contents/create.ts
bennymi: Great! I'm still getting familiar with how the builders API works 😄 I'll add it and will let you know if I have questions about it.

src/lib/builders/table-of-contents/create.ts
bennymi: @TGlide I'm still trying to get familiar with the API, so with melt attribute you mean that the `id` passed to the store will be added as an attribute to the element? And then inside the action we can grab the id from the attribute right? What is the name of the attribute I should use for this?

src/lib/builders/table-of-contents/types.ts
TGlide: I think that's a good idea! I love this tbh, it's pretty comprehensive and feature-complete.

src/lib/builders/table-of-contents/create.ts
TGlide: I think giving them the option is nice, sometimes you'd want to include the title. But the defaults make sense

src/lib/builders/table-of-contents/create.ts
TGlide: Exactly! Maybe data-id?

bennymi:
@TGlide I updated the code and the documentation. The scroll behaviour and offset can now also be set. There are two things left to do:
1. For the preview code, it currently only shows the `tailwind.svelte` file, but not the `tree.svelte` file, so we need to find a way to show multiple files (tagging @huntabyte for this as well).
2. The accordion examples all use header tags for their buttons, so they will also show up inside the table of contents. Either we remove them or we add some additional option parameter that allows us to define headings that should be ignored for the ToC.

Hugos68:
Hey, just watching this, I was looking at the documentation and first of all, nice job! It looks awesome. I just had one question that I was wondering. If I have awebsite and in my layout there is a table of contents that differs per page, how would I get the ``item`` return from the builder into those pages?

src/lib/builders/table-of-contents/types.ts
bennymi: I decided to name it `activeType` to be more precise about what it is ^^'

huntabyte:
> @TGlide I updated the code and the documentation. The scroll behaviour and offset can now also be set. There are two things left to do:
> 
> 1. For the preview code, it currently only shows the `tailwind.svelte` file, but not the `tree.svelte` file, so we need to find a way to show multiple files (tagging @huntabyte for this as well since we discussed it on discord a bit).
> 2. The accordion examples all use header tags for their buttons, so they will also show up inside the table of contents. Either we remove them or we add some additional option parameter that allows us to define headings that should be ignored for the ToC.

I want to make this as flexible yet not overtly complex as possible.

The idea of elements to ignore isn't a bad idea, just trying to consider cases where the IDs aren't known in advance or if there's something else the user may want to filter by.

Do you all think an optional `filter` function prop could work here? That would give flexibility in determining how to filter specific headings to the user and we don't end up adding a bunch of options that a single one could accomplish. It would act as a regular `array.filter` where it receives an element and whether it evaluates to true/false it stays/goes and we'd run it on the array of selected headings if it's provided.

I assume this won't be a super common need, but may be nice to have since it doesn't add a lot of maintenance overhead IMO.

To fix the issue with our site it wouldn't be a huge issue, we could just update the heading elements to use the `heading` element builder that was recently added. I also ran into that issue with the ToC I added for the docs lol - not sure if you noticed but I ended up using a `data-toc` attribute on the headings that I definitely wanted in the ToC so it still selects h2s and h3s, but only if they have that attribute which is a poor workaround but works nonetheless.

bennymi:
@huntabyte what do you think about adding a `data-toc-ignore` attribute to headings that shouldn't be included. That's what Skeleton does [here](https://www.skeleton.dev/components/table-of-contents#ignore-a-heading). Although it might not cover all the same use cases, which I can't think of right now. We could then just add the attribute to the accordion headings and leave them otherwise how they are.

huntabyte:
That will work for us internally. I'm thinking about the easiest way to cover any case the user of the TOC builder may have, which is what led me to the filter. What are your thoughts @TGlide ?

TGlide:
> That will work for us internally. I'm thinking about the easiest way to cover any case the user of the TOC builder may have, which is what led me to the filter. What are your thoughts @TGlide ?

I definitely like the filter function idea. Super expressive and easy to use.

`data-toc-ignore` can be used simultaneously, but should be overridden by the filter function IMO.

Also @bennymi for the preview, maybe we show a scrollable list with headings? And the TOC to the right? And we use the filter fn to only display those elements.

bennymi:
@TGlide @huntabyte the filter function idea is actually awesome! I love the flexibility of it. I updated the docs table of contents to filter out headings that have a parent element with `data-melt-accordion-item` attributes. 😀

When creating the preview I ran into the problem that the current scroll function wasn't scrolling the preview container, but instead it was scrolling the entire page, so I also added an extra scroll function option that allows people to overwrite the default one.

src/docs/components/table-of-contents/table-of-contents.svelte
```diff
@@ -1,49 +1,28 @@
...
+	import { createTableOfContents } from '@melt-ui/svelte';
+
+	/**
+	 * The filter function is for removing headings from the ToC
+	 * docs page where we have a preview with lots of headings.
+	 */
+	const { activeHeadingIdxs, headingsTree, item } = createTableOfContents({
+		selector: '#mdsvex',
+		exclude: ['h1', 'h4', 'h5', 'h6'],
+		activeType: 'lowest-parents',
```
TGlide, line 13: I prefer highest hehe

src/docs/previews/table-of-contents/main/tailwind.svelte
```diff
@@ -0,0 +1,106 @@
...
+				container.scrollTo({
+					top: element.offsetTop - container.offsetTop,
+					behavior: 'smooth',
+				});
+			}
+		},
+	});
+</script>
+
+<div class="relative mx-8 w-full">
```
TGlide, line 32: The preview is not working on mobile, part of it is hidden

TGlide (commented):
LGTM! Just some small comments

src/docs/previews/table-of-contents/main/tailwind.svelte
bennymi: @TGlide I updated the preview now. It's still a bit awkward given the lack of space, but now it's better than before

TGlide:
Awesome! Thanks again, amazing work 😄 

We're internally doing a refactor, so I'll merge this into a sep branch, and accomodate it to the refactor afterwards.

bennymi:
Awesome, thanks @TGlide. You created an amazing library. I'm really happy I can contribute something to it 😄

PR #190, opened 2023-07-12
[Select/Combobox] Support PageUp + PageDown keyboard controls

Vpr99:
### Summary

This PR adds support for PageUp / PageDown keyboard controls when selecting items in the Combobox and Select builders.

It also fixes some files that weren't formatted correctly—causing the CI Linting job to fail.

### Details

- Replicated some entries from `.gitignore` to `.prettierignore`
- Formatted `playwright.config.ts` and `src/docs/previews/accordion/main/tailwind.svelte`
- Added `forward` and `back` array utilities for array traversal.
- [Combobox] Added cases for <kbd>Page Up</kbd> and <kbd>Page Down</kbd> to navigate +10 items and -10 items. The [aria spec](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) is vague about the increment of movement:
  > Moves focus up an author-determined number of rows, typically scrolling so the top row in the currently visible set of rows becomes one of the last visible rows. If focus is in the first row of the grid, focus does not move.

  ...but [Downshift uses +-10](https://github.com/downshift-js/downshift/blob/1730fbeb5e64381dbc72ccf4aebd934614ab2210/src/hooks/useSelect/README.md?plain=1#L846), which seems reasonable to me.

- [Select] Migrated highlighting to use the same logic and utilities as the Combobox.

### DIFF
--- a/src/lib/builders/combobox/create.ts
+++ b/src/lib/builders/combobox/create.ts
@@ -43,7 +43,6 @@ const { name, selector } = createElHelpers('combobox');
  * Creates an ARIA-1.2-compliant combobox.
  *
  * @TODO support providing an initial selected item
- * @TODO support PAGE_UP/PAGE_DOWN navigation (+10,-10)
  * @TODO expose a nice mechanism for clearing the input.
  * @TODO would it be useful to have a callback for when an item is selected?
  * @TODO multi-select using `tags-input` builder?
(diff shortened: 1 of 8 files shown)


src/lib/builders/combobox/create.ts
```diff
@@ -271,11 +277,6 @@ export function createCombobox<T>(props: CreateComboboxProps<T>) {
 							default:
 								return;
 						}
-						/**
```
Vpr99, line 274: Turns out this was unnecessary!

TGlide (approved):
Awesome stuff! Thanks 😄

PR #192, opened 2023-07-12
[Accordion] WIP: Controlled

huntabyte:


src/lib/builders/accordion/types.ts
```diff
@@ -1,17 +1,64 @@
...
+	 * the accordion will be uncontrolled and will manage its own state. For
+	 * additional control, you can provide an `onValueChange` callback to override
+	 * the default behavior.
+	 */
+	value?: Writable<string>;
+	/**
+	 * The value of the item that should be open by default. This should only be used
+	 * when the accordion is uncontrolled.
+	 */
+	defaultValue?: string;
```
Vpr99, line 21: Thoughts on making `defaultValue` an "exclusive or" with `value`? It might prevent a common footgun to enforce this exclusivity at a type-level rather than just through documentation alone.

huntabyte: Yeah, I actually think this is a good idea! Don't look too far into this PR yet, just getting started :)

Vpr99: Fsho—definitely not trying to step on toes! It was just top-of-mind from something else I was working on 😄

TGlide:
Closing in favor of #257

PR #202, opened 2023-07-14
fixed pnpm add instead of pnpm install

n3-rd:
Installation of a package and it's dependencies should be done with `pnpm add` instead of `pnpm install`.


TGlide:
Thanks, TIL!

PR #215, opened 2023-07-15
[Accordion] Finish adding aria attributes & add example previews

huntabyte:
This PR adds the missing `aria-controls` attribute to the accordion item trigger and adds previews that showcase a `multiple` accordion, as well as an accordion with a `disabled` item. I've also added a callout to the docs that mention the importance of wrapping the trigger in either a heading element or an element with the `role="heading"` and the `aria-level` attributes to ensure their components are accessible.

### DIFF
--- a/.changeset/chatty-hotels-fetch.md
+++ b/.changeset/chatty-hotels-fetch.md
@@ -1,5 +1,7 @@
 ---
-"@melt-ui/svelte": patch
+"@melt-ui/svelte": minor
 ---
 
-[Accordion] Add missing aria attributes
+[Accordion]
+- Add missing aria attributes
+- Add `heading` builder to apply aria attributes to non-heading elements used as headings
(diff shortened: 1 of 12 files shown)


TGlide:
Should we maybe create a `heading` store, similar to radix? Or is it overkill?

huntabyte:
I guess my concern with that would be what we default the heading level to be, or if we'd just want to make that be a required argument to the call.

So something like:

```svelte
<anything melt={$heading('3')} />
```

TGlide:
> I guess my concern with that would be what we default the heading level to be, or if we'd just want to make that a required argument to the call.
> 
> So something like:
> 
> ```svelte
> <anything melt={$heading('3')} />
> ```
> 
> Let me know if you think this might be a better approach and I'll make the change and update the docs to document it!

Required argument, I'd say!

huntabyte:
You got it! @TGlide , coming right up!

PR #236, opened 2023-07-17
The Big Bad - Part 1: Builder Return Types

huntabyte:
- [x] Add a wrapper to handle creating an object of writable option stores
- [x] Update return object structure of all builders
- [x] Update examples/internal comps to use new return types 
- [x] Export types for `elements`, `options`, `actions`, `helpers`, and `states` returned from builders.


src/lib/builders/accordion/create.ts
```diff
@@ -28,10 +28,13 @@ export const createAccordion = <T extends AccordionType = 'single'>(
...
-	const options = writable({
-		disabled: withDefaults.disabled,
-	});
 
+	const type = writable(withDefaults.type);
+	const disabled = writable(withDefaults.disabled ?? false);
+	const options = {
+		type,
+		disabled,
+	};
```
TGlide, line 37: We probably can have a wrapper for this.

e.g. `const options = withWritables(withDefaults); const {type, disabled} = options`

Withdefaults should already contain the default variable for disabled, so it is weird to have `withDefaults.disabled ?? false`

huntabyte:
I'm opening this up for review to merge into `the-big-bad` branch to make the other parts easier to review. I'm thinking we wait until we get through the rest of the parts and make the final part a big documentation update, as I don't want to rewrite them 10 more times as we add more APIs 😆

TGlide (approved):
LGTM! I didn't review **all** the builder files, but looking at the overall process, I think this is great 🎉🎉🎉

PR #238, opened 2023-07-17
Add CSS examples

kaevupoiss:
Added CSS examples for
- Collapsible
- Checkbox
- Combobox

I will mention this issue for keeping track
#148 

Also, Avatar had a CSS example, but Checkbox did not.


src/docs/previews/combobox/main/css.svelte
```diff
@@ -0,0 +1,236 @@
...
+		display: block;
+
+		font-size: 0.875rem;
+		line-height: 1.25rem;
+
+		opacity: 0.7;
+	}
+
+	.check {
+		@apply absolute left-2 top-1/2 text-magnum-500;
```
TGlide, line 233: CSS shouldn't use apply, as Tailwind shouldn't be required

TGlide (changes requested):
Thank you for the PR! Small change needed

src/docs/previews/combobox/main/css.svelte
kaevupoiss: Yep, my bad. I forgot it there as the selector was already present in the tailwind example. Fixed now.

TGlide (approved):

Issue #195, opened 2023-07-13
🐛Bug: For hover card, add touch event 


jasongitmail:
**Describe the bug**
The Hover Card is currently unusable for mobile users. It's not possible to see the content.

**To Reproduce**
1. Go to https://www.melt-ui.com/docs/builders/hover-card
2. Tap on the example 
3. Nothing is shown 

**Expected behavior**
Contents should be visible to users on mobile devices.

**Smartphone (please complete the following information):**
 - OS: iOS16
 - Browser: safari
 - Version unknown

huntabyte:
When looking at [Radix's Hover Card](https://www.radix-ui.com/docs/primitives/components/hover-card), it also doesn't support touch.

It's a link so when it's touched it should take you to the link, right? Are you expecting it to function as both a touch to open the menu and also a link at the same time? @jasongitmail

jasongitmail:
Ah, in that case I'd suggest updating melt UI's description text for this to be more similar to Radix's ![image](https://github.com/melt-ui/melt-ui/assets/50032291/4db017cb-425d-416d-b30d-71eda56f8ad7) ![image](https://github.com/melt-ui/melt-ui/assets/50032291/bcd9944a-3300-432e-87d4-f79f666b65a9) 
and to include a link as part of the example. 

Otherwise, with the current description, I was expecting more of a tooltip-like behavior.

huntabyte:
Ah yeah, I see why that would cause confusion! We'll update the description, thanks for pointing that out!

PR #242, opened 2023-07-18
[Hovercard] Improve Description

huntabyte:
The current description is misleading, as the hover card is meant to preview content behind a link, rather than a replacement for a popover or tooltip.

Closes: #195


src/docs/content/builders/hover-card.md
```diff
@@ -1,7 +1,6 @@
 ---
 title: Hover Card
-description:
-  Displays additional information or options when the cursor hovers over a particular element.
+description: Enable sighted users to preview content behind a link.
```
TGlide, line 3:
```suggestion
description: Enable sighted users to preview content behind a link with mouse hover or keyboard focus.
```

Otherwise it's still not clear that touch events don't change it

geerew: Would it be worth having a call out? Make it abundantly clear that this will not render for pointer event on mobile. It is purely for keyboard/mouse events

huntabyte: @geerew Yeah I think that's a good idea tbh, to avoid any future issues around it. I'll add one somewhere toward the beginning of the docs!

huntabyte:
Not sure why this is failing CI 🤔

Issue #204, opened 2023-07-14
🐛Bug: For Tabs, keyboard navigation seems broken


jasongitmail:
**Describe the bug**

Keyboard navigation on the [tabs](https://www.melt-ui.com/docs/builders/tabs) seems broken.

**To Reproduce**

1. Go to [tabs](https://www.melt-ui.com/docs/builders/tabs) 
2. Click on Account
3. Try to use keyboard to navigate to the next tab, Password. I tried right arrow and also tab key just in case.
4. Right arrow doesn't change tabs and using Tab key (which I don't really expect to work for tabs) adds an apparent focus below the entire tab row.

**Expected behavior**

It should be possible to navigate from Account into Password using the keyboard, I assume using right arrow but I haven't looked up ARIA rules on this.

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Safari
 - Version 16.5

huntabyte:
When pressing the tab key within the tab list, it looks like the focus is below the entire tab row but it's really on the tab content panel.

![image](https://github.com/melt-ui/melt-ui/assets/64506580/390093b1-2b1b-4ad3-8319-d8ba83699b5f)
[Source - WAI-ARIA Tabs Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/)

TGlide:
This is tricky to be honest. We can check if the tab panel contains focusable elements and call it a day, and provide a way to override it?

huntabyte:
After re-reading the issue, you mention you can't navigate the tabs AT ALL on Safari with the arrow keys. This is not the correct/expected behavior.

I'm only able to reproduce it on Safari. Safari might be the death of us, lol; fortunately, I'm recently a Mac user, so I can try to figure out what the issue is here; I'm assuming Safari handles focus differently than the others.

PR #243, opened 2023-07-18
[Tabs] Force focus on click for buttons in safari

huntabyte:
This is odd but it at least allows you to click and navigate the tabs with the arrow keys. The issue of not being able to go back to the tabs using the keyboard (unless I'm missing a setting) is still there.

Closes: #204

I do think we should discuss how to best approach supporting Safari's odd native behavior.


TGlide:
Why is it always Safari 😅

Issue #251, opened 2023-07-19
🐛Bug: Dialog can be closed with mouse right button click


markkkkas:
**Describe the bug**
Right now it's possible to close Dialog with mouse right click, though it shouldn't.

**To Reproduce**
1. Go to https://www.melt-ui.com/docs/builders/dialog
2. Click on `Open Dialog`
3. Click right button outside dialog content (overlay)
4. Dialog is closed

**Expected behavior**
Dialog is not closed

PR #252, opened 2023-07-19
fix: Dialog is closed on right button click

markkkkas:
Resolves: https://github.com/melt-ui/melt-ui/issues/251

This is very straight forward changes for closing `Dialog` component only if the left click was clicked outside `content`.


TGlide:
Awesome, thank you!

PR #254, opened 2023-07-19
fix: checkboxItem and radioItem closing on click

TGlide:


huntabyte:
Isn't the expected behavior that they close on click? Even menus within VSCode or the browsers that are checkbox/radio close on click.

Maybe an option to disable it but I don't think disabling it altogether is a good idea.

TGlide:
> Isn't the expected behavior that they close on click? Even menus within VSCode or the browsers that are checkbox/radio close on click.
> 
> Maybe an option to disable it but I don't think disabling it altogether is a good idea.

Yeah, maybe you're right. I was looking at a lib that had it like this and thought it was a good idea! But now I'm thinking otherwise...

With controlled usage from #234 , this probably won't be needed. Will close it

huntabyte:
That was my thought as well - controlled would make this possible for specific use cases! I do see it being nice in certain situations, specifically, if it's like a filter menu for a table or something, I think it'd be annoying to close each time but then other times I would want it to close (selecting a color theme or something like that)!

PR #260, opened 2023-07-21
The Big Bad - Part 3: Custom Melt Events

huntabyte:
This PR introduces the automatic dispatching of custom melt (`m-`) events for each regular event that we're listening to/handling within the builder actions.

By dispatching these custom events within our normal handlers, we give users of the library the ability to prevent our default behavior should they choose to. Most of those using these will be edge cases/users that want a very particular behavior, and they understand that by overriding the default behavior, they may lose some functionality.

From a builder development perspective, there aren't a lot of differences. For example, if we look at this Toggle builder example, which has some parts omitted for brevity:

```ts
type RootEvents = {
	'on:m-click'?: MeltEventHandler<MouseEvent>;
};

const root = builder('toggle', {
	// other stuff,
	action: (node: HTMLElement): ActionReturn<unknown, RootEvents> => {
		const unsub = addMeltEventListener(node, 'click', () => {
			const $disabled = get(disabled);
			if ($disabled) return;
			pressed.update((v) => !v);
		});

		return {
			destroy: unsub,
		};
	},
});
```

The major differences are that we type the `ActionReturn`, with the events that we're listening to within that action. We also use a special event listener wrapper `addMeltEventListener`, which has all the same behavior as our normal `addEventListener`, except it also dispatches that same event as `m-<ORIGINAL_EVENT_TYPE>`, with a special detail structure.

Here's what that custom event looks like:
```ts
const customMeltEvent: MeltEvent<typeof originalEvent> = new CustomEvent(
	`m-${originalEvent.type}`,
	{
		detail: {
			cancel: () => {
				originalEvent.preventDefault();
			},
			originalEvent,
		},
	}
);
```
As you can see, the `detail` of that custom event has a convenience `cancel()` function, which will call `preventDefault()` on the original event, canceling our normal behavior. It also includes the original event as well, in case they need that.

I went back and forth on just calling the cancel method `preventDefault`, but it feels weird calling `e.detail.preventDefault()`, which will likely get mixed up with `e.preventDefault()`.

The nice part about adding those custom directives to the Action's Return Type, is that we get type safety for each element that has custom events, so by typing `on:m-` on an element with custom events, you'll see a list of all the events available for you to handle/override.
image

It will also cause a type error and for svelte-check to fail if you try to use a custom event that isn't typed for that element/action
image


src/lib/builders/accordion/create.ts
```diff
@@ -130,7 +137,7 @@ export const createAccordion = <T extends AccordionType = 'single'>(
 						return $value === itemValue ? undefined : itemValue;
 					});
 				}),
-				addEventListener(node, 'keydown', (e) => {
+				addMeltEventListener(node, 'keydown', (e) => {
```
TGlide, line 140: We should probably add a Space keydown handler, as cancelling the click makes space not work anymore. Sometimes you may want to only disable clicks (IDK why anyone would do that, butttt)

src/lib/builders/accordion/create.ts
```diff
@@ -107,9 +114,9 @@ export const createAccordion = <T extends AccordionType = 'single'>(
 				};
 			};
 		},
-		action: (node: HTMLElement) => {
+		action: (node: HTMLElement): ActionReturn<unknown, TriggerEvents> => {
```
TGlide, line 117: Rather than pass unknown to ActionReturn, since no parameters are present, we should probably specify undefined.

Also, we could export a MeltActionReturn type, that accepts Parameters just the same, but for events it just accepts a list of event types, e.g. `MeltActionReturn<undefined, 'click' | 'keydown'>`, which automatically maps the event types to the appropriate event handlers. Wdyt?

huntabyte: I think that's a great idea tbh!

TGlide (approved):

Issue #277, opened 2023-07-24
🐛Bug: Combobox Input Clicks


huntabyte:
**Describe the bug**
When the combobox menu is open and you click the input, it quickly closes and then opens again. The same thing happens when you click the label which probably shouldn't be the case.

**To Reproduce**
1. Go to https://melt-ui.com/docs/builders/combobox
2. Click the input (the menu should open)
3. WIth the menu still open, click the input again

**Expected behavior**
If you click the input and the menu is already open, it shouldn't close and reopen again

**Proposed Solution**
Use the `handler` within the internal `clickOutside` action to ignore the click if the target was the label, content within the label, or the input itself.

fkling:
It appears that the fix (or maybe some other change after that) now causes the menu to close when clicking inside the input again while it's already open. That feels wrong to me for two reasons:
- The user might click inside the input to open it but then wants to move the caret  to a different position via the mouse. Now when they click on another character the input closes.
- Double clicking on the input, to select it, causes the menu to appear and disappear.

Arguably it's likely that in both cases the user then uses the keyboard to change the value of the input, which would show the menu again, but I still find it surprising that the menu disappears.

I'm happy to file a new bug for this if desired.

TGlide:
> It appears that the fix (or maybe some other change after that) now causes the menu to close when clicking inside the input again while it's already open. That feels wrong to me for two reasons:
> 
> * The user might click inside the input to open it but then wants to move the caret  to a different position via the mouse. Now when they click on another character the input closes.
> * Double clicking on the input, to select it, causes the menu to appear and disappear.
> 
> Arguably it's likely that in both cases the user then uses the keyboard to change the value of the input, which would show the menu again, but I still find it surprising that the menu disappears.
> 
> I'm happy to file a new bug for this if desired.

You have a point. Feel free to open up a new issue.

PR #278, opened 2023-07-24
[Combobox] Fix Open/Close Flickering

huntabyte:
Closes: #277 

This PR adds a `label`  return to the Combobox builder to ensure the `aria-labelledby` of the input is set to an actual element. 

It's also to ensure the menu doesn't do the close/open flicker when clicking on the label or content within it when the menu is already open.

An alternative I considered was having the user import the label from the label builder separately, but then we'd also have to return an `id` or have them provide an `id` to use for the label and ensure they apply it to the label. This seemed like a bit much to expect. So I went with the more hands-off approach.

### DIFF
--- a/.changeset/silly-dodos-clean.md
+++ b/.changeset/silly-dodos-clean.md
@@ -1,0 +1,7 @@
+---
+"@melt-ui/svelte": patch
+---
+
+[Combobox]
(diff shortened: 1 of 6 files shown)


TGlide (approved):
Awesome! Thanks 😄

PR #284, opened 2023-07-25
✨ [New Builder] Tree

bennymi:
Since there was no PR for this yet, I thought I might give it a try 😄

**TODOs:**
- [x] be able to collapse folders
- [x] fix arrow left / right functionality to only jump in and out of folders
- [ ] asterisk functionality
- [x] click functionality
- [ ] add builder options?
- [x] update styles


TGlide:
Amazing, thanks for the PR 🎉 

Haven't taken a look at the code yet. I spotted this small bug though: 

https://github.com/melt-ui/melt-ui/assets/26071571/a380ab23-3929-4cf1-b1ae-77e03387b347

bennymi:
@TGlide yeah you're right, that looks very strange. Someone on Discord also found some flickering on the transitions. Not sure what's going on to be honest. I'll have a look at it on the weekend.

bennymi:
@TGlide I think the selection flickering should be gone now, but the animation flickering might still appear. I still saw it a few times but wasn't able to figure out what was causing it. Maybe something jumps out at you that could be the cause.

TGlide:
> @TGlide I think the selection flickering should be gone now, but the animation flickering might still appear. I still saw it a few times but wasn't able to figure out what was causing it. Maybe something jumps out at you that could be the cause.

Awesome! We can figure it out later.

Another small issue: Nor enter nor space is working to trigger the items.

Will look at the code now

src/docs/previews/tree/main/tailwind.svelte
```diff
@@ -0,0 +1,102 @@
...
+		group,
+		currentFocusedItem,
+		currentSelectedItem,
+		itemsWithHiddenChildren,
+	} = createTreeViewBuilder({ collapse: false });
+
+	setContext('tree-item', item);
+	setContext('tree-group', group);
+	setContext('hidden-children', itemsWithHiddenChildren);
+	setContext('selected', currentSelectedItem);
```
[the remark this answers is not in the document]
TGlide: Oh, now I see why, there's another component.  @huntabyte any pointers on how to add another file? Maybe we allow a folder structure `previews/tree/main/tailwind/index.svelte + tree.svelte`? While keeping the existing structure functional

src/lib/builders/tree/create.ts
```diff
@@ -0,0 +1,437 @@
...
+    EXPANDED: 'aria-expanded',
+    LABELLEDBY: 'aria-labelledby',
+    DATAID: 'data-id'
+}
+
+export function createTreeViewBuilder(args: CreateTreeViewArgs) {
+    const argsWithDefaults = { ...defaults, ...args };
+    const { collapse } = argsWithDefaults;
+
+    const { name } = createElHelpers<TreeParts>('tree-view');
```
TGlide, line 33: This can be outside of the fn, no need to recreate it for every instance of a builder

src/lib/builders/tree/create.ts
```diff
@@ -0,0 +1,437 @@
...
+        }
+    });
+
+    const label = builder(name('label'), {
+        returned: () => ({
+            id: rootIds.label
+        })
+    });
+
+    function updateFocusElement(newIndex: number) {
```
TGlide, line 70:
```suggestion
    function updateFocusedElement(newIndex: number) {
```

src/docs/previews/tree/main/tailwind.svelte
```diff
@@ -0,0 +1,102 @@
...
+	import { generateId } from '@melt-ui/svelte/internal/helpers';
+
+	const {
+		tree,
+		label,
+		item,
+		group,
+		currentFocusedItem,
+		currentSelectedItem,
+		itemsWithHiddenChildren,
```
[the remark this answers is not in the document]
bennymi: @TGlide this stores the IDs of the parents that have their children hidden. In the Svelte code you can then use it to see which subtrees to hide based on the ID of the parent. Perhaps their is a better way of doing it or naming it?

src/docs/previews/tree/main/tailwind.svelte
```diff
@@ -0,0 +1,102 @@
...
+	import type { TreeItem } from './tree.svelte';
+	import { generateId } from '@melt-ui/svelte/internal/helpers';
+
+	const {
+		tree,
+		label,
+		item,
+		group,
+		currentFocusedItem,
+		currentSelectedItem,
```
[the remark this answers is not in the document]
bennymi: @TGlide I'll update them. Do you think that `focusedItem` and `selectedItem` should contain the node, or should we have two separate stores, one which holds the focused value and one which holds the focused ID. Since in a tree you can have multiple branches which might have the same value it would be important to also return the ID. Or should we just leave it like it currently is?

bennymi:
> > @TGlide I think the selection flickering should be gone now, but the animation flickering might still appear. I still saw it a few times but wasn't able to figure out what was causing it. Maybe something jumps out at you that could be the cause.
> 
> Awesome! We can figure it out later.
> 
> Another small issue: Nor enter nor space is working to trigger the items.
> 
> Will look at the code now

I checked it on FF and Chrome, and for me it works in both. In which browser were you checking the enter and space key? They should only select, so they won't open or close any folders. For opening and closing folders you should use the left and right arrow keys.

src/docs/previews/tree/main/tailwind.svelte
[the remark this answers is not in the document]
bennymi: @TGlide maybe `closedItems` is a better name for it?

TGlide:
Hi @bennymi! Sorry that this took so long to review.

With the new changes to builder structure, this will have to be refactored a bit. Let me know if that's something you're up to doing, or if you'd rather I refactored it myself!

bennymi:
Hey @TGlide sorry for the late reply, been busy with some other stuff. I can refactor it to the new structure. Will try to commit it later today.

bennymi:
@TGlide I think a tree select builder like this one (https://github.com/riophae/vue-treeselect) would also be an awesome builder to have. It's been on my radar since March so would love to work on in down the road.

src/docs/data/builders/tree.ts
```diff
@@ -0,0 +1,183 @@
...
+	helpers: [
+		{
+			name: 'focusedItem',
+			description: 'The list item that is currently focused.'
+		},
+		{
+			name: 'selectedItem',
+			description: 'The list item that is currently selected.'
+		}
+	]
```
TGlide, line 50: These should be states, a helper is usually a function (so if we had a function `isFocused`, that'd be a helper)

src/docs/data/builders/tree.ts
```diff
@@ -0,0 +1,183 @@
...
+});
+
+const label = elementSchema('label', {
+	title: 'label',
+	description: 'The label element.',
+	dataAttributes: [
+		{
+			name: 'id',
```
TGlide, line 77: It should be data-id. If it's actually setting an id, we currently dont display that on the docs

src/docs/previews/tree/main/tailwind/index.svelte
```diff
@@ -0,0 +1,107 @@
...
+	const {
+		elements: { tree, label, item, group },
+		helpers: { focusedItem, selectedItem },
+		states: { collapsedItems },
+	} = createTreeViewBuilder({ forceVisible: true });
+
+	setContext('tree-item', item);
+	setContext('tree-group', group);
+	setContext('collapsed-items', collapsedItems);
+	setContext('selected', selectedItem);
```
TGlide, line 18: IMO this could be a single context

src/docs/data/builders/tree.ts
```diff
@@ -0,0 +1,183 @@
...
+			value: `Only applied if 'forceVisible' is false. Set to true if the group is collapsed.`
+		},
+		{
+			name: 'display',
+			value: `Only applied if 'forceVisible' is false. Set to 'none' if the group is collapsed.`
+		}
+	]
+})
+
+const keyboard: KeyboardSchema = [
```
TGlide, line 133: I would expect space or enter to have the same behaviour as click, as currently there's no way to toggle the open state of a folder

src/docs/previews/tree/main/tailwind/tree.svelte
```diff
@@ -0,0 +1,90 @@
...
+	{@const childrenAreHidden = $collapsedItems.includes(itemId)}
+	{@const isSelected = $selectedItem?.getAttribute('data-id') === itemId}
+
+	<li
+		{...$item({
+			value: title,
+			id: itemId,
+			hasChildren: children !== undefined && children.length > 0,
+		})}
+		use:item
```
TGlide, line 55: We should use the preprocessor here

src/docs/previews/tree/main/tailwind/tree.svelte
```diff
@@ -0,0 +1,90 @@
...
+			<span class="select-none">{title}</span>
+
+			<!-- Selected icon. -->
+			{#if isSelected}
+				<svelte:component this={icons['highlight']} />
+			{/if}
+		</div>
+
+		{#if children && !childrenAreHidden}
+			<ul {...$group({ id: itemId })} use:group transition:slide>
```
TGlide, line 75: We should use the preprocessor here

src/docs/previews/tree/main/tailwind/tree.svelte
```diff
@@ -0,0 +1,90 @@
...
+{#each treeItems as { title, icon, children, id }, i (i)}
+	{@const itemId = `${id}`}
+	{@const childrenAreHidden = $collapsedItems.includes(itemId)}
+	{@const isSelected = $selectedItem?.getAttribute('data-id') === itemId}
+
+	<li
+		{...$item({
+			value: title,
+			id: itemId,
+			hasChildren: children !== undefined && children.length > 0,
```
TGlide, line 53:
```suggestion
			hasChildren: children?.length > 0,
```

src/docs/previews/tree/main/tailwind/index.svelte
```diff
@@ -0,0 +1,107 @@
...
+		tree.forEach((item) => {
+			item.id = generateId();
+
+			if (item.children && item.children.length > 0) {
+				addRandomId(item.children);
+			}
+		});
+	}
+
+	addRandomId(treeItems);
```
TGlide, line 78: Maybe add a comment as to why the ids are needed

src/lib/builders/tree/create.ts
```diff
@@ -0,0 +1,422 @@
...
+        }
+    });
+
+    const label = builder(name('label'), {
+        returned: () => ({
+            id: rootIds.label
+        })
+    });
+
+    function updateFocusedElement(newIndex: number) {
```
TGlide, line 71: Nice! We have a rovingFocus helper for this, maybe it helps?

src/lib/builders/tree/create.ts
```diff
@@ -0,0 +1,422 @@
...
+                };
+            };
+        },
+        action: (node: HTMLLIElement) => {
+
+            const unsubEvents = executeCallbacks(
+                addEventListener(node, 'keydown', async (e) => {
+                    const { key } = e;
+
+                    const isLetter = /^[a-z]$/i.test(key);
```
TGlide, line 220: This is useful enough to be a global helper

src/docs/previews/tree/main/tailwind/tree.svelte
TGlide: Also, I think this transition causes the flickering. I'd say we should have a collapsedGroups rather than a collapsedItems stores. But given a tree structure like this:
```
lib/
	index.ts
	ui/
		button.svelte
		card.svelte
```

If ui is open, and lib is clicked, only lib should enter the collapsed groups. Otherwise the ui slide will trigger as well and trigger double transitions.

I feel this may be confusing for some though, so we should be careful when documenting it, or change the naming

src/lib/builders/tree/types.ts
```diff
@@ -0,0 +1,19 @@
...
+	| 'label'
+	| 'item'
+	| 'group';
+
+export type ItemDescription = {
+    hasChildren: boolean;
+    childrenIdxs: number[];
+};
+
+export type CreateTreeViewReturn = ReturnType<typeof createTreeViewBuilder>;
```
TGlide, line 19: There are more types needed here, take a look at the other builders to see what they export

TGlide (commented):
Awesome stuff, truly! I left some comments so we can discuss

src/docs/previews/tree/main/tailwind.svelte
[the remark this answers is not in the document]
TGlide: I think having the node is sufficient!

src/docs/previews/tree/main/tailwind.svelte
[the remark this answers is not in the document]
TGlide: Resolving since i think this was since changed

src/docs/data/builders/tree.ts
TGlide: Oh, I saw you mentioned the left and arrow keys now, my bad.

src/docs/data/builders/tree.ts
bennymi: Hey @TGlide, I think we would need the id here for the `aria-labelldby` attribute. We could also not add it through the builder and just show it in the preview code. How would I document that the id gets added through the builder though?

image

Right now I just added it in the data attributes. In the element schema I don't see another option of where to put it.

src/lib/builders/tree/create.ts
bennymi: Awesome, yeah that does the same thing. Using it!😄

src/lib/builders/tree/types.ts
bennymi: I also added the elements, states, and helper types now. Are there other types that also need to be exported?

TGlide:
Thank you so much @bennymi !

Will merge this into a separate branch, so I can do some tweaks if needed, as I need this for a project of mine. Will probably release it today!

Great work 🎉

PR #304, opened 2023-07-30
The Big Bad - Part 4: The PP

AdrianGonz97:
This PR replaces all instances of the `melt` attribute for the `use:melt` action.

Before merging, I'd like to:
- [x] update the PP to fix this issue: https://github.com/melt-ui/preprocessor/issues/31


TGlide (approved):
LET'S GOOOOOOOOOOO

PR #305, opened 2023-07-31
fix: Remove `@melt-ui/svelte` alias, add text transform

paoloricciuti:
This PR removes `@melt-ui/svelte` alias in favor of `$lib` while at the same time adding a regex to replace `$lib` with `@melt-ui/svelte` in the docs.

I've also replaced some imports from deeper in the tree (importing directly from a builder folder) with a more short and better `$lib`.

A couple of noted:

1. The regex replace every import that starts with `$lib` with `@melt-ui/svelte` (only in the preview code obviosuly) so even occurrencies of "$lib/builders/builderName" will be substituted.
2. Could enforce or at least warn of imports deeper in the tree in the previews folder make sense? It might be an easy enaugh custom eslint rule.
3. I've talked about this with @TGlide and that's why i'm making this PR, it's not out of the blue 😆

### DIFF
--- a/.changeset/sour-beds-clap.md
+++ b/.changeset/sour-beds-clap.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+fix: Remove `@melt-ui/svelte` alias, add text transform
--- a/src/app.d.ts
+++ b/src/app.d.ts
@@ -2,7 +2,7 @@
 /// <reference types="@sveltejs/kit" />
 /// <reference types="unplugin-icons/types/svelte" />
 
-import type { TextDirection } from '@melt-ui/svelte/internal/types';
+import type { TextDirection } from '$lib/internal/types';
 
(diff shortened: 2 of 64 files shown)


TGlide (approved):
Awesome, thank you so much! 🎉

TGlide:
> Could enforce or at least warn of imports deeper in the tree in the previews folder make sense? It might be an easy enaugh custom eslint rule.
@paoloricciuti maybe! At least we should warn against `$lib/builders/...`. Will merge this PR, but this could make sense for a future PR

paoloricciuti:
> > Could enforce or at least warn of imports deeper in the tree in the previews folder make sense? It might be an easy enaugh custom eslint rule.
> 
> @paoloricciuti maybe! At least we should warn against `$lib/builders/...`. Will merge this PR, but this could make sense for a future PR

Perfect I'll look into this later ;)

PR #306, opened 2023-07-31
chore: add eslint rule to prevent imports from lib/builders

paoloricciuti:
This rule prevent accidentally importing from `$lib/builders/*` given that the preferred way should be importing from `$lib` directly.

I've also fixed one occurence of this lint rule.

### DIFF
--- a/.changeset/green-apes-explode.md
+++ b/.changeset/green-apes-explode.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+chore: add eslint rule to prevent imports from lib/builders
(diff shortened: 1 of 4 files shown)


.changeset/green-apes-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+chore: add eslint rule to prevent imports from lib/builders
```
TGlide, line 5: No need to add changeset, as a new release for this won't impact users

paoloricciuti: Yeah I'm an idiot

TGlide: Nah you're fine! Thanks 😄

TGlide (approved):
Thanks a bunch!

PR #308, opened 2023-08-01
Add a more complete explanation for global usage of the toast builder

Manuel-DaSilva:
This is the first approach to improve the docs of the toast builder, which will help users understand how to use the toast as global functionality.

After some testing, I found that it was better to introduce this in the "Usage" section, so that the first example remains a unit. Another alternative is to introduce this as an additional example. Let me know what you think!

PD: The formatter wouldn't allow me to place the `// ...path/to/file` comments at the top of each snippet.

### DIFF
--- a/src/docs/content/builders/toast.md
+++ b/src/docs/content/builders/toast.md
@@ -24,20 +24,19 @@ you call the builder function `createToaster` inside the lib directory. This wil
 accessible to all components in your application.
 
 ```typescript
-// lib/toast.ts
 import { createToaster } from '@melt-ui/svelte'
 
 export type ToastData = {
-  title: string
-  description: string
-  color: string
+	title: string
+	description: string
+	color: string
 }
 
 export const {
-  elements: { content, title, description, close },
-  helpers: { addToast },
-  states: { toasts },
-  actions: { portal }
+	elements: { content, title, description, close },
+	helpers: { addToast },
+	states: { toasts },
+	actions: { portal }
 } = createToaster<ToastData>()
 ```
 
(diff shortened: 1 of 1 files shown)


TGlide:
Hey @Manuel-DaSilva thank you for this PR! 🎉 

The explanation in general is pretty great. I'd say what should be different, is that it shoudn't be a necessity to have two separate files.

You can do the following: https://github.com/melt-ui/reproduction-template/blob/toaster/src/lib/components/Toaster.svelte

I think this is more elegant and self-contained.

Manuel-DaSilva:
Hello @TGlide! That is an awesome idea, much better and elegant! 

I will make the respective changes, thank you!

src/docs/content/builders/toast.md
```diff
@@ -18,8 +18,110 @@ description: A succinct message that is displayed temporarily.
 
 ## Usage
 
-To create a toast, use the `createToast` builder function. You can then reference the anatomy or
-example above to create your toast.
+Unlike most builders, the toast is not component-based. Instead, it provides a global functionality
+that can be accessed from anywhere in your application. To accomplish this, it is recommended that
+you call the builder function `createToaster` inside the lib directory. This will make the function
+accessible to all components in your application.
```
TGlide, line 24:
```suggestion
Unlike most builders, the toast is not component-based. Instead, it provides a global functionality
that can be accessed from anywhere in your application. To accomplish this, it is recommended that
you create a global component that is called on the root of your application.
```

src/docs/content/builders/toast.md
```diff
@@ -18,8 +18,110 @@ description: A succinct message that is displayed temporarily.
...
+
+The first step is to create a `Toast` component that will be used to render toast notifications. We
```
TGlide, line 26: I'd call it Toaster tbh

TGlide (commented):
Awesome! Just 2 small changes needed 😄

src/docs/content/builders/toast.md
Manuel-DaSilva: Yes now that you say it, Toaster seems better!

PR #310, opened 2023-08-01
feat: store usingPreprocessor in localStorage

paoloricciuti:
This PR stores the usingPreprocessor store in the local storage.

Note: the preprocessor get's defaulted to true and flip when it reaches the browser...this could be solved by saving the value in a cookie, getting the value in the root layout and passing the default value to the `getUsingPreprocessor` function. It seemed a bit of an overkill but if you prefer do it this way i can still do it.


TGlide:
This is great! Just need to solve a small conflict @paoloricciuti

paoloricciuti:
> This is great! Just need to solve a small conflict @paoloricciuti

I'll try to solve it as soon as I get back to the desk

paoloricciuti:
Solved the conflict @TGlide

TGlide (approved):

PR #311, opened 2023-08-01
Existing CSS examples de-tailwindified

kaevupoiss:
The CSS examples do not look like CSS that humans would write. This PR is to introduce the new style of CSS examples and refactors all the existing CSS exampes to it.

CSS examples would use only color variables.

The `globals.css` file still needs to be updated. 

It only needs to include the TW reset and variables in the format of `--color-[name]-[shade]: XXX YYY ZZZ;` for colors neutral, magnum, zinc and grey. Of course also `--color-black: 0 0 0;` and `--color-white: 255 255 255;`.


TGlide:
Awesome!!!

@kaevupoiss you mentioned the globals.css file needs to be changed, but I saw that you did change it?

kaevupoiss:
@TGlide Yes! I just followed Huntabyte's instructions. Shoud've changed the PR description.

TGlide (approved):

Issue #323, opened 2023-08-01
🐛Bug: Issues with `createSlider` builder


abdel-17:
**Describe the bug**

Slider 1:
`min = 0, max = 5, step = 2`

Slider oversteps its max value.
Bug appears to happen when `max - min` is not a multiple of `step`.

Slider 2:
`min = 2, max = 6, step = 4`

Slider's value increases as soon as I focus
Bug disappears when I change `min` to 4 and `max` to 8. It appears to happen when `min` is not a multiple of `step`.

Slider 3:
`min = -5, max = 10`
Bug disappears when I change `min` to a positive value.

**To Reproduce**
1. Go to https://stackblitz.com/edit/github-wgnbi5-tntk9e?file=src%2Froutes%2F%2Bpage.svelte
2. You'll find the three sliders in the home page.

**Screenshots**

https://github.com/melt-ui/melt-ui/assets/88583085/787a2ee9-c02b-44ca-9649-b1946117514a

PR #324, opened 2023-08-01
fix: various bug slider

paoloricciuti:
This fixes basically every bug in #323 and add the ability to specify min = 0, max = 1, step = .1 (it was not possible before)

I've also added some helpers to better manage floating point add, sub, mul and div (given that numbers can be floating point in the slider).

I'm making this a draft since i want to addresso some other thing before merging

- [x] try to add tests
- [x] address setting the values from the store not adhere to the step (we decided to forgo this because it might be an unpleasant user experience)
- [x] check if we want the sliders to stop when they meet (even if set via store)...i was discussing this with @TGlide 
- [x] verify that everything is correct (it's 1 am and i'm tired, i might have messed something up)

Sorry for the weird git history

### DIFF
--- a/.changeset/dull-wasps-search.md
+++ b/.changeset/dull-wasps-search.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+fix various slider bugs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # @melt-ui/svelte
 
+## 0.31.0
+
+### Minor Changes
+
+- cba5bb7c: convert controllable returned states from readables to writables
+- 714f4b3c: Combobox: add controlled support
+
+## 0.30.1
+
+### Patch Changes
+
+- 98e05474: fix: Select & Menu keydowns leaking
+
 ## 0.30.0
 
 ### Minor Changes
--- a/other/globalcss.html
+++ b/other/globalcss.html
@@ -1,153 +1,62 @@
 <div data-rehype-pretty-code-fragment="">
 	<pre
+		class="s-k9tA7pZhqxSB mb-4 mt-6 max-h-[650px] overflow-x-auto rounded-lg border border-neutral-500/50 bg-neutral-950/60 py-4 font-bold"
+		tabindex="0"
 		data-language="css"
 		data-theme="default"
-		class="!mt-2"
-	><code data-language="css" data-theme="default"><span data-line=""><span style="color: #B392F0">:root</span><span style="color: #E1E4E8"> {</span></span>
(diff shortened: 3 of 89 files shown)


TGlide:
@paoloricciuti Oh I think we definitely want them to stop when we meet. The problem is, when they meet, and we stop dragging it, when we start dragging again, it only allows going towards one direction, which is not ideal. 

It should allow going to both, and then whichever direction you choose, is the one it should stick with until you stop dragging it again. Maybe this isn't possible though

Current behaviour:
https://github.com/melt-ui/melt-ui/assets/26071571/1cd7d9ff-b1d4-4a7e-a5a3-0b848474ac9c

paoloricciuti:
> 

The range slider component from material UI doesn't stop for example and just start dragging the other handle which to be fair seems like a reasonable thing to do (unless the user literally want's a range of 0)...maybe we should go this route too?

paoloricciuti:
https://github.com/melt-ui/melt-ui/assets/26281609/bebc1195-54c6-44d9-bdce-88da1ff1c587

I got this for the moment...it still has some quirk, like the fact that you can never have two range perfectly one over the other and the fact that depending on how fast you move you can leave your thumb way before but...does this kinda work for you guys?

TGlide:
@paoloricciuti 

This is quite good to be honest! I do think we should allow having one thumb perfectly over the other though. 

A nitpick, which is not crucial but would be nice: In your example, if you have a thumb at 10, the other at 20, and drag the left thumb quickly over the other thumb to go to 30, I would expect the thumbs to be 20-30. This doesn't seem to be the case. As I said, this is not super important, but would be ideal

paoloricciuti:
> @paoloricciuti
> 
> This is quite good to be honest! I do think we should allow having one thumb perfectly over the other though.
> 
> A nitpick, which is not crucial but would be nice: In your example, if you have a thumb at 10, the other at 20, and drag the left thumb quickly over the other thumb to go to 30, I would expect the thumbs to be 20-30. This doesn't seem to be the case. As I said, this is not super important, but would be ideal

Yeah i was trying to get the same effect too...the problem with this is that basicallly on pointermove it tries to get the actual closest thumb and swap to that. However when you move fast enaugh the mouse will "skip some bit"...i'll try to make it work however

paoloricciuti:
> @paoloricciuti
> 
> This is quite good to be honest! I do think we should allow having one thumb perfectly over the other though.
> 
> A nitpick, which is not crucial but would be nice: In your example, if you have a thumb at 10, the other at 20, and drag the left thumb quickly over the other thumb to go to 30, I would expect the thumbs to be 20-30. This doesn't seem to be the case. As I said, this is not super important, but would be ideal

Update, i got this working

TGlide (approved):
Flawless, thank you so much! 🎉

Issue #319, opened 2023-08-01
🌟Feature request: Pin input requires pressing backspace twice to delete digit


cd-slash:
**Describe the bug**
Removing a digit in the pin input requires pressing backspace twice.

**To Reproduce**
Seen on the [MeltUI demo page](https://www.melt-ui.com/docs/builders/pin-input) for the pin input.

**Expected behavior**
 a single backspace should remove the current digit and each subsequent press should remove another digit until the pin input is clear (in my opinion)

**Desktop (please complete the following information):**
 - Browser: Chrome, Polypane

wesharper:
I agree that the behavior is a bit odd although I could go either way. However, based on the documentation, `backspace` does the following: `Deletes the value of the current input. If the input is empty, moves to the previous input.`. I'd recommend changing this to a feature request instead.

cd-slash:
I did a bit of investigation by logging into various services that I know require OTP including eBay, PayPal, Amazon and various others to understand what users might expect based on typical OTP experiences. In all cases, backspace removes the previous digit.

I think this is consistent with a normal typing experience, i.e. if this was 1 single input field instead of separate inputs, a backspace would remove a character, rather than just moving to the next character and not removing it. If the user is not looking at the input and trying to type e.g. "1234" but types "1,2,4", realises they've made a mistake, hits backspace and types "3,4" in a normal input they'd get "1234" but with this behavior they'd get "1243" with the last "4" lost because the digit limit is reached.

Happy to adjust to a feature request since the current behavior was intended, it just seems inconsistent with what a given user would (probably) expect.

wesharper:
@cd-slash I've changed the functionality in my draft PR. Will wait for the maintainers to give the green light before I polish it up and attempt to get it merged.

PR #355, opened 2023-08-06
change/pin-deletion

wesharper:
This is a draft that changes the functionality of the pin-input builder to address issue #319. 

I still need to update tests and stories, but want feedback on whether this is a desirable change before doing that work.

### DIFF
--- a/.changeset/four-tomatoes-search.md
+++ b/.changeset/four-tomatoes-search.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': minor
+---
+
+Modified the behavior of the backspace key in the Pin Input builder
--- a/src/docs/data/builders/pin-input.ts
+++ b/src/docs/data/builders/pin-input.ts
@@ -150,8 +150,7 @@ const keyboard: KeyboardSchema = [
 	},
 	{
 		key: KBD.BACKSPACE,
-		behavior:
-			'Deletes the value of the current input. If the input is empty, moves to the previous input.',
+		behavior: 'Deletes the value of the current input and moves to the previous input.',
 	},
 	{
 		key: KBD.DELETE,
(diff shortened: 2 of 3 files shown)


TGlide:
This is awesome! One thing though, hitting backspace shouldn't immediately go back. Instead, it should only delete the current item, and keep the focus there. If you enter backspace on an empty item, it should move the focus back **and** delete the item on the newly focused input

wesharper:
@TGlide what you're describing is how the builder already works. I did this in response to feature request #319, which requests different behavior based on other pin inputs in the wild.

TGlide:
> @TGlide what you're describing is how the builder already works. I did this in response to feature request #319, which requests different behavior based on other pin inputs in the wild such as Amazon and PayPal. I essentially want your blessing to change the default behavior altogether.

It is not how it currently works. Currently, pressing backspace on an input without value just moves the focus to the previous input, but keeps the previous input value intact. It should delete it as well. 

Compare the current behaviour with https://zagjs.com/components/react/pin-input

cd-slash:
I wonder if it's easier to give an example - if I type `1` `2` `4` `backspace` `3` I should have the first 3 input boxes populated with `123` and the next keystroke should either populate the next box, or if a `backspace` then the 3 would be removed.

This is different from the current behavior where typing  `1` `2` `4` `backspace` `3` would leave the pin populated with `1243` because the current behavior is to simply let the backspace move the focus back to the previous input element.

Perhaps quantifying / testing the PR in terms of what the anticipated outcome is will allow us to describe the intended / desired behavior?

TGlide:
Closed in favor of https://github.com/melt-ui/melt-ui/commit/adcf8853c7840f3dc2367723efe22132d7633214

Thank you for the PR though 🙏

TGlide:
Closed in favor of https://github.com/melt-ui/melt-ui/commit/adcf8853c7840f3dc2367723efe22132d7633214

Thank you for the PR though 🙏

wesharper:
No worries, @TGlide sorry I couldn't get to it. Busy few days at work. Thanks for merging this!

PR #362, opened 2023-08-07
feature: Add trim option to input-tags

leoalves:
Added `trim` option to the `input-tags` with default value to `true`.


TGlide (approved):
Awesome!

PR #374, opened 2023-08-11
[Types] Expose ComponentEvent types

huntabyte:
This PR exports/exposes `ComponentEvent` types for the various builders which can be used to properly type components with our custom `m-` events that we dispatch.

### The problem
When you apply Melt's builders to regular elements, the attributes are properly typed to provide type safety & autocomplete for the custom events we dispatch, such as `m-click`, `m-keydown`, and so on.

Here's an example of a type that our actions infer & set the proper element attributes for:

```ts
type TriggerEvents = {
	'on:m-click'?: MeltEventHandler<MouseEvent>
}
```

However, events forwarded/dispatched from _components_ are typed a bit differently. More specifically, we don't include the `on:` portion, so the same example above, but typed for a component, would look something like this:

```svelte
<!-- AccordionTrigger.svelte (example) -->
<script lang="ts">
	type $$Events = {
		'm-click':  MeltEventHandler<MouseEvent>
	};
</script>

<button on:m-click>Forwarding an m-click event</button>
```

then it would be used and have autocomplete, properly typed event, ability to cancel, etc. like so:
```svelte
<script lang="ts">
	import { AccordionTrigger } from './AccordionTrigger.svelte'
</script>

<AccordionTrigger on:m-click={(e) => console.log(e)} />
```

Currently, this isn't possible because we don't export the `MeltEventHandler` type, as we shouldn't because it's really for internal use only. Nor would we want to, as this should be as simple as possible for the users to consume.

## The solution
I added some new internal type helpers to construct a `ComponentEvents` type for each of the builders.

At a high level, they look something like this:

```ts
type AccordionComponentEvents = {
	trigger: {
		'm-click': MeltEventHandler<MouseEvent>;
        'm-keydown': MeltEventHandler<KeyboardEvent>;
	},
   // ...any other elements that have events from the accordion builder 
}
```

They can then be used in the previous component example like so:

```svelte
<!-- AccordionTrigger.svelte (example) -->
<script lang="ts">
	import type { AccordionComponentEvents } from '@melt-ui/svelte';
	type $$Events = AccordionComponentEvents["trigger"];
</script>

<button on:m-click>Forwarding an m-click event</button>
```

This has 0 impact/implications for the existing event types we are using elsewhere, this is just in addition to.

### DIFF
--- a/.changeset/nervous-toys-sing.md
+++ b/.changeset/nervous-toys-sing.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+[Types] Expose ComponentEvent types
--- a/src/lib/builders/accordion/events.ts
+++ b/src/lib/builders/accordion/events.ts
@@ -1,7 +1,8 @@
-import type { GroupedEvents } from '$lib/internal/types.js';
+import type { GroupedEvents, MeltComponentEvents } from '$lib/internal/types.js';
 
 export const accordionEvents = {
 	trigger: ['keydown', 'click'] as const,
 };
 
 export type AccordionEvents = GroupedEvents<typeof accordionEvents>;
+export type AccordionComponentEvents = MeltComponentEvents<AccordionEvents>;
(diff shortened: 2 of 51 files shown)


TGlide (approved):
Awesome!!

Issue #373, opened 2023-08-11
🐛Bug: Menubar lagging


Lakelimbo:
**Describe the bug**
The [Menubar builder](https://www.melt-ui.com/docs/builders/menubar) is presenting a significant lag when clicking/touching. On mobile (or maybe touchscreen specifically) it seems like the focus closes the dropdown when alternating between menus.

**To Reproduce**
1. Build a Menubar, or just go to the one on the [docs](https://www.melt-ui.com/docs/builders/menubar)
2. Click on any of the menus
3. Notice the lag, and the weird behavior on touch

**Expected behavior**
The dropdowns should not lag, and the menu should not close this way on mobile/touch.

 - OS: Windows, Android, iOS (geerew on Discord)
 - Browser: Chrome, Edge, Firefox, Safari (geerew on Discord)

wesharper:
This is actually not unique to mobile, the desktop version is slow on hover as well

PR #377, opened 2023-08-11
[Menubar] Improve performance / Fix Lag

huntabyte:
Closes: #373

### DIFF
--- a/src/lib/builders/menu/create.ts
+++ b/src/lib/builders/menu/create.ts
@@ -1090,7 +1090,7 @@ export function createMenuBuilder(opts: _MenuBuilderOptions) {
 
 			const unsubs: Array<() => void> = [];
 
-			if (opts.selector !== 'menubar-menu' && $rootOpen && $preventScroll) {
+			if (opts.removeScroll && $rootOpen && $preventScroll) {
 				unsubs.push(removeScroll());
 			}
 
--- a/src/lib/builders/menu/create.ts
+++ b/src/lib/builders/menu/create.ts
@@ -1083,49 +1083,52 @@ export function createMenuBuilder(opts: _MenuBuilderOptions) {
 		}
 	});
 
-	effect([rootOpen, rootActiveTrigger], ([$rootOpen, $rootActiveTrigger]) => {
-		if (!isBrowser) return;
+	effect(
+		[rootOpen, rootActiveTrigger, preventScroll],
+		([$rootOpen, $rootActiveTrigger, $preventScroll]) => {
+			if (!isBrowser) return;
 
-		const unsubs: Array<() => void> = [];
+			const unsubs: Array<() => void> = [];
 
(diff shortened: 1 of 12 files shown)


src/lib/builders/menubar/create.ts
```diff
@@ -422,11 +444,35 @@ export function createMenubar(props?: CreateMenubarProps) {
...
+		}
+
+		// do something with remove scroll
+
+		return () => {
+			if (!get(activeMenu)) {
+				unsubs.forEach((unsub) => unsub());
+				scrollRemoved = false;
+			}
+		};
```
TGlide, line 464:
```suggestion
	effect([activeMenu], ([$activeMenu]) => {
		if (!isBrowser) return;
				
		if (!$activeMenu) {
			unsubs.forEach((unsub) => unsub());
			scrollRemoved = false;
		} else if (!scrollRemoved) {
			unsubs.push(removeScroll());
			scrollRemoved = true;
		}
	)};
```

Wouldn't this achieve the same result?

src/lib/builders/menubar/create.ts
```diff
@@ -71,15 +75,13 @@ export function createMenubar(props?: CreateMenubarProps) {
 		},
 		action: (node: HTMLElement) => {
 			const menuTriggers = Array.from(node.querySelectorAll('[data-melt-menubar-trigger]'));
-			if (!menuTriggers.length || !isHTMLElement(menuTriggers[0])) return {};
+			if (!isHTMLElement(menuTriggers[0])) return {};
 			menuTriggers[0].tabIndex = 0;
 
-			const menus = Array.from(node.querySelectorAll('[data-melt-menubar-menu]')).filter(
+			scopedMenus = Array.from(node.querySelectorAll('[data-melt-menubar-menu]')).filter(
```
TGlide, line 81: This only runs once, meaning if any menu is added or destroyed later on, scopedMenus value won't be accurate.

Either we constantly keep it updated by adding handlers on the trigger's action, or we instead create a function `getScopedMenus`, that works by getting the root menubarr via its id.

src/lib/builders/menu/create.ts
```diff
@@ -1083,49 +1083,52 @@ export function createMenuBuilder(opts: _MenuBuilderOptions) {
...
-			unsubs.push(removeScroll());
-		}
+			const unsubs: Array<() => void> = [];
 
-		if (!$rootOpen && $rootActiveTrigger) {
-			handleRovingFocus($rootActiveTrigger);
-		}
+			if (opts.selector !== 'menubar-menu' && $rootOpen && $preventScroll) {
+				unsubs.push(removeScroll());
+			}
```
TGlide, line 1095: Let's add a comment here explaining the menubar scroll is handled on their end. Alternatively, this could be an option of createMenu?

TGlide (changes requested):

TGlide (approved):

Issue #309, opened 2023-08-01
🐛Bug: Form Actions Dont Work Inside Context Menu


rondDev:
**Describe the bug**
Forms don't seem to work inside of a context menu.

**To Reproduce**

Create a [Context Menu](https://www.melt-ui.com/docs/builders/context-menu#usage).
Something like:
```html
<div style="width: 10rem; height: 10rem;" melt={$trigger}></div>
  <div class="menu" melt={$menu}>
    <form action="?/someaction" method="POST">
      <input type="hidden" name="file" value={"somevalue"} />
      <button melt={$item} type="submit">Trigger</button>
    </form>
</div>
```
Where you have the action `someaction` in +page.svelte.ts:
```ts
const someaction: Action = async ({ request }) => {
  doSomething();
};

export const actions: Actions = { remove };
```

This should in theory submit a form action, but doesn't seem to be doing anything.

**Expected behavior**
Form action should be submitted when button in context menu is pressed.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Firefox Developer Edition ( Version 116.0b8 (64-bit )) and Chrome ( Version 115.0.5790.110 )

[Template to recreate the issue on stackblitz](https://stackblitz.com/github/ronddev/melt-ui-context-menu/tree/main)
On Stackblitz an action is called when hitting the 'Delete' button, but doesn't get triggered when you try to use the context menu button even though they call the same form action.

huntabyte:
Sorry for taking so long to address this issue - this is 100% some sort of race condition. We most likely need to wait for a tick before taking action in the click event handler of the items.

Will do some tests and open a PR once I can resolve it.

PR #379, opened 2023-08-12
[Context Menu] Fix inability to submit forms

huntabyte:
Closes: #309

### DIFF
--- a/src/lib/builders/menu/create.ts
+++ b/src/lib/builders/menu/create.ts
@@ -332,7 +332,10 @@ export function createMenuBuilder(opts: _MenuBuilderOptions) {
 						handleRovingFocus(itemEl);
 						return;
 					}
-					rootOpen.set(false);
+					// Allows forms to submit before the menu is removed from the DOM
+					new Promise((resolve) => setTimeout(resolve)).then(() => {
+						rootOpen.set(false);
+					});
 				}),
 				addMeltEventListener(node, 'keydown', (e) => {
 					onItemKeyDown(e);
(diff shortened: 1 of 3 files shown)


src/lib/builders/menu/create.ts
```diff
@@ -332,7 +332,10 @@ export function createMenuBuilder(opts: _MenuBuilderOptions) {
 						handleRovingFocus(itemEl);
 						return;
 					}
-					rootOpen.set(false);
+					// Allows forms to submit before the menu is removed from the DOM
+					sleep(1).then(() => {
+						rootOpen.set(false);
```
TGlide, line 337: I wonder why this is needed? Dialog works without this, for example. If this is needed here, I'd assume it'd be needed anywhere we use a portal.

Also, devs should simply cancel the behaviour with event listeners?

huntabyte: Yeah but they don't necessarily want to cancel the behavior, they want the menu to close but they also want a form to submit. The dropdown menu works without this as well, which is strange because the context menu uses the same `item` as the dropdown, but the context doesn't,

huntabyte: More specifically, this is what I get in the browser console when trying to submit with:
image

Skeleton had a similar issue with their stepper which is where I got the idea to try this:
https://github.com/skeletonlabs/skeleton/pull/1333/files

huntabyte: Could also check the selector and only `sleep(1)` for the context menu if you see it being an issue!

TGlide: I see, let's leave it be then!

TGlide (approved):

PR #384, opened 2023-08-14
Clarified the warning in introduction.md

ivoilic:
Updated the under construction warning in the introduction based on this info: https://discord.com/channels/1116682155809067049/1140452054888873994

> Once version 1.0 is released, semantic versioning will be followed,  (limiting breaking changes to major releases). Until then, breaking changes might happen on non-major releases.

> ^ this - we're prerelease. We do bump minor/patch versions. So right now, minor could consist of a breaking change where patch changes should not.

This makes it a lot clearer where breaking changes might happens and what the next goalpost is.


TGlide:
Great idea. Thank you!

PR #385, opened 2023-08-14
[New Builder] Sortable

geerew:
A sortable builder.

## How it works

- There are zones which contain 1 or more items. 
- Zones have `pointerdown`, `pointerleave` and `pointerenter` events.
- Items can support a handle can be used, which can be anything that has `{...$handle}` under an zone item.
- Zones can accept items from all other zones, no other zones or selected zones.
- Zones and items can be disabled 
- Zones can be dropzones; accepts items but items cannot be sorted or selected (thus moved to other zones).
- When an item is selected (pointerdown), a clone of the element is created and follows the mouse. 
- As the mouse moves between items and zones, checks are performed to see if the pointer is intersecting an item. If it is logic is carried out and the selected item moves to its new position.
- When the ghost created, the events `pointermove` and `pointerup` are added to the document.
- A custom FLIP animation was created to handle animating items as they shift around DOM.

## Todo: 

Not all of these need implemented. More suggestions and all are up for discussion.

- [x] Documentation needs finished.
- [ ] Edge scroll. When you get to the edge scroll.
- [ ] Events.
  - While the pointer is held down and the selected item moves position send an event.
  - When the pointer is released, send a event with infomation about the final position.
- [x] Nested items. items within items ... itemception.
- [ ] Rework return to home.
  - When the pointer is released outside of a zone, the item will return to its origin.
  - This is implemented but with haste.
- [ ] Copy
  - Allow copying an item into a new zone instead of moving it. SortableJS has this.
  - Zonal prop or maybe an item prop .. undecided.
- [ ] Unique
  - Only allow adding an item to a new zone if it is unique (based upon item id).
  - A scenario for this is when copy is implemented, you drag an item into another zone and then drag it in again. You now have 2 versions of this item.
- [ ] Static zone. Reverse of a dropzone.
  - Items cannot be reorded within the zone. 
  - Items can be moved to another zone.
  - Items cannot be moved into this zone.
- [x] Item Group
  - At the minute items can be anywhere underneath a zone.
  - It would be good to have an (optional) element that can be a container of items.
  - When Items are move to another zone or items are moved into this zone, they are placed within the zone item container.


geerew:
@TGlide It is. There are a few things I would like to get in to make it more feature complete. However, I would still appreciate you having a look at the logic when you get time as the core of it is there and is unlikely to change.

TGlide:
> @TGlide It is. There are a few things I would like to get in to make it more feature complete. However, I would still appreciate you having a look at the logic when you get time as the core of it is there and is unlikely to change.

Of course! I'll take a look as soon as I can :)

wesharper:
What is the reasoning for zone item container vs just a zone? I worry that without strong justification it may make the API confusing to newcomers who might think that it's required.

wesharper:
Curious if some of these helpers are necessary for the builder. It feels like copy, unique, and even static zones are all things that are pretty use-case specific and that it might be better if the responsibility for implementation falls on the consumer of the builder.

wesharper:
Also curious if it might be convenient to expose something like a drop zone builder as its own primitive as part of this work, which could be useful for something like a file upload drop zone.

geerew:
> What is the reasoning for zone item container vs just a zone? I worry that without strong justification it may make the API confusing to newcomers who might think that it's required. I've run into this problem myself with some of the primitives. For instance, `portalled` on the dialog builder doesn't seem to be required, but it's ambiguous when it would be useful to have (perhaps the ambiguity could be resolved with a simple update to the docs).

Zones are required. Events are placed on the zone.  Items within zones are also required. 

Zones are like groups. Think a trello board or kanban board. You can move items between zones, for example from ‘not started’ to ‘working’ to ‘done’. Zones optionally accept items from other zones. You may not want to allow items from ‘not started’ to be moved into ‘done’

Within zones you have items. You could have 5 elements within a zone but only want 3 to be  sortable. Originally I was just setting a zone and all items within the zone were sortable however it became clear that this restricts design and I didn’t want to do that. So you explicitly say was is a sortable item within a zone. 

The docs will be clear on this fact.

geerew:
> Curious if some of these helpers are necessary for the builder. It feels like copy, unique, and even static zones are all things that are pretty use-case specific and that it might be better if the responsibility for implementation falls on the consumer of the builder.

You are correct that they have specific use cases. The idea was to bring this builder as close to sortable js as possible, which is one of the more popular sortable libraries. They are on the backburner for now and can be added down the road if wanted

wesharper:
> What you are referring to with ‘zone item container’ I ended up calling item groups. The idea is that within a zone you could have another element and within that element all your items. If you were to move all those items to another zone and then move 1 back, the item doesn’t move back into the element. Instead it becomes a child of the overall zone element. To combat this you can have groups within zones. When you move items into a zone, if you have no groups the items is simply added as a child of the zone. If you have a group (or more groups within the 1 zone) then the item is added to the group it was moved into.

@geerew is this essentially the same as a zone within a zone or am I thinking about it wrong? I understand that having a sub-zone that could be indicative of a distinct sub-grouping could be useful, but to me this is achievable with either:
1. 2 zones that are styled to look like part of the same parent but functionally are just 2 separate zones from which the end user can decide between.
2. A parent zone with 2 or more child zones, any of which the end user can drag an item into.

I'm just trying to understand it a bit better. I'll pull the branch and look at the docs though to see if it answers my questions.

geerew:
@wesharper 

sortable is an extremely complex builder with a lot of logic going on behind-the-scenes. My two main goals are to not restrict developer experience and speed. 

Zones contain the majority of the properties as well as all events. When you select an item and begin dragging, only the items within the current zone are checked for an intersection/hit. If the pointer leaves the zone no intersection checks are carried out. If you enter a new zone, intersection checks begin again for the items in this zone.

Item groups are purely a developer experience thing to make your life easier.

raymonddaikon:
Looking forward to this! Is this still in draft mode or pretty much ready to go?

geerew:
@raymonddaikon its pretty much there. Probably a few bugs as it’s complex and might need a few fix ups following a review.

There are 2 items that need done to make it fully useable. 

1. Edge scroll. This is less important but a nice feature to have. 
2. Events. This is extremely important. I was thinking we expose 3 override functions, onStart, onChange and onFinish. 

onStart fires as when an item is selected and starts moving. It will return information about the item selected such as the element, zone it’s in, parent element (if there is one), etc

onChange fires when the selected item intersects another item and a swap occurs AND while the selected item continues to be selected. It will return information about the selected element, old zone, new zone (might be the same as old zone), element it swapped with, old index position, new index position, if nesting is enabled the old parent and new parent (might be the same), whatever else seems important

onFinish is called when the selected item is released and will send back similar information as onChange


The most important thing here is the documentation needs to make it abundantly clear that this information should NOT be used to update the elements yourself. The sortable builder handles all this. These functions allow you to do things like add styles update db information or even update an array in your compost (so long as changing this array will not cause a page render if the zone elements

@TGlide unfortunately I won’t have much more time to work on this for the foreseeable. Work has increased and I’ve other commitments. The code is in a good state and ready for a review. I’m more than happy to guide with what needs done but I’ve taken it as far as I can for now. Apologies

geerew:
Oh I also recommend we change the name of ‘drop zone’ feature. It’s confusing as it’s not a real drop zone. It’s just another sortable ‘zone’ with different conditions. In a normal zone you can rearrange items. Also items from other zones can be dragged in and moved to any position with the zone

In a ‘drop zone’ items are static and items dragged in from other zone append to the end. This also means items can other be dragged out. You would have to implement a delete feature to remove items from this zone. 

Maybe a better name is ‘static’ zone.

the idea for this feature was more to do with say tags or list of items. You have a zone that contains a bunch of sortable items and another static zone that can receive items but cannot be sorted.

TGlide:
Closing due to inactivity/organization. Would be nice to revisit this though in the future, as this is amazing work!

geerew:
@austerj I took inspiration from SortableJS but in a svelte/melt-ui way. SortableJS is great but has some quirks which my version didnt have. 

Unfortunately I got super busy in work/life and wasn’t able to continue (for now)

saturnonearth:
> @austerj I took inspiration from SortableJS but in a svelte/melt-ui way. SortableJS is great but has some quirks which my version didnt have.
> 
> Unfortunately I got super busy in work/life and wasn’t able to continue (for now)

No worries!! If anyone needs this functionality look out for https://github.com/atlassian/pragmatic-drag-and-drop

PR #399, opened 2023-08-17
stop touching hover card button from immediately clicking the button

### LINKED ISSUE #398: 🐛Bug: touching button of hover card immediately triggers a click
**Describe the bug**

When using a touch interface, touching the button attached to a hover card immediately visits the link. This is not what would normally happen, usually you tap and release. This means that trying to, for example, scroll down the page you may accidentally 'click' the button just be swiping over it.

**To Reproduce**
1. Go to 'https://www.melt-ui.com/docs/builders/hover-card'
2. Open dev tools and go into touch emulation mode
3. Try to scroll the page whilst starting on the hover card button.
4. The link the button points to is opened

**Expected behavior**
The button should behave as it would without a hover card - it should not count a click unless a click actually occurs.

**Screenshots**

https://github.com/melt-ui/melt-ui/assets/3743418/659861fb-6d77-40db-81ac-e148e64073e7

 - OS: Mac OS 13
 - Browser: Arc (basically chrome) with mobile emulation on in dev tools
 - Version: 1.3.0

 - Device: iPhone 12
 - OS: iOS 16.6
 - Browser: Safari
 - Version: 16.6?

jamesbirtles:
fixes #398 

So... I appreciate this might not be the way you want to fix this bug, but referring to the documentation:
CleanShot 2023-08-17 at 16 15 08@2x

Reading this, I'm not totally sure opening the card on focus is actually appropriate, given that it is specifically designed to only open with a pointing device, calling out keyboard as something that isn't supported.
Also if we take github's hover card as an example (e.g. hovering an issue link) they do not trigger on focus

So this PR removes the focus handling, and therefore is able to remove the hack attempts to prevent focus on mobile via touch, therefore fixing the bug!

### DIFF
--- a/.changeset/healthy-lizards-sparkle.md
+++ b/.changeset/healthy-lizards-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+Remove focus trigger on hover cards
--- a/src/lib/builders/hover-card/create.ts
+++ b/src/lib/builders/hover-card/create.ts
@@ -123,16 +123,6 @@ export function createHoverCard(props: CreateHoverCardProps = {}) {
 				addMeltEventListener(node, 'pointerleave', (e) => {
 					if (isTouch(e)) return;
 					get(handleClose)();
-				}),
-				addMeltEventListener(node, 'focus', () => get(handleOpen)()),
-
-				addMeltEventListener(node, 'blur', () => get(handleClose)()),
-				addMeltEventListener(node, 'touchstart', (e) => {
-					// prevent focus on touch devices
-					e.preventDefault();
-					const currentTarget = e.currentTarget;
-					if (!isHTMLElement(currentTarget)) return;
-					currentTarget.click();
 				})
 			);
 


TGlide:
Hey @jamesbirtles thanks for this!

To be honest I think we should approach this differently. It should work on focus, but we should provide an option to disable that behaviour. To be honest, we already do, with custom event listeners, but we need to document that behaviour.

Me and @huntabyte discussed a bit about this issue, and we noticed a few other oddities!

One questioning that came about is, why have both HoverCard and Tooltip? Well, HoverCard is not readable by screen readers, and can have interactible content (tooltip shouldn't). 

Another one is, HoverCard shouldn't contain important info, specially since it isn't tabbable from the trigger. So I think we should make that more obvious.

Here's what I propose:
- Change HoverCard's name to LinkPreview
- Document how to cancel focus behaviour, and modify previous callout that stated it didn't work on focus
- Document that LinkPreview should only contain non-vital info
- Fix the bug you mentioned

What do you think? Is this something you'd like to tackle?

jamesbirtles:
> What do you think? Is this something you'd like to tackle?

Absolutely!

> To be honest, we already do, with custom event listeners, but we need to document that behaviour.

Is that the custom events mentioned here? https://www.melt-ui.com/docs/controlled

> Document how to cancel focus behaviour

Would this just be documenting the above events on the LinkPreview page itself?

jamesbirtles:
Looking into custom events, i'm not sure they quite work as expected/desired with regard to cancelling them. The documentation states `e.detail.cancel()` will cancel the original event and stop melt ui's event handler from running too. The latter part of that doesn't seem to actually work, at least for the touchstart/focus/blur events on the hover card trigger. But also, to work around this bug for example, I don't want to prevent the browser default behaviour I want to just cancel melt-ui's behaviour. In theory this could be achieved with cancelling the custom event itself (`e.preventDefault()`) though this isn't currently implemented. Another problem is that not all browser events are cancellable such as focus, meaning it is just not possible to cancel it or melt ui's handler.

TGlide:
> Absolutely!

Awesome, thank you so much! 🎉 

> Is that the custom events mentioned here? [melt-ui.com/docs/controlled](https://www.melt-ui.com/docs/controlled)

Yes!

> Would this just be documenting the above events on the LinkPreview page itself?

Not all of them, just mentioning the focus, since some people may find it annoying that tabs opens it up. I'd add a small code-snippet in the usage section with an explanation

TGlide:
> Looking into custom events, i'm not sure they quite work as expected/desired with regard to cancelling them. The documentation states `e.detail.cancel()` will cancel the original event and stop melt ui's event handler from running too. The latter part of that doesn't seem to actually work, at least for the focus/blur events on the hover card trigger. But also, to work around this bug for example, I don't want to prevent the browser default behaviour I want to just cancel melt-ui's behaviour. In theory this could be achieved with cancelling the custom event itself (`e.preventDefault()`) though this isn't currently implemented. Another problem is that not all browser events are cancellable such as focus, meaning it is just not possible to cancel it or melt ui's handler.

Oh, I had implemented a workaround for the non-cancelable events. I seem to have forgotten to merge it, 1sec.

And you're correct, we shouldn't be cancelling the event entirely!

TGlide:
@jamesbirtles done, now it should be much better 😄

jamesbirtles:
works great now, thanks!

Issue #389, opened 2023-08-15
🌟Feature request: Ability to keep the Select menu open after selecting


Tyneor:
**Is your feature request related to a problem? Please describe.**
In this https://github.com/melt-ui/melt-ui/issues/93#issuecomment-1627823637, you mentionned an "option to keep the select open while inputting contents would be ideal."

**Describe the solution you'd like**
Either a keepOpenOnSelect option,
Or a check before closing the menu after selecting an option to ensure that the dropdown is not multiselectable (here i suppose: https://github.com/melt-ui/melt-ui/blob/0a77ae9907ded9fbf7501cb324849b5bce2bb555/src/lib/builders/select/create.ts#L515).

**Describe alternatives you've considered**
Currently, I overwrite the click event like so :
```html
<li
  use:melt={$option({ value: item, label: item })}
  on:m-click={(e) => {
    if (multiple && (Array.isArray($value) || $value === undefined)) {
      e.detail.cancel();
      $value = toggle(item, $value ?? []);
    }
  }}
>
    <Checkbox checked={$isSelected(item)}>{item}</Checkbox>
</li>
```

If it is possible, I would be happy to work on it :)

huntabyte:
I know @TGlide was recently working on the Select, but if he's done, I think what you described should be the expected behavior for multi-selects if you want to go for it!

TGlide:
@Tyneor I'm going to submit a small refactor on the Select today, and then maybe afterwards would be the perfect time for you to work on it. If you need any help, lmk! 😄

TGlide:
@Tyneor done!

PR #401, opened 2023-08-17
feat(#389) after selecting, check multiple is false before closing select menu

Tyneor:
Updated related test accordingly.

Since this is an expected behaviour I did not changed the docs. 

Side note: I tested the Esc Key to close the Select menu, that may not be necessary or a separate test


huntabyte (approved):
Thanks for updating the tests as well!

TGlide (approved):
Awesome, thank you so much! 🎉

PR #406, opened 2023-08-18
fix: context menu not closing when clicking on svg element

### LINKED ISSUE #405: 🐛Bug: Context menu not closing when clicking on svg element
**Describe the bug**
If you click on a SVG element (for example an icon) outside of a context menu, the context menu does not close.

The issue lies in the `isHTMLElement` check in the `handleClickOutside` function:

https://github.com/melt-ui/melt-ui/blob/0ff5341f24fdbf0489017a0f33c86a8b750a6081/src/lib/builders/context-menu/create.ts#L111

I replaced this line with `if (!(target instanceof Element)) return;` and it works as expected.

**To Reproduce**
1. Open the context menu example page (https://www.melt-ui.com/docs/builders/context-menu)
2. Open the context menu
3. Click on a svg icon (for example a checkmark in the features section)
4. Context menu stays open

Theta-Dev:
fixes #405

### DIFF
--- a/src/lib/builders/context-menu/create.ts
+++ b/src/lib/builders/context-menu/create.ts
@@ -108,7 +108,7 @@ export function createContextMenu(props?: CreateContextMenuProps) {
 		if (e.defaultPrevented) return;
 
 		const target = e.target;
-		if (!isHTMLElement(target)) return;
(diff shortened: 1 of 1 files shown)


TGlide (approved):
Awesome, thank you!

PR #408, opened 2023-08-18
Link Preview

jamesbirtles:
As discussed in #399 heres a list of things to do
- [x] Change HoverCard's name to LinkPreview
- [x] Document how to cancel focus behaviour, and modify previous callout that stated it didn't work on focus
- [x] Document that LinkPreview should only contain non-vital info
- [x] Fix the bug #398

let me know if you'd rather I do them in separate PRs or all in this one.


TGlide:
> let me know if you'd rather I do them in separate PRs or all in this one.

You can do them all in one!

jamesbirtles:
I think this is ready now. My fix for the bug was to only show the link preview on focuses that would be visible (they apply the `:focus-visible` pseudo-class). As far as I can tell this deals with the issue that the touchstart handling was added to fix allowing me to remove it, and additionally means that returning to the tab after clicking the link no longer erroneously opens the preview.

PR #427, opened 2023-08-21
docs: more exhaustive list of component libraries

benmccann:
users can add to this list by sending a PR to https://github.com/svelte-society/sveltesociety.dev/ as described on https://sveltesociety.dev/help/submitting?type=component

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,4 +115,5 @@ we'd love to have you there.
 
 ## Similar projects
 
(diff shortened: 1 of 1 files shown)


TGlide:
Thank you for the PR! That's a nice idea. Maybe we should have a built with melt UI section to include libraries and projects that want to be showcased. But this is great for now!

Issue #419, opened 2023-08-19
🐛Bug: body padding-right is not taken into account for builders with preventScroll set to true


Tyneor:
**Describe the bug**
I was using the select builder in Storybook, which applies a 1rem padding to the iframe body, and noticed that the page changed size while the select menu was open.
Upon further inspection, it seems that `body` `padding-right` is overridden in removeScroll to account for the scrollbar width.

**To Reproduce**
1. Go to https://www.melt-ui.com/docs/builders/select
2. Set body padding to 1rem
3. Open and close the select menu
4. Page padding changes back and forth.

**Expected behavior**
Preserve `body` `padding-right` even when the select menu is open.

PR #429, opened 2023-08-21
fix(#419) preserve body padding in removeScroll

Tyneor:
Should solve issue #419.
I used a `calc()` instead of simply adding the values because the body padding could be in any unit.


TGlide:
Awesome, thank you so much!

PR #430, opened 2023-08-21
Upgrade to Storybook 7.3 and remove unplugin-icons

benmccann:
Storybook gave me a warning that postcss is built into Vite so you should not use the `postCss` option when using Vite

https://github.com/antfu/unplugin-icons/pull/293 updated the recommended unplugin-icons type import setup


TGlide:
Thank you for the PR! We're no longer using unplugin-icons, using lucide-svelte instead, so I think we can remove it from the project.

PR #435, opened 2023-08-22
fix: Select & Combobox `forceVisible` transition bug

huntabyte:
Conditionally modifying when the `open.set()` is called `onMount` depending on if `forceVisible` is true or not.

Without this, when using forceVisible without a transition, the select breaks and does not allow the selection of items.


TGlide:
Closing since this was fixed in another commit!

Issue #390, opened 2023-08-15
🐛"Bug": disabled="false" should not be applied


fabian-michael:
**Describe the bug**
Not really a bug but CSS libraries like daisyUI solely check if the disabled attribute is apparent in order to apply disabled styles so disabled="false" will enable the selector. According to the HTML specification "false" isn't a valid value for disabled either.

**Expected behavior**
No disabled attribute should be applied if disabled is false. 

[StackBlitz](https://stackblitz.com/edit/vitejs-vite-ab5wjy?file=src%2Flib%2FListbox.svelte)

fabian-michael:
aria-disabled="true/false" could be applied instead and the disabled attribute only if disabled is true

huntabyte:
Thanks for pointing this out! Definitely something we'll address!

PR #439, opened 2023-08-22
Cleanup disabled attributes

huntabyte:
Closes: #390

Standardizes the way we apply `data-disabled` properties to various builder elements. Right now we have a mix of `[data-disabled]` `[data-disabled="true"]` `[data-disabled="false"]`.

Made them all consistent where `[data-disabled]` is only applied if the element is in fact disabled.

Additionally, only apply the `disabled` attribute if the element is disabled, otherwise, it's undefined.

### DIFF
--- a/.changeset/dull-balloons-teach.md
+++ b/.changeset/dull-balloons-teach.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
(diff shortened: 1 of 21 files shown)


TGlide:
> Additionally, only apply the disabled attribute if the element is disabled, otherwise, it's undefined.
I think Svelte deals with boolean disabled values automatically

PR #440, opened 2023-08-22
update example styles

huntabyte:
Updates some of the example styles to align with the new styles demonstrated throughout the docs.


TGlide:
Awesome, I think we just need some focus-visible styles in the pagination example

Issue #434, opened 2023-08-22
🐛Bug: Wrong props passed to radio input


Stefano-Labianca:
**Describe the bug**
When I use the RadioGroup builder, `use:melt={$itemInput(id)}` doesn't pass the right props to the `<input>` element.

**To Reproduce**
1. Copy the example code from the [docs](https://www.melt-ui.com/docs/builders/radio-group)
2. When calling `createRadioGroup`, specify the `required` option
3. Add a custom `<input>` element. Like this `<input use:melt={$itemInput(id)}>`
4. Wrap the component with `<form>` element and submit it

**Expected behavior**
When the user does not select any element, the form should focus the required radio group, but the form gets submitted without any error.

**Screenshots**
![image](https://github.com/melt-ui/melt-ui/assets/58400415/af7e6066-20c6-4132-8885-dab65c4b7081)

![image](https://github.com/melt-ui/melt-ui/assets/58400415/2d28be23-add5-4d2d-94b2-b0fd912c304d)

**Desktop (please complete the following information):**
 - OS: Debian WSL
 - Browser chrome
 - Version @melt-ui/svelte 0.37.5

**Additional context**
The input element should probably be of type `radio` instead of `hidden` (At least I suppose so).

TGlide:
@Stefano-Labianca Please fill the issue description

Stefano-Labianca:
> @Stefano-Labianca Please fill the issue description

My bad, I miss-clicked

huntabyte:
@TGlide was this issue solved with the PR you merged?

TGlide:
> @TGlide was this issue solved with the PR you merged?

Yes! @Stefano-Labianca I'll be closing this, but if you still have issues, let us know.

Issue #394, opened 2023-08-16
⚙Tech Debt: Add name prop to createRadioGroup


TGlide:

huntabyte:
Hey @TGlide did your PR address this issue?

TGlide:
Nope!

huntabyte:
The `name` prop would be used for the `hiddenInput` for use with form submission.

PR #441, opened 2023-08-22
[RadioGroup] - Fix input `required` & `disabled` & add `name` prop to builder

huntabyte:
Fixes: #434
Closes: #394 

Applying the `required` and `disabled` attributes in the return object of the store like we do other attributes does not work properly.

The fix I came up with is setting `data-disabled` & `data-required` attributes on the input on SSR, and then when the action runs, checking for those attributes to properly set the `node.disabled` and `node.required`. Additionally, using an `effect` we're watching to see if the `disabled` or `required` option stores are updated and then updates the input accordingly.

This is likely the case with the other builder inputs, so I will adjust those in #439 which I've already started to handle the `disabled` attributes.

Additionally, added a `name` option to the radio group builder to apply to the inputs for form submission.

### DIFF
--- a/.changeset/tough-insects-sip.md
+++ b/.changeset/tough-insects-sip.md
@@ -1,0 +1,6 @@
+---
+"@melt-ui/svelte": patch
+---
+
+- Properly apply `required` and `disabled` attributes on `itemInput`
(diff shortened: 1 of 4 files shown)


TGlide:
Closing since it was resolved in another commit 🎉

Issue #428, opened 2023-08-21
⚙Tech Debt: migrate from `eslint-plugin-svelte3` to `eslint-plugin-svelte`


benmccann:
`eslint-plugin-svelte3` is the old plugin and has been replaced with `eslint-plugin-svelte`

PR #442, opened 2023-08-22
[Select] Fix form bug

huntabyte:
Since we weren't specifying `type="button"` for the select trigger, it was catching the click event of the form, causing all sorts of weird issues.

Closes: #410

PR #443, opened 2023-08-23
refactor: migrate from `eslint-plugin-svelte3` to `eslint-plugin-svelte`

mcmxcdev:
## Changes
- install `eslint-plugin-svelte` dependency
- remove `eslint-plugin-svelte3` dependency
- modernize eslint config
- configure `@typescript-eslint/no-unused-vars` properly to ignore underscore-prefixed vars and args
- opt in to additional `eslint-plugin-svelte` rules which are not part of the recommended config
- fix or disable reported lint issues

## Motivation
Close https://github.com/melt-ui/melt-ui/issues/428

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,34 +1,60 @@
+/** @type {import('eslint').Linter.Config} */
 module.exports = {
 	root: true,
 	parser: '@typescript-eslint/parser',
 	extends: [
 		'eslint:recommended',
 		'plugin:@typescript-eslint/recommended',
+		'plugin:svelte/recommended',
 		'plugin:storybook/recommended',
 		'prettier',
 	],
-	plugins: ['svelte3', '@typescript-eslint'],
-	ignorePatterns: ['*.cjs'],
+	plugins: ['@typescript-eslint'],
 	overrides: [
 		{
 			files: ['*.svelte'],
-			processor: 'svelte3/svelte3',
+			parser: 'svelte-eslint-parser',
+			parserOptions: {
+				parser: '@typescript-eslint/parser',
+			},
 		},
 	],
-	settings: {
-		'svelte3/typescript': () => require('typescript'),
-	},
 	parserOptions: {
 		sourceType: 'module',
-		ecmaVersion: 2020,
+		ecmaVersion: 'latest',
+		extraFileExtensions: ['.svelte'],
 	},
 	env: {
 		browser: true,
-		es2017: true,
+		es2024: true,
 		node: true,
 	},
 	rules: {
+		// eslint
 		'no-console': 'warn',
+
+		// @typescript-eslint
+		'@typescript-eslint/no-unused-vars': [
+			'warn',
+			{
+				argsIgnorePattern: '^_',
+				varsIgnorePattern: '^_',
+			},
+		],
+
+		// eslint-plugin-svelte
+		'svelte/no-target-blank': 'error',
+		'svelte/no-immutable-reactive-statements': 'error',
+		'svelte/no-trailing-spaces': 'error',
+		'svelte/prefer-style-directive': 'error',
+		'svelte/no-reactive-literals': 'error',
+		'svelte/no-useless-mustaches': 'error',
+		// TODO: opt in to these at a later stage
+		'svelte/button-has-type': 'off',
+		'svelte/require-each-key': 'off',
+		'svelte/no-at-html-tags': 'off',
+		'svelte/no-unused-svelte-ignore': 'off',
+		'svelte/require-stores-init': 'off',
 	},
 	globals: {
 		NodeJS: true,
--- a/src/docs/components/site-header.svelte
+++ b/src/docs/components/site-header.svelte
@@ -40,7 +40,7 @@
 			<a
 				href={siteConfig.links.discord}
 				target="_blank"
-				rel="noreferrer"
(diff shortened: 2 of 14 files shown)


.eslintrc.cjs
```diff
@@ -1,34 +1,60 @@
...
+		'svelte/no-trailing-spaces': 'error',
+		'svelte/prefer-style-directive': 'error',
+		'svelte/no-reactive-literals': 'error',
+		'svelte/no-useless-mustaches': 'error',
+		// TODO: opt in to these at a later stage
+		'svelte/button-has-type': 'off',
+		'svelte/require-each-key': 'off',
+		'svelte/no-at-html-tags': 'off',
+		'svelte/no-unused-svelte-ignore': 'off',
+		'svelte/require-stores-init': 'off',
```
mcmxcdev, line 57: These are additional helpful lint rules from eslint-plugin-svelte that I would recommend, but they do produce a lot of new lint issues, so should be adopted incrementally.

https://sveltejs.github.io/eslint-plugin-svelte/rules/button-has-type/ would have been very useful to have enabled from the start to avoid major bug like: https://github.com/melt-ui/melt-ui/pull/442

src/docs/components/site-header.svelte
```diff
@@ -40,7 +40,7 @@
 			<a
 				href={siteConfig.links.discord}
 				target="_blank"
-				rel="noreferrer"
+				rel="noopener noreferrer"
```
mcmxcdev, line 43: If I am not mistaken, `noreferrer` alone would actually be enough, but `eslint-plugin-svelte` wants `noopener` to be specified too.

src/routes/docs/builders/[name]/+page.svelte
```diff
@@ -12,11 +12,13 @@
 
 	export let data: PageData;
 
+	// eslint-disable-next-line no-undef
 	type Component = $$Generic<typeof SvelteComponent>;
 	$: component = data.doc.default as unknown as Component;
 	$: doc = data.doc.metadata;
 	$: snippets = data.snippets;
 	$: mainPreview = data.mainPreview as unknown as Component;
+	// eslint-disable-next-line svelte/no-reactive-literals, svelte/no-immutable-reactive-statements
```
Comment (mcmxcdev) [inline src/routes/docs/builders/[name]/+page.svelte:21]: `viewCode` below is marked as reactive, but it doesn't react to any variables and is a simple boolean. Can be solved whenever.

src/docs/components/site-header.svelte
benmccann: maybe the plugin wasn't updated after https://github.com/sveltejs/svelte/pull/8230 was merged. might want to file an issue with `eslint-plugin-svelte`

src/routes/docs/builders/[name]/+page.svelte
Comment (TGlide) [reply src/routes/docs/builders/[name]/+page.svelte:21]: You can change it to a let, no problem!

src/docs/components/site-header.svelte
TGlide: I'm fine with it 😄

.eslintrc.cjs
TGlide: Nice, thanks for the suggestion

src/docs/components/site-header.svelte
benmccann: https://github.com/sveltejs/eslint-plugin-svelte/issues/571

src/routes/docs/builders/[name]/+page.svelte
Comment (mcmxcdev) [reply src/routes/docs/builders/[name]/+page.svelte]: Updated it in [`adb2d53` (#443)](https://github.com/melt-ui/melt-ui/pull/443/commits/adb2d5316e38eb25d9378e6ecc271d582b2d46d3)

.eslintrc.cjs
```diff
@@ -1,34 +1,60 @@
...
+		'@typescript-eslint/no-unused-vars': [
+			'warn',
+			{
+				argsIgnorePattern: '^_',
+				varsIgnorePattern: '^_',
+			},
+		],
+
+		// eslint-plugin-svelte
+		'svelte/no-target-blank': 'error',
```
benmccann, line 46: this is the one forcing us to add `noopener`. perhaps we should remove it. since we're not compiling the docs in legacy mode it serves no purpose
```suggestion
```

mcmxcdev: Shouldn't we just update the rule configuration to https://sveltejs.github.io/eslint-plugin-svelte/rules/no-target-blank/#allow-referrer-true here to make the rule require rel="noopener" but not noreferrer?

TGlide: It should be the other way around though, requiring `noreferrer` but not `noopener`.

mcmxcdev: I agree: https://stackoverflow.com/questions/57628890/why-people-use-rel-noopener-noreferrer-instead-of-just-rel-noreferrer

Then that rule option makes no sense, how should we proceed?

benmccann: I posted in `eslint-plugin-svelte` to see if we can address it there: https://github.com/sveltejs/eslint-plugin-svelte/issues/571#issuecomment-1691839668

TGlide: For now I'll merge this PR and we can update the plugin once its fixed

TGlide (approved):
Thank you!!! 🎉

PR #446, opened 2023-08-24
Passing custom `regex` to Tags-Input to validate the input

anhtuank7c:
Add an option for `regex` to validate tag.


geerew:
We have an add function that can be used to validate input. 

It’s up to @TGlide but I think it’s overkill when this regex can just be called from inside the add function.

anhtuank7c:
I have discussed with Thomas in the Discord channel, understood the usage purpose of validating in `add` function.
I will close this PR.
Thanks for review it @geerew 
Have a good day.

Issue #451, opened 2023-08-24
🐛Bug: Select inside dialog causes dialog to close when an item is selected


buhodev:
**Describe the bug**

Selecting an item from a Select component inside a dialog causes the dialog to close (I suppose due to the way the outsideClick logic is handled). This also happens with any element outside the dialog (e.g. an always on display bottom navbar with a hamburguer button that opens the dialog (and changes to a X button to close it when the dialog is open).

**To Reproduce**
1. Go to https://www.sveltelab.dev/2kezbln53cyozjz
2. Click on 'Open Dialog' button
3. Select a flavor from the Select inside the Dialog
4. See error (dialog closes)

**Expected behavior**
Dialog _shouldn't_ close when selecting a item from a Select component inside the dialog

TGlide:
Thanks for raising this issue! @buhodev since there are two bugs, please create a separate issue for the second bug 😄

buhodev:
> Thanks for raising this issue! @buhodev since there are two bugs, please create a separate issue for the second bug 😄

Done https://github.com/melt-ui/melt-ui/issues/452 :wink:

softwarecurator:
https://stackblitz.com/edit/kit-ts-tw-melt-mclbpf?file=package.json seems to be still an issue on `0.39.0`

TGlide:
> https://stackblitz.com/edit/kit-ts-tw-melt-mclbpf?file=package.json seems to be still an issue on `0.39.0`

0.39.0 does not try to address this issue, this is expected.

huntabyte:
The same happens inside of a `Popover`. It appears the select isn't respecting any portal parent's and is always portalling to the body, even when `portal: null`

thomasefbland:
> The same happens inside of a `Popover`. It appears the select isn't respecting any portal parent's and is always portalling to the body, even when `portal: null`

Interestingly though, when you "fix" this by providing portal: selector, to the dialog portal element, it now nests inside correctly, yet the closing on select issue persists

PR #462, opened 2023-08-27
Fix: select nested portal bug

huntabyte:
Closes: #451

We had an extra portal action in there causing the portal to always be set to the body, regardless of the `portal` prop or parent portal situation. Removing the 2nd portal action solves the issue and it now works as expected.


huntabyte:
Lint failing due to a previous merge to main with the tag input formatting.

TGlide (approved):

PR #464, opened 2023-08-27
fix: Combobox isSelected function

bennymi:
Hey, I noticed that the `isSelected` function is currently always returning `false` since we're comparing two non-primitive object types. I changed the function to use `JSON.stringify()` instead which fixes that. Now the checkmark also shows up in the preview.

### DIFF
--- a/.changeset/fair-bags-try.md
+++ b/.changeset/fair-bags-try.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+fix: Combobox isSelected function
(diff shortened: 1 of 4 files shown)


TGlide:
Thanks!! We have a deep equal function that can be used instead

bennymi:
@TGlide great, I didn't know about that. I'll update it in a bit.

bennymi:
@TGlide updated 😄

TGlide (approved):
Awesome, thank you! 🎉

Issue #461, opened 2023-08-26
⚙Tech Debt: remove `deep-equal` as dependency


benmccann:
`deep-equal` has over 40 dependencies if I recall correctly. Also, `fast-deep-equal` and `dequal` are already in the dependency tree, so I'd recommend switching to one of those libraries

PR #471, opened 2023-08-29
refactor: migrate from deep-equal to dequal

mcmxcdev:
## Changes
- remove `deep-equal` and `@types/deep-equal` dependencies
- install `dequal` dependency
- replace usages of `deep-equal` with `dequal` in combobox `create.ts` and `array.ts` helper

## Motivation

I had a look at https://npmtrends.com/deep-equal-vs-dequal-vs-fast-deep-equal and compared the different options through:

- https://arve0.github.io/npm-download-size/#deep-equal
- https://arve0.github.io/npm-download-size/#fast-deep-equal
- https://arve0.github.io/npm-download-size/#dequal

Looking at the lockfile, I found more dependencies using `dequal` rather than `fast-deep-equal` under the hood, so I went with that. The lib is also much newer and had its latest commit in 2022. 

Also I remember lukeed (author of `dequal`) being part of the Svelte community, so there might be some overlap and an easier path forward to reach him in case of bugs etc.

Closes https://github.com/melt-ui/melt-ui/issues/461

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
 	"dependencies": {
 		"@floating-ui/core": "^1.3.1",
 		"@floating-ui/dom": "^1.4.5",
-		"deep-equal": "^2.2.2",
+		"dequal": "^2.0.3",
 		"focus-trap": "^7.5.2",
 		"nanoid": "^4.0.2"
 	},
@@ -89,7 +89,6 @@
 		"@testing-library/jest-dom": "^5.17.0",
 		"@testing-library/svelte": "^4.0.3",
 		"@testing-library/user-event": "^14.4.3",
-		"@types/deep-equal": "^1.0.1",
 		"@types/jest-axe": "^3.5.5",
 		"@types/prettier": "^2.7.3",
 		"@types/testing-library__jest-dom": "^5.14.9",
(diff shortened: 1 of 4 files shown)


mcmxcdev:
Looking at the CI failures, it seems that https://github.com/melt-ui/melt-ui/actions/runs/6018311020/job/16326259233?pr=471 is definitely not related to my PR and should be handled separately.

Not sure about https://github.com/melt-ui/melt-ui/actions/runs/6018311020/job/16326259086?pr=471, I guess I will wait for PR feedback.

TGlide:
@mcmxcdev thank you for the PR!

About the tests, the Portal test is a bit flakey. I'll merge this and continue testing to see if it still occurs.

It does worry me that dequal hasn't been updated in over a year, but it has zero deps, and is only a single file with an MIT license, so worse case scenario we just copy it over and do modifications on it.

mcmxcdev:
> It does worry me that dequal hasn't been updated in over a year, but it has zero deps, and is only a single file with an MIT license, so worse case scenario we just copy it over and do modifications on it.

I agree with you, but I think it's not a big deal for the same reasons you mentioned!

PR #485, opened 2023-09-03
[Table of Contents]: add mutation observer

### LINKED ISSUE #450: 🌟Feature request: Include the heading node in TableOfContentsItem
**Is your feature request related to a problem? Please describe.**
I want to customise the way `TableOfContentsItem` is rendered, having the possibility of adding icons and deciding wether the hierarchy is expanded/collapsed by default.

**Describe the solution you'd like**
Passing the whole heading node to `TableOfContentsItem` will provide maximum flexibility to the way we render the elements.

```
export type TableOfContentsItem = {
    title: string;
    index: number;
    id: string;
    node: HTMLHeadingElement; <- this
    children?: TableOfContentsItem[];
};
```

```
<h2 data-icon="house" data-collapsed>Hello</h2>
```
I can then read the properties I care about from the node and decide how to render/style them.

**Describe alternatives you've considered**
- It could also be possible to only pass a whitelisted set of properties to it
```
createTableOfContents({ extraPropertiesPrefix: 'data-toc-', ... })
```
The type will be:
```
export type TableOfContentsItem = {
    title: string;
    index: number;
    id: string;
    extraProperties: Record<string, string>
    children?: TableOfContentsItem[];
};
```
```
<h2 data-toc-icon="house" data-toc-collapsed>Hello</h2>
```
- Also, since I have access to the id of the element, I could read those properties myself without having to change anything on the library, but I think this will be helpful for others.

**Additional context**

Captura de pantalla 2023-08-24 a les 15 33 57

### LINKED ISSUE #476: 🌟Feature request: Make Table of Contents reactive so that it changes if the page content does
**Is your feature request related to a problem? Please describe.**

Currently the Table of Contents does not update when the page content updates. When the ToC is in a `layout` file we can just wrap it with a `#key` block, but when changing content within a page which makes headers appear or disappear it causes problems, because instead of the new headings appearing we still have the old ones, which stop working on click of course because the element they belong to has been deleted.

**Describe the solution you'd like**

We could add a change event listener to the selector element and update the headings list if there are changes to the headings.

**Describe alternatives you've considered**

It could be possible to use stores within pages that feed back to the layout and refresh the ToC there, but that is unintuitive and not very elegant.

bennymi:
**Changes:**
- add a mutation observer that checks if new headings were added or removed and updates the headings list if yes
- add the original heading node to the list of returned items so people have more control over how headings are displayed in the ToC
- update returns of builder function to match current structure
- update types

Addresses the following issues:
- Closes #476 
- Closes #450

### DIFF
--- a/src/lib/builders/table-of-contents/create.ts
+++ b/src/lib/builders/table-of-contents/create.ts
@@ -8,6 +8,7 @@ import type { Defaults } from '$lib/internal/types';
 
 import { onMount } from 'svelte';
 import { get, writable, type Writable } from 'svelte/store';
+import { dequal } from 'dequal';
 
 import type {
 	Heading,
@@ -59,13 +60,17 @@ export function createTableOfContents(args: CreateTableOfContentsArgs) {
 	let headingsList: HTMLHeadingElement[] = [];
 	let elementsList: Element[] = [];
 	/** Lookup to see which heading an element belongs to. */
-	const elementHeadingLU: ElementHeadingLU = {};
+	let elementHeadingLU: ElementHeadingLU = {};
 	/** Lookup to see which parent headings a heading has. */
-	const headingParentsLU: HeadingParentsLU = {};
+	let headingParentsLU: HeadingParentsLU = {};
 	/** List of the active parent indexes. */
 	const activeParentIdxs: Writable<number[]> = writable([]);
 	/** List of the indexes of the visible elements. */
 	const visibleElementIdxs: Writable<number[]> = writable([]);
+
+	let elementTarget: Element | null = null;
+
+	let mutationObserver: MutationObserver | null = null;
 
 	let observer: IntersectionObserver | null = null;
 	const observer_threshold = 0.25;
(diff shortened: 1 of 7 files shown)


src/lib/builders/table-of-contents/create.ts
```diff
@@ -109,10 +115,13 @@ export function createTableOfContents(args: CreateTableOfContentsArgs) {
...
 
-	function generateInitialLists(): void {
+	function generateInitialLists(elementTarget: Element) {
+		let headingsList: HTMLHeadingElement[] = [];
+		let elementsList: Element[] = [];
+
 		const includedHeadings = possibleHeadings.filter((h) => !exclude.includes(h));
 
-		const elementTarget = document.querySelector(selector);
+		// const elementTarget = document.querySelector(selector);
```
TGlide, line 124: We can remove this comment :)

TGlide (approved):
Awesome, LGTM! Just one small comment, and please make sure the lint and check pipelines are passing.

P.S. You can use `pnpm run lint:write` and `pnpm run check`

TGlide (approved):
Thank you!!

crisguitar:
thanks so much for this!

PR #486, opened 2023-09-03
fix: Select isSelected function for single option.

wildcant:
Hey guys, this is the same as https://github.com/melt-ui/melt-ui/pull/464 but for the Select builder.
It's not letting me use object as values.


TGlide (approved):
Thank you for the PR!

PR #487, opened 2023-09-04
fix: ensure type safety for `disabled` prop

kelvindecosta:
I'm wondering whether it makes more sense to do the following instead, since I've seen `data-disabled="undefined"` sometimes:

```diff
{
-    disabled: $disabled ?? undefined,
+    ...($disabled ? { disabled: true } : {}),
}
```

### DIFF
--- a/src/lib/builders/accordion/create.ts
+++ b/src/lib/builders/accordion/create.ts
@@ -100,7 +100,7 @@ export const createAccordion = <Multiple extends boolean = false>(
 				// generate the content ID here so that we can grab it in the content
 				// builder action to ensure the values match.
 				return {
-					disabled: $disabled || disabled,
+					disabled: ($disabled || disabled) ?? undefined,
 					'aria-expanded': isSelected(itemValue, $value) ? true : false,
 					'aria-disabled': disabled ? true : false,
 					'data-disabled': disabled ? true : undefined,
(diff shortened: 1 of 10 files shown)


src/lib/builders/accordion/create.ts
```diff
@@ -86,7 +87,7 @@ export const createAccordion = <Multiple extends boolean = false>(
 
 				return {
 					'data-state': isSelected(itemValue, $value) ? 'open' : 'closed',
-					'data-disabled': disabled ? true : undefined,
+					'data-disabled': disabledAttr(!!disabled),
```
TGlide, line 90: I don't think we should required the `!!`. If its `undefined` we treat it as false inside the function.

TGlide (commented):
Thank you for the PR!

src/lib/builders/accordion/create.ts
kelvindecosta: It is currently required for type-safety since `disabledAttr` takes in a `boolean`:

https://github.com/melt-ui/melt-ui/blob/2201f0adc75e20601b0748710110e4a81c56d441/src/lib/internal/helpers/attr.ts#L1-L3

Would it make more sense to take in `boolean | undefined`?

TGlide: > Would it make more sense to take in boolean | undefined?

Yes!

TGlide (approved):

Issue #488, opened 2023-09-04
🌟Feature request: Focus to first input on click when Pin Input is fully empty regardless of which individual input was clicked


yekta:
**Is your feature request related to a problem? Please describe.**
Kinda. When the `Pin Input` is fully empty and when I click on any of the individual inputs, I expect the first input field to be focused since there is not much point for a user to want to focus to say 4th input if the pin input field is fully empty.

**Describe the solution you'd like**
If `Pin Input` is fully empty, I would like the first input field to be focused when I click anywhere in the `Pin Input`. This makes hitting the `Pin Input` target easier. I don't see a use case for actually wanting to click say 4th input while the entire `Pin Input` field is empty. Usually you just want to click the field as fast as possible and enter the pin as fast as possible.

yekta:
> This is tricky. Here are my thoughts.
> 
> This would make it similar in behaviour to a normal input would, with just different looks. If that is desired, what you describe is ideal. But it goes beyond that.
> 
> Let's imagine we have a pin input with 4 characters.
> 
> * Fully empty and I click? Focus on the first.
> * If we have `1 2 ○ ○`, when I click on an empty field, it should probably always focus the first empty field.
> * What if I have `1 2 3 4`, and delete the first two numbers? Should it become `3 4 ○ ○`, like a normal input would, or `○ ○ 3 4`, how we visually would expect it to?
> * If I click on an empty field in `○ 2 ○ 4`, what should it do?
> * ...
> 
> I think this will lead us to a road where the behaviour of pin-input becomes unpredictable. There are some things we can do to improve experience though.
> 
> * Make it so only the first empty element has tabIndex="0". If all are filled, make the last filled item have tabIndex="0".
> * Add Home and End key handlers
> 
> And one thing that we already do: No matter which input is focused when you paste a code. it will do its best to fill the pin input accordingly. e.g. if the code is `1234`, and I'm focused on the 3rd input and I press paste, the result will be `1 2 3 4`, and not `○ ○ 1 2`.

I don’t think you need to answer all those questions to implement the rule I suggested 🤔

Rest of the rules you can evaluate completely separately. I’m suggesting a single rule, based on a single condition. I don’t think there is a single use-case for wanting to click to say the third field if all fields are empty. If anything, I think the rule I suggested makes the pin field more predictable not less. I have opinions about other rules as well but I didn’t write them in detail because majority of the gain in my opinion comes from the rule I suggested. Rest of the rules include more complicated conditions.

[...]

PR #492, opened 2023-09-06
Feat: Focus first empty input on pininput focus

huntabyte:
Closes: #488

When focusing on an individual input within a pin input, we'll check to see if the previous input's value is empty. If so, we focus on the previous input and do so until we find the first input in the pin input with a value, or focus all the way back to the first input.


TGlide:
Closed for reasons mentioned in #488

PR #514, opened 2023-09-11
Add Fathom analytics

huntabyte:
Just need to add those two env vars to Vercel and this will be good to go :)


TGlide (approved):
Awesomeee, thank you! ❤️

PR #515, opened 2023-09-11
tooltip builder `disableHoverableContent` option

aaronjbecker:
This PR adds `disableHoverableContent` as an option to the tooltip builder function. The behavior is consistent with same option in Radix and the implementation simply removes the content element from the polygon used for isMouseInTooltipArea. 

Default for the option is false, same as current behavior. I experimented with the docs example and the change works but I haven't implemented any rigorous testing. 

I also updated the documentation for the tooltip builder to reflect the new option.

### DIFF
--- a/.changeset/shaggy-fans-cheer.md
+++ b/.changeset/shaggy-fans-cheer.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": minor
+---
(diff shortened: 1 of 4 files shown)


TGlide (approved):
Amazing, thank you! 🎉

PR #522, opened 2023-09-12
Update docs for combobox and add debounce example

wesharper:
https://github.com/melt-ui/melt-ui/assets/13283469/d0aa962c-8d47-45fe-a3aa-40879741b070


buhodev:
Doesn't the combobox export a [`debounce`](https://www.melt-ui.com/docs/builders/combobox#:~:text=Change%20Functions-,debounce,-0) property already? :)

PR #524, opened 2023-09-12
Resolve linter issues preventing PR checks from passing

wesharper:
Runs prettier and fixes one eslint issue.


wesharper:
cc @huntabyte, this should unblock your changeset PR

PR #532, opened 2023-09-14
[New Builder] Autocomplete Tags

Bubbet:
Multi select with easy clearing/displaying of selected options.


TGlide:
Closing as I want to address this in a different manner. Will try and make combobox and tags-input composable somehow. Thank you for the hard work!

PR #539, opened 2023-09-18
fix: select controlled store type

MadeBaruna:
The default value of the store can be undefined.


huntabyte:
If `defaultSelected` is set to undefine, wouldn't `S` be typed as `undefined` thus making it accurately typed? Or maybe default the `S` type argument to undefined in the event `defaultSelected` isn't passed?

PR #545, opened 2023-09-19
feat: pin input mask option

### LINKED ISSUE #493: 🌟Feature request: Mask option in pin input

0xHati:
I'm only wondering if default values are set than the mask is immediately shown. Then again I don't think it's logical to use default values and mask together?

Fixes #493

### DIFF
--- a/src/lib/builders/pin-input/create.ts
+++ b/src/lib/builders/pin-input/create.ts
@@ -44,6 +44,7 @@ const defaults = {
 	type: 'text',
 	name: undefined,
 	defaultValue: [],
+	mask: false,
 } satisfies Defaults<CreatePinInputProps>;
 
 export function createPinInput(props?: CreatePinInputProps) {
@@ -95,7 +96,7 @@ export function createPinInput(props?: CreatePinInputProps) {
 					placeholder: $placeholder,
 					disabled: disabledAttr($disabled),
 					type: $type,
-					value: currValue,
+					value: options.mask && currValue ? '*' : currValue,
 				};
 			};
 		},
--- a/src/lib/builders/pin-input/create.ts
+++ b/src/lib/builders/pin-input/create.ts
@@ -51,7 +51,7 @@ export function createPinInput(props?: CreatePinInputProps) {
 	const withDefaults = { ...defaults, ...props } satisfies CreatePinInputProps;
 
 	const options = toWritableStores(omit(withDefaults, 'value'));
-	const { placeholder, disabled, type, name: nameStore } = options;
+	const { placeholder, disabled, type, name: nameStore, mask } = options;
 
 	const valueWritable = withDefaults.value ?? writable(withDefaults.defaultValue);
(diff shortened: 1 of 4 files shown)


src/lib/builders/pin-input/types.ts
```diff
@@ -55,6 +55,13 @@ export type CreatePinInputProps = {
 	 * @see https://melt-ui.com/docs/controlled#change-functions
 	 */
 	onValueChange?: ChangeFn<string[]>;
+
+	/**
+	 * Whether you want to hide the input with a mask.
+	 *
+	 * @default false
+	 */
+	mask?: boolean;
```
TGlide, line 64: Rather than a boolean, maybe it should allow a string? So they can choose what appears in the input.

src/lib/builders/pin-input/create.ts
```diff
@@ -95,7 +96,7 @@ export function createPinInput(props?: CreatePinInputProps) {
 					placeholder: $placeholder,
 					disabled: disabledAttr($disabled),
 					type: $type,
-					value: currValue,
+					value: options.mask && currValue ? '*' : currValue,
```
TGlide, line 99: Does the hiddenInput value still reflect the correct value with this?

TGlide (commented):
Thank you for the PR!

src/lib/builders/pin-input/types.ts
0xHati: Yes, that's definitely better than boolean, allows for more flexibility. I'll update it.

src/lib/builders/pin-input/create.ts
0xHati: Hmm I see what you mean, I was too quick on this one 😅. The hidden input receives its value from 'value' which will be the maskvalue and not the real value.

I'm on mobile so could only look briefly. The easiest way would be to set the input type to password if masked is selected, but not sure if that's compatible with letting the user change the character.

0xHati:
With regards to your two suggestions:
- using a string vs boolean
- the disconnect with the input field in the hidden input fields for form submission

What I did now what set the input to type password when `masked === true`, which will result in the browser default behavior for masking. I fixed the disconnect between the input and hiddenInput. I've been thinking on how to use a string instead for the mask like you suggested, but I don't know how that should be implemented, probably it will add some complexity and don't know if it's really worth it. What do you think? Overriding the password symbol used by the browser doesn't seem possible.

TGlide:
@0xHati Hmm. In that case, I'd say the `mask` prop is not necessary. We can narrow the `type` prop's type to be 'text' | 'password', instead of string, and document this password behaviour. Wdyt?

0xHati:
Updated to your suggestion, I think removing the prop is ok. It's a choice between being more explicit or not, but with the docs it should be clear how to achieve a masked pin input.

TGlide (approved):
Thank you! 🎉

PR #547, opened 2023-09-20
[New Builder] Color Picker

bennymi:
A color-picker builder. Still stuff to do:

TODOs:
- [x] create preview example of vertical sliders
- [x] bind color store to changes in canvas
- [x] create input builders
- [x] helper function for detecting the browser for eye dropper?
- [x] make preview mobile friendly
- [x] make dragging work on mobile
- [x] update docs
- [ ] Allow changing canvas axes (similar to [React Aria](https://react-spectrum.adobe.com/react-aria/useColorArea.html))


.changeset/popular-ducks-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@melt-ui/svelte': minor
+---
+
+add color picker builder
```
TGlide, line 5:
```suggestion
New Builder: Color Picker
```

src/docs/data/builders/color-picker.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const builder: APISchema = {
+	title: 'createColorPicker',
+	description: DESCRIPTIONS.BUILDER('color picker'),
+	props: [
+		{
+			name: 'defaultColor',
+			type: 'string',
+			description: 'The default selected color. Should be a hex value, eg: #424242.',
+		},
```
TGlide, line 13: It should allow passing in a color store, and a onColorChange fn. Maybe it should be changed to value?

Also, should we maybe allow hsl and rgb? At least we should have derived states, or helpers

src/lib/builders/color-picker/create.ts
```diff
@@ -0,0 +1,898 @@
...
+
+export function createColorPicker(args?: CreateColorPickerProps) {
+    const argsWithDefaults = { ...defaults, ...args };
+
+    let dragging = false;
+    let hueDragging = false;
+    let alphaDragging = false;
+    let insideUpdate = false;
+    let inputUpdate = false;
+    const speepUpStep = 5;
```
TGlide, line 27: What does this variable do?

src/lib/builders/color-picker/create.ts
```diff
@@ -0,0 +1,898 @@
...
+
+    const color: Writable<string> = writable(defaults.defaultColor);
+
+    const colorCanvasDims: Writable<NodeElement<HTMLCanvasElement>> = writable({ height: 1, width: 1 });
+    const colorPickerDims: Writable<NodeSize> = writable({ height: 1, width: 1 });
+    const colorPickerPos: Writable<{ x: number; y: number; }> = writable({ x: 0, y: 0 });
+
+    const hueSliderDims: Writable<NodeElement<HTMLCanvasElement>> = writable({ height: 1, width: 1 });
+    const huePickerDims: Writable<NodeSize> = writable({ height: 1, width: 1 });
+    // const huePickerPos: Writable<Position> = writable({ x: 0, y: 0 });
```
TGlide, line 44: Leftover

src/lib/builders/color-picker/create.ts
```diff
@@ -0,0 +1,898 @@
...
+    const hueAngle: Writable<number> = writable(0);
+
+    const alphaSliderDims: Writable<NodeElement<HTMLCanvasElement>> = writable({ height: 1, width: 1 });
+    const alphaPickerDims: Writable<NodeSize> = writable({ height: 1, width: 1 });
+    const alphaValue: Writable<number> = writable(100);
+
+    let eye: EyeDropperType | null = null;
+
+    const isValidHex = (hex: string) => /^#([0-9a-f]{3}){1,2}$/i.test(hex);
```
TGlide, line 53: This can be defined outside of the builder (in the same file)

src/lib/builders/color-picker/create.ts
```diff
@@ -0,0 +1,898 @@
...
+     * Update the canvases when the color changes.
+     */
+    // const updateCanvases = derived([color], ([$color]) => {
+    //     if (isValidHex($color)) updateOnColorInput($color);
+    // });
+
+    /**
+     * Returns the current color of the color picker.
+     */
+    const getCurrentColor: Readable<() => ReturnedColor> = derived([hueAngle, colorPickerPos, colorCanvasDims, alphaValue], ([$hueAngle, $colorPickerPos, $colorCanvasDims, $alphaValue]) => {
```
TGlide, line 67: Why is this needed when you already have `$color`? 

If it's to get access to RGB values, maybe it makes more sense to have derived states, one for RGB, one for HSL, on for HEX, and one for HSV (am I missing more?). Or even just a single derived state with all of them. And it doesn't need to be a function I believe

src/lib/builders/color-picker/create.ts
```diff
@@ -0,0 +1,898 @@
...
+
+        return () => ({ rgb: { ...rgb, a: $alphaValue }, hex });
+    });
+
+    const colorCanvas = builder(name('color-canvas'), {
+        stores: [hueAngle],
+        returned: ([$hueAngle]) => {
+            return {
+                'aria-label': 'Color canvas for showing saturation and brightness.',
+                style: `background-color: hsl(${$hueAngle}, 100%, 50%); background-image: linear-gradient(to top, rgba(0, 0, 0, 1), rgba(0, 0, 0, 0)), linear-gradient(to right, rgba(255, 255, 255, 1), rgba(255, 255, 255, 0));`
```
TGlide, line 82: Use the `styleToString` helper, it'll be cleaner :)

src/lib/builders/color-picker/create.ts
```diff
@@ -0,0 +1,898 @@
...
+
+    const colorPicker = builder(name('color-picker'), {
+        stores: [colorPickerPos, colorPickerDims, color],
+        returned: ([$colorPickerPos, $colorPickerDims, $color]) => {
+            const top = Math.round($colorPickerPos.y - $colorPickerDims.height / 2);
+            const left = Math.round($colorPickerPos.x - $colorPickerDims.width / 2);
+
+            return {
+                'aria-label': 'Button on color canvas, used to select the saturation and brightness.',
+                // style: `position: absolute; top: ${top}px; left: ${left}px; background-color: rgb(${r}, ${g}, ${b});`
```
TGlide, line 154: Leftover

src/lib/builders/color-picker/create.ts
```diff
@@ -0,0 +1,898 @@
...
+                addMeltEventListener(node, 'keyup', (e) => {
+                    const { key } = e;
+                    const keys = ['ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight'];
+
+                    if (!keys.includes(key)) return;
+
+                    keyDurations[<ArrowKeys>key] = null;
+                }),
+                addMeltEventListener(node, 'keydown', (e) => {
+                    // const { key } = e;
```
TGlide, line 182: Leftover

src/lib/builders/color-picker/create.ts
```diff
@@ -0,0 +1,898 @@
...
+    });
+
+    const huePicker = builder(name('hue-picker'), {
+        stores: [hueAngle, huePickerDims, hueSliderDims],
+        returned: ([$hueAngle, $huePickerDims, $hueSliderDims]) => {
+            const left = Math.round($hueAngle / 360 * $hueSliderDims.width - $huePickerDims.width / 2);
+
+            return {
+                'aria-label': 'The button to select the hue color.',
+                style: `position: absolute; background: hsl(${$hueAngle}, 100%, 50%); left: ${left}px; top: 50%; transform: translateY(-50%);`
```
TGlide, line 311: Use `styleToString`

src/lib/builders/color-picker/create.ts
```diff
@@ -0,0 +1,898 @@
...
+            return {
+                destroy() {
+                    unsubEvents();
+                }
+            }
+        }
+    });
+
+    // Helper functions
+    function RGBtoHex(rgb: ColorRGB) {
```
TGlide, line 542: The ones that are not store dependent can be outside of the builder fn (same file, or in `internal/helpers`) for better organization

TGlide (commented):
Thank you, this is massive!!

I have a few suggestions:
- When blurring the input, or pressing enter, if the value is not a valid color, it should revert to the last valid color input
- KBD support (Ark is a good example https://ark-ui.com/docs/components/color-picker)
- Don't forget to populate the `events.ts` file 🙏 

Again, thank you SO much. This is truly amazing 🎉

src/lib/builders/color-picker/create.ts
bennymi: @TGlide that's a typo actually, didn't notice that. It's supposed to be `speedUpStep`, basically when you use arrow keys to move the color picker around the main canvas, the picker moves pretty slowly (1 pixel at a time), so I added this larger step size that gets activated after an arrow key has been pressed for more than 1 second.

src/lib/builders/color-picker/create.ts
bennymi: @TGlide It's actually a leftover from a previous implementation where I was using a different approach. I think having one `$value` store with a hex value, and a derived store with different color representations makes sense. I'll add it.

src/docs/data/builders/color-picker.ts
bennymi: @TGlide I was thinking about that as well, but having multiple ways of changing the color would quickly lead to circular dependencies I believe, so I think we should stick with one `$value` store where the hex value can be changed and some derived stores for other color values. We can also export functions so users can easily change between different color formats and mention those in the docs.

TGlide: Hmm, why would it lead to circular deps? I think you'd just need a derived that converts it into a single format. So you'd have the value store, that can be anything, and then a derived that tries and converts it. If it's an invalid format, it returns null, or the previously valid input. Do you see that working out?

bennymi: @TGlide I meant that if the user can change the color through multiple stores (one for hex, one for rgb, etc) then it could lead to circular dependencies I believe, because if the hex value changes you would have to update the rgb values and if the rgb values change you would have to update the hex value. But if we just have one `$value` store to control the color and some readable derived stores than it's not a problem.

src/lib/builders/color-picker/create.ts
```diff
@@ -0,0 +1,960 @@
...
+
+            return {
+                destroy() {
+                    unsubEvents();
+                },
+            }
+        }
+    });
+
+    const hueSlider = builder(name('hue-slider'), {
```
TGlide, line 264: Would be cool to have keyboard events on the sliders as well 😄

src/lib/builders/color-picker/create.ts
```diff
@@ -0,0 +1,960 @@
...
+        }
+    }
+
+    /**
+     * Returns the current color of the color picker.
+     * This function is used in some builder return parameters,
+     * as well as the subscribe methods, since we sometimes need to
+     * update the $colors value when the hue angle is changed.
+     */
+    const getCurrentColor: Readable<() => { rgb: ColorRGB, hex: string }> = derived([hueAngle, colorPickerPos, colorCanvasDims, alphaValue], ([$hueAngle, $colorPickerPos, $colorCanvasDims, $alphaValue]) => {
```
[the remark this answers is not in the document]
bennymi: @TGlide yeah we still need this function. `derivedColors` is based on the `value` store, but sometimes we also need to update the `value` store based on changes in the hue. Maybe there's a different way to achieve this, but I wasn't able to think of it.

src/lib/builders/color-picker/create.ts
bennymi: @TGlide ah, you mean that if the slider is focused, rather than just the slider button?

TGlide:
Closing, will be added in Runes version

PR #550, opened 2023-09-21
Up key should focus last menu item if there is no selection

acurrieclark:
When a menu such as `DropDownMenu` is opened there is initially no focussed item, which is expected.

Current behaviour is that when either the down or up arrow key are pressed, the first item is always highlighted. To my mind, down should focus the first item, while up should focus the last. This matches the current behaviour of the ComboBox component.

This PR implements what I think should be the expected behaviour.

### DIFF
--- a/.changeset/tricky-plums-speak.md
+++ b/.changeset/tricky-plums-speak.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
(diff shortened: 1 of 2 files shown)


TGlide (approved):
Awesome! TBH I also realized: pressing up while the trigger is focused should also open it. And it should loop around. I'll create issues for these though.

PR #552, opened 2023-09-22
[New Builder] Spinbutton

NDOY3M4N:
I tried to implement a basic spinbutton builder by following this [example](https://www.w3.org/WAI/ARIA/apg/patterns/spinbutton/examples/datepicker-spinbuttons/). This is still a work in progress, but I'll continue to work on it.


TGlide:
Awesome, thank you so much! I'll mark this as a draft while it's in progress ;)

TGlide:
@huntabyte maybe you want to adapt your spinbutton builder instead?

TGlide:
Will adapt from Date Picker's spin buttons later

PR #575, opened 2023-10-02
Add tooltip `group` param

abdel-17:
This new parameter allows tooltip to be grouped together, coupling their `open` states so that no more than 1 tooltip can be open in a single group.

This pr includes changes to the `createTooltip` function and its types, and some unit tests to validate the functionality.

### DIFF
--- a/src/lib/builders/tooltip/create.ts
+++ b/src/lib/builders/tooltip/create.ts
@@ -246,18 +246,18 @@ export function createTooltip(props?: CreateTooltipProps) {
 	let isMouseInTooltipArea = false;
 
 	effect(open, ($open) => {
-		if ($open) {
-			const groupValue = get(group);
-			if (groupValue === undefined || groupValue === false) {
-				return;
-			}
-
-			// Close the currently open tooltip in the same group
-			// and set this tooltip as the open one.
-			const currentOpen = _groupMap.get(groupValue);
-			currentOpen?.set(false);
-			_groupMap.set(groupValue, open);
+		if (!$open) return
+		const groupValue = get(group);
+		if (groupValue === undefined || groupValue === false) {
+			return;
 		}
+
+		// Close the currently open tooltip in the same group
+		// and set this tooltip as the open one.
+		const currentOpen = _groupMap.get(groupValue);
+		currentOpen?.set(false);
+		_groupMap.set(groupValue, open);
+	
 	});
 
 	effect([isVisible, activeTrigger], ([$isVisible, $activeTrigger]) => {
--- a/src/lib/builders/tooltip/create.ts
+++ b/src/lib/builders/tooltip/create.ts
@@ -23,7 +23,7 @@ import {
 import { useFloating, usePortal } from '$lib/internal/actions/index.js';
 import type { MeltActionReturn } from '$lib/internal/types.js';
 import { onMount, tick } from 'svelte';
-import { get, writable } from 'svelte/store';
+import { get, writable, type Writable } from 'svelte/store';
 import type { TooltipEvents } from './events.js';
 import type { CreateTooltipProps } from './types.js';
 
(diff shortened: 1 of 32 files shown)


abdel-17:
Okay, so somehow the previous slider docs pr got merged into this one and I don't know how to fix it. Sorry for that.

src/lib/builders/tooltip/create.ts
```diff
@@ -40,11 +40,15 @@ const defaults = {
...
 	closeOnEscape: true,
 	disableHoverableContent: false,
+	group: undefined,
 } satisfies CreateTooltipProps;
 
 type TooltipParts = 'trigger' | 'content' | 'arrow';
 const { name } = createElHelpers<TooltipParts>('tooltip');
 
+// Store a global map to get the currently open tooltip.
+const _groupMap = new Map<string | boolean, Writable<boolean>>();
```
TGlide, line 50: Probably should be `string | true`, as we don't want `false` to be a valid group option

src/lib/builders/tooltip/create.ts
```diff
@@ -240,6 +245,21 @@ export function createTooltip(props?: CreateTooltipProps) {
...
+			if (groupValue === undefined || groupValue === false) {
+				return;
+			}
+
+			// Close the currently open tooltip in the same group
+			// and set this tooltip as the open one.
+			const currentOpen = _groupMap.get(groupValue);
+			currentOpen?.set(false);
+			_groupMap.set(groupValue, open);
+		}
```
TGlide, line 260:
```suggestion
		if (!$open) return
		const groupValue = get(group);
		if (groupValue === undefined || groupValue === false) {
			return;
		}

		// Close the currently open tooltip in the same group
		// and set this tooltip as the open one.
		const currentOpen = _groupMap.get(groupValue);
		currentOpen?.set(false);
		_groupMap.set(groupValue, open);
	
```

src/lib/builders/tooltip/create.ts
TGlide: Also, why the underline?

TGlide (commented):
Thank you for the PR! Please don't forrget to add documentation for the group feature :)

src/lib/builders/tooltip/create.ts
abdel-17: I did handle the `false` case below, where it acts similar to `undefined`. I think it's a valid choice as both as "falsy", but I can remove it if you like.

The underline, eh, idk? It's internal global state, I just did it out of habit. I'll remove it.

abdel-17:
> Thank you for the PR! Please don't forrget to add documentation for the group feature :)

Hey! I wanted to add documentation, but couldn't figure out a way to clearly explain what `group` does. Can you suggest an explanation to add to the JSDoc? I'll add it to the documentation site.

TGlide:
> > Thank you for the PR! Please don't forrget to add documentation for the group feature :)
> 
> Hey! I wanted to add documentation, but couldn't figure out a way to clearly explain what `group` does. Can you suggest an explanation to add to the JSDoc? I'll add it to the documentation site.

Maybe:

> When `true`, whenever you open a tooltip, all other tooltips with `group` also set to `true` will close. If you pass in a `string` instead, only tooltips with the same `group` value will be closed.

abdel-17:
> > > Thank you for the PR! Please don't forrget to add documentation for the group feature :)
> > 
> > 
> > Hey! I wanted to add documentation, but couldn't figure out a way to clearly explain what `group` does. Can you suggest an explanation to add to the JSDoc? I'll add it to the documentation site.
> 
> Maybe:
> 
> > When `true`, whenever you open a tooltip, all other tooltips with `group` also set to `true` will close. If you pass in a `string` instead, only tooltips with the same `group` value will be closed.

Hey! I add `group` to the documentation. I wanted to add an example, but the Tooltip didn't have any and I'm not familiar with how to add one, so I'll leave this one up to you if you want to add one.

Issue #574, opened 2023-10-01
📁Docs: Convert all previews to CSS-only


TGlide:
Currently, previews for builders can be found in `src/docs/previews/BUILDER_NAME/EXAMPLE_NAME/STYLING/*`, where `STYLING` is either `tailwind` or `css`. All builders have tailwind examples, but some are missing CSS examples.

I want to do two things. First, add CSS examples to all previews. Then, I want to remove the Tailwind examples to make it easier to maintain. I need help with the first task.

In the branch `css-examples`, I've already converted `previews/accordion/main` to CSS. You can see an example there, and you can preview the component by heading to the `Accordion` preview in the docs, when running it locally. Anytime you add a CSS example, that's what's going to show in the preview docs of the `css-examples` branch instead of the `tailwind` example. As such, all PRs related to this matter should be based on `css-examples`, and point to it.

A single PR should not convert all of the previews! Instead, 1-5 previews would be best 😄 

## Please mention which examples you'll be working on!

Dhoni77:
Hi @TGlide  Can you provide some information on what is expected in this issue

TGlide:
Hi @Dhoni77! I updated the description :)

Sanchita1304:
HI, I would like to do this. Can you assign me.
And In this I just need to make css file for all previews and paste the tailwind to css and delete the tailwind file right?

TGlide:
> HI, I would like to do this. Can you assign me. And In this I just need to make css file for all previews and paste the tailwind to css and delete the tailwind file right?

Hi! This issue doesn't need assigning as multiple people can tackle it, working with different examples. 

Please refer to the description! Copy-pasting the Tailwind to CSS won't be enough. You need to actually convert the code to not use tailwind anymore. And no need to delete the Tailwind file, I'll do that after all examples have CSS code.

Sanchita1304:
Ok I will do this

Amanjot-001:
@TGlide Hi, i am working on it but need a little help in how to set the preview in docs to be loaded from css file so that i can check  the changes. and also how can i create a test.

henrikvilhelmberglund:
I'm a bit confused, wouldn't having examples using Tailwind be more maintainable? I guess it doesn't matter in the end as they are unstyled but having some nice examples in Tailwind format too would be pretty useful, even if it's an unofficial unmaintained thing.

[...]

PR #578, opened 2023-10-02
fix: created css for combobox #574

Amanjot-001:
# created a  css file for the combobox main preview.
issue: #574


TGlide:
Hi @Amanjot-001, thank you for the PR!

The PR should be based and pointing to the `css-examples` branch. The CSS structure should also more or less follow the example I've created there for the main Accordion preview

PR #581, opened 2023-10-03
list items gets pushed down

Ajitesh72:
the ChevronsUpDown to get replaced by X and the list items gets pushed further down the initial items in the docs
 
 previously:

https://github.com/melt-ui/melt-ui/assets/95878363/70a1c955-4c6b-4460-ac08-d34b32156e74

now:

https://github.com/melt-ui/melt-ui/assets/95878363/06bafbe6-24d1-4db7-848b-9ba0296bfa65


huntabyte (changes requested):
The updates look great! It looks like you accidentally committed a modified `package.json` and `pnpm-lock.yaml` which has `@melt-ui/svelte` as a linked dependency, so just have to discard those changes!

Ajitesh72:
reverted back to package.json and pnpm-lock.yaml file present before my PR

TGlide:
> I would also say, I think the opened version should be centered, while the closed one doesn't need to be

@Ajitesh72 Don't forget to address this comment, if you can 🙂

Ajitesh72:
> > I would also say, I think the opened version should be centered, while the closed one doesn't need to be
> 
> @Ajitesh72 Don't forget to address this comment, if you can 🙂

@TGlide  I want to make sure I understand your suggestion correctly. Could you please clarify what you mean by "the opened version should be centered, while the closed one doesn't need to be"?
won't that make the user experience a little complicated since he would have to hover to x then back to ChevronsUpDown,and won't it be the same as before?

TGlide:
> @TGlide I want to make sure I understand your suggestion correctly. Could you please clarify what you mean by "the opened version should be centered, while the closed one doesn't need to be"? won't that make the user experience a little complicated since he would have to hover to x then back to ChevronsUpDown,and won't it be the same as before?

What I mean is: Currently in your PR, when the Collapsible is closed, it's centered. When it's open, it's off-center.
image
image

My suggestion is that the opposite should happen, the open state should be centered.

Ajitesh72:
@TGlide let me know if there is any better approach to tackle this,i choose not to make the initial state as open because all the other components in the docs had initial state as closed

TGlide:
> @TGlide let me know if there is any better approach to tackle this,i choose not to make the initial state as open because all the other components in the docs had initial state as closed

I think this is good, we can change it in the future if need be :)

PR #584, opened 2023-10-03
Add `aria-disabled` to slider root

abdel-17:
This PR adds an aria-disabled attributes to the Slider's root element.

PS: sorry for doing 2 PRs, the old branch got messed up, so I deleted it.

### DIFF
--- a/src/lib/builders/slider/create.ts
+++ b/src/lib/builders/slider/create.ts
@@ -55,6 +55,7 @@ export const createSlider = (props?: CreateSliderProps) => {
 		returned: ([$disabled, $orientation]) => {
 			return {
 				disabled: disabledAttr($disabled),
+                'aria-disabled': $disabled ? 'true' : 'false',
 				'data-orientation': $orientation,
 				style: $disabled ? undefined : 'touch-action: none;',
 				'data-melt-id': ids.root,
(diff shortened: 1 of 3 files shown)


src/lib/builders/slider/create.ts
```diff
@@ -55,6 +55,7 @@ export const createSlider = (props?: CreateSliderProps) => {
 		returned: ([$disabled, $orientation]) => {
 			return {
 				disabled: disabledAttr($disabled),
+				'aria-disabled': $disabled ? 'true' : 'false',
```
TGlide, line 58: I think we can use the `disabledAttr` helper. It's easier for styling, and according to MDN it's valid behaviour to remove it when not disabled. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled#associated_roles

abdel-17: The problem with the helper is that it sets it to the boolean true, not the string, so Tailwind’s `aria-disabled:` doesn’t work. So, you’d need to change the helper or just make another helper for aria disabled.

TGlide: I see what you mean. In that case, `'aria-disabled': $disabled ? 'true' : undefined` should suffice I believe, right?

abdel-17: Yeah sure. That should support both use cases `[aria-disabled]` and `[aria-disabled="true"]`.

TGlide: Perfect then :)

TGlide: @abdel-17 it seems Svelte resolves `true` to `"true"` in this case

abdel-17: Strange. Looks like `aria-disabled` and `aria-disabled={true}` behave differently when it comes to DOM elements.

PR #590, opened 2023-10-04
Fix bug with Tooltip `group` parameter

abdel-17:
Tooltips with a `group` would prevent themselves from reopening again unless another tooltip in the same group was opened. This is due to the effect not cleaning up after the tooltip closes, so it stays in the "open tooltips map" forever.

This PR fixes the bug and adds a test to validate the fix.


TGlide (approved):
Thank you! 🎉

PR #591, opened 2023-10-04
chore: return ids from builders

### LINKED ISSUE #585: ⚙Tech Debt: Return the `id` object from builders
We're currently generating unique IDs that are applied to various elements within the builders.

While this is great, it makes it hard to bring in custom elements to use alongside builders. For example, if you wanted to use a label that wasn't provided by a builder, you'd need to know the ID of whatever element the label is `for`.

We already have the `ids` that we use internally inside of nice objects within the builders, so we should also return them for users to access/use outside as well.

It would be another property of the object returned from the builders, so to access them, you'd do something like this:

```ts
const { 
	ids: { ... },
	elements: { ... },
	// etc...
} = createWhatever()
``` 

With access, you can now select specific elements via custom actions, use custom labels, etc. The possibilities are truly endless for this revolutionary change!

0xHati:
closes #585

### DIFF
--- a/.changeset/hungry-queens-hide.md
+++ b/.changeset/hungry-queens-hide.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": minor
+---
+
+- return an `ids` object from builders which contains the `id`s Melt applies to certain elements internally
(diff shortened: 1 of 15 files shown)


huntabyte (approved):
Looks good!

TGlide (approved):
Awesome, thank you tons! 🎉

Issue #554, opened 2023-09-22
🌟Feature request: individual toast close delay 


geerew:
**Is your feature request related to a problem? Please describe.**
At the minute the close delay is set on the toaster and defaults to 5 seconds. It would be nice if I could override this when adding individual toasts.

for example, maybe I want ‘error’ toast to show for a bit longer and I want ’success’ toasts to show for less time

**Describe the solution you'd like**
An option to be passed on to the addToast helper

**Describe alternatives you've considered**
Currently the only way to do it is to have multiple createToast builders will different timeouts. It’s not optimal.

**Additional context**
N/a

lolcabanon:
I would also like to see this implemented! :)

lolcabanon:
Ok I just looked at the code and it is already possible!

It is not really documented, tho, but here you go :

```js

const {
  elements,
  helpers,
  states,
  actions
} = createToaster({
  closeDelay: 5000 // this is the default (in ms)
});

helpers.addToast({
	data: {
		title,
		description,
		type
	},
	closeDelay: 10000 // this overrides the default (10 seconds instead of 5)
});
```

geerew:
Good work. In that case this is a docs issue..

PR #597, opened 2023-10-06
Added code example to override closeDelay for individual toast (see #554)

lolcabanon:
Very simple. Added code example in the doc to show how to override default values of the toaster for individual toasts.

Feel free to point me if there would be some other place that would benefit this info!

### DIFF
--- a/src/docs/content/builders/toast.md
+++ b/src/docs/content/builders/toast.md
@@ -111,9 +111,7 @@ the application.
 				title: 'Success',
 				description: 'The resource was created!',
 				color: 'bg-green-500'
-			},
-			closeDelay: 10000, // overrides the default delay (5000)
-			type: 'background' // overrides the default type ('foreground')
+			}
 		})
 	}
 </script>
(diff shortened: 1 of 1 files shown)


geerew:
Is there a better way to show the parameters of the helper function? 

Examples are great but it needs drawn out in the schema too..

TGlide (commented):
Thank you for the PR! I think it'd be best to add a sub-section under usage detailing this behaviour instead.

lolcabanon:
Hey!

@TGlide ! I just updated the doc with a new section. Tell me what you think of it!

@geerew I agree it should be in the schema... But I am not sure how to do it! If you can give me an example or template, that would be great. Would you just add it to the description?

I guess you are referring to this part in the doc?

```md
<!-- melt-ui/src/docs/content/builders/toast.md -->

## API Reference

<APIReference {schemas} />
```
```ts
// melt-ui/src/docs/data/builders/toast.ts

const builder = builderSchema(BUILDER_NAME, {
        // ...
	helpers: [
		{
			name: 'addToast',
			type: '(props: AddToastProps<T>) => void',
			description: 'A helper function to add a toast to the toasts store.',
		},
                // ...
	],
        // ...
});
```

Thank you both for the follow-up! :sun_with_face:

TGlide:
> I agree it should be in the schema... But I am not sure how to do it! If you can give me an example or template, that would be great. Would you just add it to the description?

This is a docs issue currently. We can include longer schemas, but it's quite cumbersome currently. In a future iteration the API reference should be auto-generated, where it won't be a problem, but for now this is sufficient.

TGlide (approved):
Thank you! 🎉

PR #601, opened 2023-10-08
docs: add CSS example for radio-group

suliskh:
Add CSS example for Radio Group. Part of https://github.com/melt-ui/melt-ui/issues/574


TGlide (approved):
Thank you! 😄

PR #616, opened 2023-10-12
fix: `melt` action now throws an error if it's used without the PP

### LINKED ISSUE #606: 📁Docs: make importance of preprocessor setting more obvious in examples
First, thank you for what is already a great library.

But in writing my first component with it, I discovered a potential onboarding issue. After installing, I skipped straight to the example for the component I wanted to use and thought "Wow, that looks easy." And then proceeded to spin my wheels for quite some time trying to figure out why my radio buttons did absolutely nothing when I clicked them. 

_Eventually_ I figured out that I had to turn off the preprocessor switch in the lower left to see the code sample for my use case. To spare the next users who jump straight to the examples a similar frustration, perhaps something could be done to make the significance of that setting more prominent??

AdrianGonz97:
Partially fixes #606

Here's how the error will appear in the dev console:

console error

We could keep it brief like this by only providing a link to the PP docs, or we can further add install instructions such as: 

`Install the preprocessor with: "npm install -D @melt-ui/pp"`

or

`Install the preprocessor with: "npx @melt-ui/cli init"`

### DIFF
--- a/.changeset/eighty-items-march.md
+++ b/.changeset/eighty-items-march.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+The `melt` action now throws a helpful error if it's used without our [preprocessor](https://github.com/melt-ui/preprocessor)
--- a/src/lib/internal/actions/melt/index.ts
+++ b/src/lib/internal/actions/melt/index.ts
@@ -33,6 +33,7 @@ export function melt<
 	A extends Record<string, any>,
 	Param = never
 >(node: Element, params: Builder): ActionReturn<Builder, Attributes> {
-	// @ts-expect-error calls the action for debugging purposes
-	return params.action(node);
+	throw new Error(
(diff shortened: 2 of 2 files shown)


TGlide:
This is perfect, IMO. Thank you!

EquinoxFlowerJG:
I followed the installation section of the official documentation step by step, but I still encountered this error. What should I do about it?
My @melt-ui/svelte version is 0.70.0, and @melt-ui/pp version is 0.3.0. 
This is the code in my svelte.config.js file:
```js
import { vitePreprocess } from "@sveltejs/vite-plugin-svelte";
import { preprocessMeltUI, sequence } from "@melt-ui/pp";

const config = {
    // ... other svelte config options
    preprocess: sequence([
        vitePreprocess(),
        // ... other preprocessors
        preprocessMeltUI(), // add to the end!
    ]),
};

export default config;

```

PR #617, opened 2023-10-12
Docs:Document differences between tooltip and link-preview

### LINKED ISSUE #432: 📁Docs: Document differences between tooltip and link-preview

harshita1611:
Fixes #432


TGlide (changes requested):
Thanks for the PR!

Where is this page? I can't find it in the preview.

Nonetheless, this does not fit the issue. It's not accurate as to what the specifics of each builder and its differences are, and a dedicated page for it is a bit overkill. I'll be closing the PR :)

PR #623, opened 2023-10-13
Refresh component preview file tabs

huntabyte:
Makes the file tabs that come along with the component previews feel more connected and provides a familiar IDE tab experience.


TGlide (approved):
Awesome!

PR #627, opened 2023-10-14
Docs: add CSS example for switch component

Vikram-Hegde:
Added CSS example for switch component


TGlide (approved):
Awesome, thank you! 🎉

Issue #468, opened 2023-08-29
🐛Bug: Dialog overlay doesn't stop touch events from firing elements behind it


GingerAdonis:
**Describe the bug**
Dialog overlay doesn't stop touch events from firing other elements like it does with click events.

**To Reproduce**
Steps to reproduce the behaviour on desktop:
1. Go to: https://www.melt-ui.com/docs/builders/dialog
2. Scroll to 'Drawer' example component.
3. Enable responsive UI via dev tools. When clicking/touching any clickable element which is behind the overlay the event goes through.

**Expected behaviour**
Overlay stops touch events just like click events.

**Video**
https://github.com/melt-ui/melt-ui/assets/2751672/358bceea-ea04-4ed7-b7ba-18b1c81fa16b

**Desktop**
 - OS: Win11 x64
 - Browser: Chrome 116

**Smartphone**
 - OS: Android 13
 - Browser: Chrome 116

GingerAdonis:
This problem also occurs for me on a Pixel 5 with Android 13 (Chrome 116).

Video:
https://github.com/melt-ui/melt-ui/assets/2751672/9f728566-46fc-4e61-8b49-378748246e3f

alireza-ahmadi:
I experience the same issue on Safari (iOS 16.6). After digging a bit deeper, I found out using `transition:fade` causes this issue (if you remove the transition, everything works as expected).

As a temporarily workaround, you can break your overlay to two separate element and apply the background color and transition to the child element. In my case, it was something like this:

```svelte
<div
  {...$overlay}
  use:overlay
  class="fixed inset-0 z-50 bg-primary-950/50"
  transition:fade={{ duration: 250, easing: sineInOut }}
/>
```

turned into this:

```svelte
<div {...$overlay} use:overlay class="fixed inset-0 z-50">
  <div
    class="w-full h-full bg-primary-950/50"
    transition:fade={{ duration: 250, easing: sineInOut }}
  />
</div>
```

GingerAdonis:
Can confirm the work-around 'fixes' the issue. Thank you @alireza-ahmadi

evelinabe:
So we know the solution, it just needs a PR? I can do that :)

evelinabe:
This workaround bugs me, because that extra div shows as an example in the code in the docs and does not seem to make sense. And since this is more of a workaround to fix a bug in Svelte my vote would be to not fix it, perhaps mentioning it as a suggestion somehow.
[...]

PR #630, opened 2023-10-15
Increase transition duration when transition is used with overlay

evelinabe:
Dialog overlay doesn't stop touch events from firing other elements like it does with click events. It seems like increasing the duration for the transitions can prevent the touch events to fire events behind the overlay. This PR increases the duration from `150`to `350` in all instances where an overlay has a transition. This fixes the issue on iPad Air and iPhone 15 in xCode simulator. It's only tested in simulator, though, if a real touch event lasts longer it might need to be increased. Not tested on Android devices.

LInked to issue: https://github.com/melt-ui/melt-ui/issues/468

### DIFF
--- a/src/docs/components/custom-event-dialog.svelte
+++ b/src/docs/components/custom-event-dialog.svelte
@@ -26,7 +26,7 @@
     translate-x-[-50%] translate-y-[-50%] rounded-md
     bg-neutral-800 p-4 shadow-lg md:p-8"
 		transition:flyAndScale={{
-			duration: 150,
+			duration: 350,
 			y: 8,
 			start: 0.96,
 		}}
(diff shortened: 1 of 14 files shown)


GingerAdonis:
I believe this doesn't change anything for Android devices. Can test right now, but from what I recall the duration doesn't really matter. :(

evelinabe:
> I believe this doesn't change anything for Android devices. Can test right now, but from what I recall the duration doesn't really matter. :(

- You're right. Got it to work in Android Studio Emulator but duration didn't have any effect. Haven't found any solutions besides inserting an extra element :/

TGlide (commented):
Hi! Thank you for the PR, however I don't believe we should go with this approach. We should fix the underlying issue, rather than use a workaround in the docs. 

I'll be closing the PR, but I truly appreciate your effort 🙂

PR #639, opened 2023-10-18
docs: make toasts close automatically

diaasaur:
this pr removes `closeDelay: 0` so that toasts in docs can close automatically using the default `closeDelay` (5000ms)


TGlide (approved):
Thank you!

PR #640, opened 2023-10-18
Escape additional conflicting characters in markdown

huntabyte:
We're currently only escaping one of the 4 characters which causes issues when trying to render them within inline code blocks. This PR escapes the other entities as well.


huntabyte:
I'm an idiot

PR #645, opened 2023-10-19
Add toggle group tests

huntabyte:
- Add toggle group tests
- Fix aria attributes applied when in single/multiple mode which were causing violations due to the role being changed to radio.


src/lib/builders/toggle-group/create.ts
```diff
@@ -72,7 +72,8 @@ export const createToggleGroup = <T extends ToggleGroupType = 'single'>(
 					'data-disabled': disabledAttr(disabled),
 					'data-state': pressed ? 'on' : 'off',
 					'data-value': itemValue,
-					'aria-pressed': pressed,
+					'aria-pressed': $type === 'single' ? undefined : pressed,
+					'aria-checked': $type === 'multiple' || $type === undefined ? undefined : pressed,
```
TGlide, line 76: Inverted logic can be confusing

huntabyte:
g2g @TGlide

TGlide (approved):

PR #647, opened 2023-10-19
Implement disableTriggerRefocus

### LINKED ISSUE #646: 🐛Bug: Deep links (hash) inside dropdown menu does not work
**Describe the bug**
When using the dropdown menu with deep links `href="#some-element"` the scroll position is briefly moved to the element but then back to the menu

**To Reproduce**
1. open https://stackblitz.com/edit/melt-ui-playground-dghk9t?file=src%2Froutes%2F%2Bpage.svelte
2. Click on "This works" and see it works
3. Click on "Click me" to open drop down
4. Click on "This does not work" to view it does not work

**Expected behavior**
Expected page to be scrolled to the element

AndreasHald:
closes #646

Expose the existing `disableTriggerRefocus` from the `menu` builder to the `dropdown menu` builder, and exit early on the 2 calls that refocuses the trigger on close

This causes deep links (hash links) to work properly as when the property is true, the focus will not be moved back to the trigger

### DIFF
--- a/.changeset/empty-timers-enjoy.md
+++ b/.changeset/empty-timers-enjoy.md
@@ -1,0 +1,6 @@
+---
+"@melt-ui/svelte": patch
+---
+
+DropdownMenu: Implement `disableTriggerRefocus` prop
+
--- a/src/docs/constants.ts
+++ b/src/docs/constants.ts
@@ -46,6 +46,12 @@ export const PROPS = {
 		description: DESCRIPTIONS.FORCE_VISIBLE(args.name ?? 'element'),
 		default: args.default ?? 'false',
 	}),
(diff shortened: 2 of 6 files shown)


TGlide (commented):
Awesome! We also need to update the dropdownMenu docs to include this prop.

AndreasHald:
Screenshot 2023-10-22 at 18 37 00

Sweet docs setup :)! was easy to add or am I missing something?

TGlide:
> Screenshot 2023-10-22 at 18 37 00
> 
> Sweet docs setup :)! was easy to add or am I missing something?

No, that's it! Thank you 😄

TGlide (approved):

PR #657, opened 2023-10-21
Docs - difference between link preview  and tooltip

adeleke5140:
This PR will add a one-liner difference between using a `link-preview` and a `tooltip`. 

The question I tried to answer was: "Is there an alternative that solves keyboard accessibility?"

Fixes: #432

### DIFF
--- a/src/docs/components/index.ts
+++ b/src/docs/components/index.ts
@@ -25,4 +25,4 @@ export { default as TypeDialog } from './type-dialog.svelte';
 export { default as Logo } from './logo.svelte';
 export { default as CustomEventDialog } from './custom-event-dialog.svelte';
 export { default as Contributors } from './contributors.svelte';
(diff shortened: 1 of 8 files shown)


TGlide:
Hi! Thanks for the PR. 

This does not seem to document the difference between the two, especially since Tooltip isn't more accessible, it's just a different use case. I'll be closing this, but thank you for the effort nonetheless 🙂

adeleke5140:
> Hi! Thanks for the PR. 
> 
> 
> 
> This does not seem to document the difference between the two, especially since Tooltip isn't more accessible, it's just a different use case. I'll be closing this, but thank you for the effort nonetheless 🙂 

Got it. What differences would you like to see written about?

To me, it felt like they did the same function of showing some extra information in a given context except the tooltip can be accessed through the keyboard. 

Since I missed the mark, an outline of what this looks like would help me contribute better.

PR #659, opened 2023-10-22
fix(slider): avoid max overstep (#470)

### LINKED ISSUE #470: 🐛Bug: Slider step bug
**Describe the bug**
When the slider oversteps, it goes to the max value rather than stay in place.

As an example, this is what happens when `min = 0, max = 7, step = 6`.

https://github.com/melt-ui/melt-ui/assets/88583085/658eb1a7-97d9-4e72-bb14-36a22e0acead

**To Reproduce**
1. Go to https://stackblitz.com/edit/github-wgnbi5-ryrhwl?file=src%2Froutes%2F%2Bpage.svelte
2. Drag the slider all the way to the right (or use keyboard navigation)
3. See error

**Expected behavior**
The slider should stay in place at the value 6. This is the default HTML behavior and what `react-slider` does.
 
https://github.com/melt-ui/melt-ui/assets/88583085/51909cc2-e6cb-41d6-babe-84f7d78c93fe

darkresq14:
Added actual_max calculation and set to actual_max instead of overstepping on meta + arrow right, meta + arrow up, end and drag. Added check on arrow right and arrow up to not update in case new value would be bigger than max

Fixes #470


src/lib/builders/slider/create.ts
```diff
@@ -194,13 +194,15 @@ export const createSlider = (props?: CreateSliderProps) => {
 				const $value = get(value);
 				const $orientation = get(orientation);
 
+				const $actual_max = add($min, (mul(Math.floor(div(sub($max, $min), $step)), $step)))
```
TGlide, line 197: Prefer `camelCase` syntax, otherwise we'll have conflicting cases.

Also, `add($min, (mul(Math.floor(div(sub($max, $min), $step)), $step)))` is really hard to parse. I'd suggest offloading the calculation into other constants/parts, and then joining them all together in a human-readable way.

Also, I assume we also would need an `actualMin` constant?

TGlide (changes requested):
Thank you for the PR! I'm requesting a few small changes.

src/lib/builders/slider/create.ts
darkresq14: I gave it a lot of thought if `actualMin` is relevant or not. At the moment the native range input doesn't do defaultValue like this. For example if you have min 0, step 3 and set value to 2, the actual value will be 3 from the start. For our case if we do that, the initialValue shows 2 but when we move it the ticks are always starting from min so it will be either 0 or 3. This is why `actualMin` is not relevant imho.

What is your suggestion for this case? 
Btw I am very new to os contributing so I'll take all the tips I can get :)

src/lib/builders/slider/create.ts
```diff
@@ -338,12 +350,23 @@ export const createSlider = (props?: CreateSliderProps) => {
...
-					const newValue = mul(Math.round(div(val, step)), step);
-
-					updatePosition(newValue, activeThumbIdx);
+					const min = $min;
+					const actualMax = $actualMax;
+
+					const currentStep = Math.floor((val - min) / step);
+					const midpointOfCurrentStep = min + currentStep * step + step / 2;
+					const midpointOfNextStep = min + (currentStep + 1) * step + step / 2;
+					const newValue =
```
darkresq14, line 362: I found an extra issue with dragging the slider where it could land on incorrect ticks. I fixed it and ensured that the change happened right between 2 ticks.

src/lib/builders/slider/create.ts
TGlide: > I gave it a lot of thought if `actualMin` is relevant or not. At the moment the native range input doesn't do defaultValue like this. For example if you have min 0, step 3 and set value to 2, the actual value will be 3 from the start. For our case if we do that, the initialValue shows 2 but when we move it the ticks are always starting from min so it will be either 0 or 3. This is why `actualMin` is not relevant imho.
> 
> What is your suggestion for this case? Btw I am very new to os contributing so I'll take all the tips I can get :)

I see! We should follow native behaviour IMO.

PR #660, opened 2023-10-22
Avoid calling openTooltip on every mousemove.

pauldemarco:
openTooltip is already being called in pointerevent and focus. The effect should only be concerned with closing the tooltip once the pointer leaves the grace area.

This fixes a couple issues:
1. onOpenChange was being called for every mouse move within the tooltip.
2. Overlapping grace areas were fighting over the group.

You can see the bug occurring in the Tooltip example here:
https://github.com/pauldemarco/melt-ui/tree/tooltip-group-bug

### DIFF
--- a/.changeset/rotten-worms-warn.md
+++ b/.changeset/rotten-worms-warn.md
@@ -1,0 +1,8 @@
+---
+"@melt-ui/svelte": patch
+---
+
+Tooltip: Avoid calling openTooltip on every mousemove, fixing a couple of bugs:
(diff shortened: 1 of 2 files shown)


TGlide:
This "breaks" when closeDelay is set to a value different than 0:

https://github.com/melt-ui/melt-ui/assets/26071571/ece68353-571e-4d0f-a46f-55b6d8aa5497

Although maybe this is fine, and even better than what we had before.

PR #663, opened 2023-10-24
Controlled `openFocus` & `closeFocus` props

huntabyte:
This PR introduces two new props, `openFocus`, and `closeFocus` to a few of the builders where you might want to override how we handle focus on open and close.

They accept a `FocusProp` type, which looks like this:
```ts
export type FocusTarget = string | HTMLElement | SVGElement | null;
export type FocusProp = FocusTarget | ((defaultEl?: HTMLElement) => FocusTarget);
```

Whatever is returned from this function will be focused instead of our default behavior. If `null` is passed, we don't focus on anything and expect you to handle it outside of the builder altogether.

We provide the `defaultEl` argument as a fallback mechanism, which could be implemented like so:

```ts
const customOpenFocus: FocusProp = (defaultEl) => {
 	const customElToFocus = document.getElementById('focusMe')
	if (!customElToFocus) {
		return defaultEl
	}
	return customElToFocus
}
```

In this scenario, if we aren't able to select our custom element, we'll fall back to the default, ensuring we keep the a11y box checked.

The following builders have both the `openFocus` and `closeFocus` props:
- Dialog
- Popover

These builders only have the `closeFocus`:
- Dropdown Menu
- Menubar (`createMenu` returned from the Menubar)
- Context Menu

This PR also removes the need for the `disableTriggerRefocus` prop added to the menus in a previous PR.


src/lib/builders/menu/create.ts
```diff
@@ -1099,8 +1101,17 @@ export function createMenuBuilder(opts: _MenuBuilderOptions) {
...
-				handleRovingFocus($rootActiveTrigger);
+			const $closeFocus = get(closeFocus);
+
+			if (!$rootOpen && $rootActiveTrigger) {
+				handleFocus({ prop: $closeFocus, defaultEl: $rootActiveTrigger, roving: true });
+			}
+
+			if (!$rootOpen && !$rootActiveTrigger) {
+				const triggerEl = document.getElementById(rootIds.trigger);
+				if (!triggerEl) return;
```
TGlide, line 1112: Same here

src/lib/builders/popover/create.ts
```diff
@@ -247,6 +253,11 @@ export function createPopover(args?: CreatePopoverProps) {
 			if ($preventScroll) {
 				unsubs.push(removeScroll());
 			}
+
+			const triggerEl = $activeTrigger ?? document.getElementById(ids.trigger);
+			if (triggerEl) {
```
TGlide, line 258: Same here

src/lib/internal/helpers/focus.ts
```diff
@@ -0,0 +1,39 @@
...
+	sleep,
+} from '$lib/internal/helpers/index.js';
+
+export type FocusTarget = string | HTMLElement | SVGElement | null;
+export type FocusProp = FocusTarget | ((defaultEl?: HTMLElement) => FocusTarget);
+
+type HandleFocusArgs = {
+	prop?: FocusProp;
+	defaultEl: HTMLElement;
+	roving?: boolean;
```
TGlide, line 14: Hmm, this is a bit confusing to me. Why do we deal with roving here as well? And why only with defaultEls?

src/lib/internal/helpers/rovingFocus.ts
```diff
@@ -10,16 +10,18 @@ import { sleep } from './sleep.js';
...
-	currentFocusedElement.tabIndex = -1;
-	if (nextElement) {
-		nextElement.tabIndex = 0;
-		sleep(1).then(() => nextElement.focus());
-	}
+	sleep(1).then(() => {
+		const currentFocusedElement = document.activeElement;
+		if (!isHTMLElement(currentFocusedElement)) return;
+
+		if (currentFocusedElement === nextElement) return;
```
TGlide, line 17: You can probably put them in the same if statement

TGlide (changes requested):

src/lib/internal/helpers/focus.ts
huntabyte: Roving focus is for the menus, and we only modify the tab index's for elements we own/control, which is why it's only for the default el :)

src/lib/builders/dialog/create.ts
```diff
@@ -324,6 +337,13 @@ export function createDialog(props?: CreateDialogProps) {
 
 		if ($preventScroll && $open) unsubs.push(removeScroll());
 
+		if ($open) {
+			const contentEl = document.getElementById(ids.content);
+			if (contentEl) {
+				handleFocus({ prop: get(openFocus), defaultEl: contentEl });
+			}
```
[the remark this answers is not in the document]
huntabyte: ✅

src/lib/builders/dialog/create.ts
```diff
@@ -83,11 +95,11 @@ export function createDialog(props?: CreateDialogProps) {
 	function handleClose() {
 		open.set(false);
 		const triggerEl = document.getElementById(ids.trigger);
-		if (triggerEl) {
-			tick().then(() => {
-				triggerEl.focus();
-			});
-		}
+		if (!triggerEl) return;
```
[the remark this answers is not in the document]
huntabyte: ✅

src/lib/builders/menu/create.ts
```diff
@@ -1099,8 +1101,17 @@ export function createMenuBuilder(opts: _MenuBuilderOptions) {
 				unsubs.push(removeScroll());
 			}
 
-			if (!$rootOpen && $rootActiveTrigger && opts.disableTriggerRefocus === false) {
-				handleRovingFocus($rootActiveTrigger);
+			const $closeFocus = get(closeFocus);
+
+			if (!$rootOpen && $rootActiveTrigger) {
```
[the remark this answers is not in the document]
huntabyte: Do need to check here as if it doesn't exist we want do get the element in the next conditional

TGlide (approved):

PR #666, opened 2023-10-25
Fix: Tooltip ghost hitbox in docs

toiglak:
Current behavior:

wrong

Fixed behavior:

correct


TGlide:
Thank you, nice catch!

PR #676, opened 2023-10-27
Menu & Listbox: improvements, bug fixes, tests

huntabyte:
- Fixes a bug where we would refocus the subtrigger even if focus was within its submenu (double focusing).
- Add controlled `open` prop to submenus.
- Fixes a bug where the menus would not respect the `closeOnEscape` prop when it is set to false due to a submenu listener.
- Add tests for the `closeOnEscape` and `closeOnOutsideClick` props.
- Remove incorrect aria attributes from the context menu's trigger
- Fixed a bug where the `listbox` wasn't respecting the `closeOnEscape` prop & added a test for it

### DIFF
--- a/.changeset/great-wasps-kneel.md
+++ b/.changeset/great-wasps-kneel.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+add controlled submenu props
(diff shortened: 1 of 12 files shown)


.changeset/great-wasps-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+add controlled submenu props
```
TGlide, line 5:
```suggestion
Menu: add controlled submenu props
Listbox: Respect `closeOnEscape`
```

TGlide (approved):

PR #693, opened 2023-11-05
fix: timing issue with `handleFocus`

huntabyte:
Since elements like the popover and other floating/conditionally visible elements do not exist in the DOM until after they are opened, there is a timing issue where it is possible that the element you wish to focus on open/close is not yet in the DOM which results in unexpected behavior.

This PR moves the execution of the `FocusTarget` function inside of the `sleep()` promise to ensure we give the DOM a chance to update before calling that function.

### DIFF
--- a/.changeset/green-worms-punch.md
+++ b/.changeset/green-worms-punch.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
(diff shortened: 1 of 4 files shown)


huntabyte:
Test failing due to issue with Slider, unrelated to this PR.

TGlide (approved):

PR #697, opened 2023-11-06
Use properly cased hidden input element name

iolyd:
Fix wrong element name in the builder data of `Select` docs.


TGlide (approved):
Thank you!

PR #699, opened 2023-11-07
Minor corrections to docs

DePasqualeOrg:


src/docs/content/builders/pin-input.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Pin Input
+title: PIN Input
```
TGlide, line 2: Pin Input is fine IMO

src/docs/content/builders/dialog.md
```diff
@@ -143,7 +143,7 @@ Dialogs can be nested. For example, here's a dialog that opens another dialog.
 ## Accessibility
 
 Adheres to the
-[Dialog WAI-ARIA design pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/) &
+[Dialog WAI-ARIA design pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/) and
```
TGlide, line 146: We can keep the &s. Makes it quicker to read

src/docs/content/builders/pin-input.md
DePasqualeOrg: "Pin" is a word in English with a separate meaning. The correct way to write this is PIN.

src/docs/content/builders/dialog.md
DePasqualeOrg: It's non-standard and inconsistent.

src/docs/content/builders/pin-input.md
TGlide: If we changed it to PIN, we'd also have to rename the method to `createPINInput`, which is quite hard to read. 

Even if Pin is not the most correct form, it's easily understandable in context. We can keep it as it was before.

src/docs/content/builders/dialog.md
TGlide: I don't mind 🙂

src/docs/content/builders/pin-input.md
DePasqualeOrg: You don't have to rename the method. A common convention with acronyms in camel case is to capitalize only the first letter of the acronym.

TGlide: No qualms there! PIN Input it is. Just make sure the sidebar also reads PIN Input 😄 

Also, maybe it'd make sense to rename this to OTP Input in the future. Thoughts @huntabyte ?

huntabyte: I think PIN is more common & recognizable atm!

Issue #703, opened 2023-11-07
🐛Bug: Context Menu won't close


diaasaur:
**Describe the bug**
Context Menu does not close after being opened, tried clicking outside the menu, using esc key and selecting a menu item, nothing worked

**To Reproduce**
1. Go to `Context Menu` on docs
2. Right click on the trigger to open the menu

**Expected behavior**
It should close either on click outside or by using esc key or just by selecting one of the options

**Desktop (please complete the following information):**
 - OS: macOS Ventura
 - Browser: Safari
 - Version: 16.5.2

PR #704, opened 2023-11-07
fix [object Object] ids on trigger and menu

diaasaur:
fixes: #703 

noticed this error in browser console and it looks like an issue of value mismatch for ids (trigger + menu)
Screenshot 2023-11-08 at 1 24 48 AM

### DIFF
--- a/src/lib/builders/context-menu/create.ts
+++ b/src/lib/builders/context-menu/create.ts
@@ -114,7 +114,7 @@ export function createContextMenu(props?: CreateContextMenuProps) {
 		const target = e.target;
 		if (!(target instanceof Element)) return;
 
-		const isClickInsideTrigger = target.closest(`#${ids.trigger}`) !== null;
+		const isClickInsideTrigger = target.closest(`#${get(ids.trigger)}`) !== null;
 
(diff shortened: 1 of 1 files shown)


src/lib/builders/context-menu/create.ts
```diff
@@ -138,8 +138,8 @@ export function createContextMenu(props?: CreateContextMenuProps) {
 				style: styleToString({
 					display: $isVisible ? undefined : 'none',
 				}),
-				id: ids.menu,
-				'aria-labelledby': ids.trigger,
+				id: get(ids.menu),
```
TGlide, line 141: These should be reactive, meaning you should reference ids.menu in `stores`, then use it in `returned`.

TGlide (changes requested):
Thank you! Just one thing that needs to be fixed.

huntabyte:
Just now seeing this but I already took care of this as well as the submenus and some other issues + tests in #707

diaasaur:
> Just now saw this but I already took care of this as well as the submenus and some other issues (like the screenshot of the DOM Exception) + tests in #707

no worries! I will close mine since your PR seems to take care of tests as well :D

Issue #702, opened 2023-11-07
🐛Bug: using esc key to close a dialog ( and the like ) exits fullscreen on safari (fullscreen -> minimized)


diaasaur:
**Describe the bug**
Trying to close an opened `Dialog`, `Combobox`, `Popover` ( anything with a dropdown/popover? ) using `esc` key has this side effect on safari where it goes from fullscreen to minimized window. 

**To Reproduce**
1. On safari, go to `Dialog` on docs
2. Open the dialog
3. Try to close it using `esc`

**Expected behavior**
Dialog and the like should close without safari exiting from fullscreen

**Desktop (please complete the following information):**
 - OS: macOS Ventura
 - Browser: Safari
 - Version: 16.5.2

PR #705, opened 2023-11-07
fix: prevent safari from minimizing when esc on dialogs

diaasaur:
fixes: #702

### DIFF
--- a/src/lib/internal/actions/escape-keydown/action.ts
+++ b/src/lib/internal/actions/escape-keydown/action.ts
@@ -55,6 +55,8 @@ export const useEscapeKeydown = (node: HTMLElement, config: EscapeKeydownConfig 
 			return;
 		}
 
+		e.preventDefault();
+
 		// If an ignore function is passed, check if it returns true
 		if (options.ignore) {
 			if (isFunction(options.ignore)) {
--- a/.changeset/itchy-drinks-taste.md
+++ b/.changeset/itchy-drinks-taste.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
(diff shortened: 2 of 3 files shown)


src/lib/internal/actions/escape-keydown/action.ts
```diff
@@ -55,6 +55,8 @@ export const useEscapeKeydown = (node: HTMLElement, config: EscapeKeydownConfig
 			return;
 		}
 
+		e.preventDefault();
```
huntabyte, line 58: This can't exist by itself without making changes elsewhere, as we're checking for default prevention elsewhere on this event, like [here](https://github.com/melt-ui/melt-ui/blob/bdf8bf25531b2c17bc9b0520b3300e0938e18577/src/lib/internal/actions/popper/action.ts#L79-L90) where if the event is `defaultPrevented` we _don't_ close the popper element.

huntabyte (changes requested):
Just needs the check referenced in my other comment to be removed.

src/lib/internal/actions/escape-keydown/action.ts
diaasaur: thanks! I noticed this but wasn't really sure if I can safely remove this check 😮‍💨

huntabyte: If something breaks the test should fail, otherwise it's a great chance to identify which tests we're lacking! But after taking another look it appears that the link preview was the only one depending on it being handled there.

This is something we can address with #708.

huntabyte (approved):
Thanks for taking the time to tackle this!

TGlide (approved):

TGlide:
Thank you!!! 🎉

PR #706, opened 2023-11-08
Fix: Dialog layout shift when using svelte transitions

huntabyte:
This PR fixes an issue where the layout would shift due to the scrollbar being re-added before the outro animation/transition was complete.

It does not address/attempt to address the issue for those _not_ using `forceVisible`, as this will require extra effort and thought into how we're handling `display: none` depending on `open`.

### What changed?

Previously, regardless of `forceVisible`, we were removing the scroll as soon as `open = true`, and adding it back as soon as `open = false`.

Now, we still remove it as soon as `open = true`, however, for those using `forceVisible`, since the content conditionally rendered based on the value of `open`, we're unsubbing from the scroll removal in the `destroy` of the `content`, which is called _after_ all transitions/animations have completed.

### DIFF
--- a/src/docs/previews/dialog/alert/css/index.svelte
+++ b/src/docs/previews/dialog/alert/css/index.svelte
@@ -17,6 +17,7 @@
 		states: { open },
 	} = createDialog({
 		role: 'alertdialog',
+		forceVisible: true,
 	});
 </script>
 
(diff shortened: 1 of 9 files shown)


TGlide:
GENIUS

PR #718, opened 2023-11-09
Add css example to pin-input docs

jorri11:
Added css example to the pin-input docs. Same as the tailwind styles


TGlide:
Closed since there's a sep branch for css examples which follows quite a different convention. Thank you nonetheless!

PR #727, opened 2023-11-10
Docs: Dialog fixes

### LINKED ISSUE #715: 📁Docs: errors in anatomy section example
The Anatomy section example from the Dialog builder contains two errors:

1. `portal` should be changed to `portalled`
2. The open state should be imported from states

## The first example which is correct

All other examples appear to be correct.

symsmith:
Closes #715 and another unreported issue (missing `portalled` element in the [API reference](https://www.melt-ui.com/docs/builders/dialog#:~:text=for%20the%20elements.-,Elements,-Open%20popover))


huntabyte (approved):
Thanks for this!

TGlide (approved):

PR #738, opened 2023-11-14
💡 Light mode

### LINKED ISSUE #512: 📁Docs: Add light mode for docs site?
Dark mode is cool, but light mode is essential for a daily-basis docs site.

symsmith:
This PR adds light mode for the documentation. Closes #512

Added:
- A theme switcher in the top-right corner, to switch between dark (default, to keep current behavior for existing users), light and system mode. The system mode chooses depending on the system of the user. The selected mode is written to localStorage, managed via [mode-watcher](https://github.com/huntabyte/mode-watcher).
- Global CSS variables for every color used, that change according to the current theme.
- Tailwind colors based on these variables, with utility functions to map CSS variables to Tailwind colors.
- A special `force-dark` class, that when applied to a DOM element, will use the dark CSS variables for all of its children. This allows to easily escape light mode, if the behavior of dark mode was already good for light mode (this is the case for the main previews).
- The `force-dark` class is removed from the code displayed in the documentation, even though it is technically present in the file.

Feel free to suggest any change needed to the styling of the light mode!


huntabyte:
Hey @symsmith, just now taking a look at this! Would something like https://github.com/huntabyte/mode-watcher clean things up a bit/add more functionality around user prefers mode/system prefers mode?

symsmith:
Hey @huntabyte, I implemented mode-watcher, but I can’t figure out how to set the mode by default (on first visit) to dark (to keep current behavior for existing users), is there a way to do this?

huntabyte:
> Hey @huntabyte, I implemented mode-watcher, but I can’t figure out how to set the mode by default (on first visit) to dark (to keep current behavior for existing users), is there a way to do this?

I don't think that should be necessary. If someone's system prefers light then its likely they'd prefer light for Melt UI as well 😁

symsmith:
Makes sense!

symsmith:
Alright, so how would we do it 😄

huntabyte:
In that case it'll probably be a bit hacky as mode watcher goes with the system or browser prefers in the `svelte:head` tag if nothing is already set, which is how it prevents the FOUC. I can try to investigate when I find some time

TGlide:
> > Hey @huntabyte, I implemented mode-watcher, but I can’t figure out how to set the mode by default (on first visit) to dark (to keep current behavior for existing users), is there a way to do this?
> 
> I don't think that should be necessary. If someone's system prefers light then its likely they'd prefer light for Melt UI as well 😁

Unfortunately when people do not set defaults, the browser sets it to light theme. It should be fine, but eeeeeh maybe we can make it default to dark, as long as the theme selector is pretty visible.

ollema:
@symsmith I have an idea for how this could potentially be solves within mode-watcher. let me investigate!

ollema:
I have a solution up now which seems to work, see https://github.com/huntabyte/mode-watcher/pull/41

but it's a bit hacky 😅

symsmith:
Nice, that was fast! I'll push my changes tonight 😀

symsmith:
I’ve added the default theme

package.json
```diff
@@ -68,6 +68,7 @@
 		"@internationalized/date": "^3.5.0",
 		"dequal": "^2.0.3",
 		"focus-trap": "^7.5.2",
+		"mode-watcher": "^0.1.2",
```
TGlide, line 71: Should be a devDependency, otherwise when people install melt this'll be installed too

src/docs/components/code-block.svelte
```diff
@@ -15,29 +15,41 @@
...
+	class={cn($$restProps.class, 'force-dark')}
+	{...$$restProps}
+	data-rehype-pretty-code-fragment
+	use:preTabIndex
+>
 	<slot />
 </div>
 <button
-	class={cn('absolute right-4 top-12 z-10', copyBtnClasses)}
+	class={cn('force-dark absolute right-4 top-12 z-10 text-white', copyBtnClasses)}
```
TGlide, line 28: Would be nice to see the code blocks follow the theme too, but this is a great start :)

src/docs/utils/preview.ts
```diff
@@ -60,10 +60,16 @@ async function createPreviewsObject({
 
 		if (match) {
 			const [, groupKey, styleKey, fileKey] = match; // Destructure the matched parts
-			const { content } = obj;
+			const { content: clutteredContent } = obj;
 
 			if (!isSvelteFile(fileKey)) return;
 
+			const content = clutteredContent
```
TGlide, line 67: Niceee

src/docs/components/site-header.svelte
```diff
@@ -4,20 +4,22 @@
...
-		'container flex h-14 items-center rounded-md bg-neutral-800',
-		isRoot && 'md:bg-transparent'
+		'container flex h-14 items-center rounded-md bg-neutral-800 text-white !outline outline-2 !outline-neutral-700 dark:!outline-none',
+		isRoot && '!outline-0 md:bg-transparent'
 	)}
 >
 	<div class="flex">
 		<a href="/" class="mr-6 flex items-center transition-opacity hover:opacity-75">
-			<Logo class="h-9" withText textColor="white" />
+			<Logo class="h-9" withText textColor={$mode === 'light' ? 'black' : 'white'} />
```
ollema, line 22: sorry for being so late to comment on this, I was busy with IRL yesterday.

for FOUC prevention. this works for dark mode users, but I'm not sure it will prevent FOUC for light mode users.

I think the best way in general to do stuff like this that can cause FOUC is to not rely on stores (since they load later than the FOUC prevention snippet) and instead rely on presence of the `dark` class.

for example:
```svelte
<Logo class="h-9 text-black block dark:hidden" withText />
<Logo class="h-9 text-white hidden dark:block" withText />
```

you can use `$mode` and the other stores to interact with the theme switching but to fully avoid FOUC it's not great.

I think the current implementation above will cause FOUC if the user has selected light mode and they do a full refresh.

PR #739, opened 2023-11-15
Fix: Close on ESC in Select prevents closing Dialog 

### LINKED ISSUE #700: 🐛Bug: Close on ESC in Select prevents closing Dialog
**Describe the bug**
Dialogs can't be closed with ESC when focus is on a Select.

**To Reproduce**
1. Create a Dialog
2. Add a Select to it
3. Open the Dialog
4. Click ESC key

**Expected behavior**
I expect the ESC key to close the Select if it's open but to close the Dialog if the Select is already closed.
If I understand [these WAI guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/) right, pressing ESC on a closed Select/Combobox shouldn't do anything (like preventing closing a modal).

**Screenshots**
N/A

 - OS: macOS
 - Browser: Chrome, Safari, Arc
 - Version:

diaasaur:
closes #700 

Tweaked `useEscapeKeydown` to check for the node closest to the target that has both `data-escapee` and is open (`data-state='open'`).  This allows Dialog to close when the Select inside it is already closed (without the user needing to click away/removing focus from Select's trigger) and other similar cases with nested escape-able elements

### DIFF
--- a/src/lib/builders/listbox/create.ts
+++ b/src/lib/builders/listbox/create.ts
@@ -261,6 +261,7 @@ export function createListbox<
 				'aria-expanded': $open,
 				'aria-labelledby': $labelId,
 				// autocomplete: 'off',
+				'data-state': $open ? 'open' : 'closed',
 				id: $triggerId,
 				role: 'combobox',
 				disabled: disabledAttr($disabled),
--- a/src/lib/builders/listbox/create.ts
+++ b/src/lib/builders/listbox/create.ts
@@ -444,6 +444,7 @@ export function createListbox<
 				id: $menuId,
 				role: 'listbox',
 				style: styleToString({ display: $isVisible ? undefined : 'none' }),
+				'data-state': $isVisible ? 'open' : 'closed',
 			} as const;
 		},
 		action: (node: HTMLElement): MeltActionReturn<ListboxEvents['menu']> => {
--- a/src/lib/internal/actions/escape-keydown/action.ts
+++ b/src/lib/internal/actions/escape-keydown/action.ts
@@ -51,7 +51,7 @@ export const useEscapeKeydown = (node: HTMLElement, config: EscapeKeydownConfig 
 		if (!e || !isEnabled()) return;
 		const target = e.target;
 
-		if (!isHTMLElement(target) || target.closest('[data-escapee]') !== node) {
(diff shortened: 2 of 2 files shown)


diaasaur:
> do the other elements that rely on this behavior already have the `data-state` as well? 

Yes!  Context Menu, Link Preview, Menu, Menu Bar, Popover (all use popper) already have `data-state` on trigger and content/menu. 
Adding Dialog to the list as well, Tooltip while not using popper seems to use portal but is handling esc on it's own without `useEscapeKeydown`

> We have to be careful when modifying the logic within the actions as they are used throughout various other floating components via the popper action.

makes sense, how can I make sure this doesn't break anything else? (other than checking if tests for all these elements are testing for escape on close)

huntabyte:
If you're up for it, I'd say it would be a good opportunity to add some additional test cases for "within dialog" to ensure we're getting the right behavior. If not I can try to add them to this PR when I have the chance.

Basically for all the floating elements, we want to ensure that pressing escape, outside clicks, etc. all work as they should _inside_ a dialog where it only closes the top layer! This will prevent us from accidentally regressing in the future.

TGlide:
> If you're up for it, I'd say it would be a good opportunity to add some additional test cases for "within dialog" to ensure we're getting the right behavior. If not I can try to add them to this PR when I have the chance.
> 
> Basically for all the floating elements, we want to ensure that pressing escape, outside clicks, etc. all work as they should _inside_ a dialog where it only closes the top layer! This will prevent us from accidentally regressing in the future.

We have the nested tests currently. It can be added there I presume

TGlide:
> closes #700
> 
> Tweaked `useEscapeKeydown` to check for the node closest to the target that has both `data-escapee` and is open (`data-state='open'`). This allows Dialog to close when the Select inside it is already closed (without the user needing to click away/removing focus from Select's trigger) and other similar cases with nested escape-able elements

Also @diaasaur instead of checking for data-state="open", we can probably check for element visibility 🤔

TGlide (commented):
Rather than depend on data-state, which not all builders have, we can disable the handling whenever it is open. onEscapeKeydown accepts a readable `enabled` store for these purposes.

PR #742, opened 2023-11-15
Prevent TOC content from being unreachable (overflow below screen edge)

lolcabanon:
Changed `overflow-visible` to `overflow-x-visible overflow-y-auto` to prevent TOC content (Calendar builder and maybe more?) from overflowing at the bottom of the screen and become unreachable. (1920x1080px screen)

Screenshot :
image

### DIFF
--- a/src/routes/docs/builders/[name]/+page.svelte
+++ b/src/routes/docs/builders/[name]/+page.svelte
@@ -49,7 +49,7 @@
 		<!-- <DocsPager /> -->
 	</div>
 	<div class="hidden text-sm xl:block">
-		<div class="fixed top-16 h-[calc(100vh-4rem)] overflow-visible pt-6">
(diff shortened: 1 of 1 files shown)


huntabyte:
Hey @lolcabanon, CI is failing due to formatting, so just run `pnpm format` to clean that up.

One thing I'm noticing is that we need more space on the right side as the scrollbar is touching the last letter:
CleanShot 2023-11-15 at 10 40 15@2x

Perhaps making the TOC take up the full width of its area would be best to ensure the scrollbar is consistent as you navigate between multiple pages that have a scrollbar for the TOC.

huntabyte (changes requested):
See the comment I left on the PR!

TGlide (approved):
Thank you!

PR #743, opened 2023-11-15
Added plugin tailwind-scrollbar with examples

lolcabanon:
Added pretty scrollbars to some inner components with [tailwind-scrollbar](https://github.com/adoxography/tailwind-scrollbar) plugin.

Two examples : sidebar nav and TOC (see [Calendar builder](https://www.melt-ui.com/docs/builders/calendar) for scrollable TOC) :

- Sidebar uses `scrollbar scrollbar-w-2`
- TOC uses `scrollbar-thin`

I prefer the sidebar one (pretty much the same on Chromium based browser, but nicer on Firefox).

Check the doc for more info on the plugin : https://adoxography.github.io/tailwind-scrollbar/examples

Screenshots : 

Chrome :
image

Firefox :
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,6 +143,7 @@
 		"svelte-preprocess": "^5.0.4",
 		"svelte-sequential-preprocessor": "^2.0.1",
 		"tailwind-merge": "^1.14.0",
+		"tailwind-scrollbar": "^3.0.5",
 		"tailwind-variants": "^0.1.8",
 		"tailwindcss": "^3.3.3",
 		"tslib": "^2.6.1",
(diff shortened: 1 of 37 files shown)


TGlide:
Thank you for the PR! We used to have stylized scrollbars, but I opted to remove them due to some weird browser behaviours. I also don't think it makes sense to add them to component examples using a TW plugin, since the plan is to sunset using tailwind on examples, and opting instead to only use CSS.

I'll be closing this, unfortunately, but I really do appreciate the effort!

lolcabanon:
Ahh I see! I was about to add them in CSS first but since the website is done with tailwind I tought I'd follow the path... (I'm not a tailwind user myself) Didn't knew you were going away!

Btw, I was adding examples for you to see how it would be applied **on MeltUI site**, not for example components for the lib.

Anyway, I understand, scrollbars are behaving pretty weird across different browsers and styling is unnecessary, but the default are so ugly in Chrome lol!

Thanks for your answer :) ✌️

PR #747, opened 2023-11-16
Fix: listbox 'multiple' type

leonardsimonse:
Without this change the only valid types for 'multiple' are either false or undefined.

Fixes: https://github.com/huntabyte/shadcn-svelte/issues/428

### DIFF
--- a/src/lib/builders/listbox/types.ts
+++ b/src/lib/builders/listbox/types.ts
@@ -136,7 +136,7 @@ export type CreateListboxProps<
 	 */
 	forceVisible?: boolean;
 
-	multiple?: Multiple;
+	multiple?: boolean;
 
 	/**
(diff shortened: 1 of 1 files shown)


TGlide:
That is not the case: 
Screenshot 2023-11-20 at 16 46 47

I suspect the issue lies elsewhere. Probably someone is trying to do:
Screenshot 2023-11-20 at 16 47 19

In which case there is a simple fix:
Screenshot 2023-11-20 at 16 47 44

leonardsimonse:
Can no longer replicate this typing error in the identical codespaces (of shadcn-svelte and bits-ui) as last time 🤔. Probably down to some part of the tooling getting confused. Thanks for looking into this though!

PR #748, opened 2023-11-18
Fix/calendar keyboard navigation

ollema:
this should do the trick, but...

unfortunately, I could not get the test to work. I'm not really sure why. based on the `toHaveFocus()` tests it seems like pressing the down arrow key works great at first, you get:

- 1 January
- 8 January
- 15 January
- 22 January
- 29 January

and then I would expect to see:
- 5 February

but for some reason the focused date is instead:
- 5 January

no idea why... am I doing something stupid @huntabyte?

so for that reason the added test is not working yet. if I figure out how to do it correctly, the test should pass with this patch and fail without.

the change itself is working as intended in the dev preview

### DIFF
--- a/src/lib/builders/calendar/create.ts
+++ b/src/lib/builders/calendar/create.ts
@@ -957,7 +957,8 @@ import { generateIds } from '../../internal/helpers/id'
 			 * of the array.
 			 */
 
-			// shift the calendar back a month
+			// shift the calendar back a month unless previous month is disabled
+			if (get(isPrevButtonDisabled)) return;
 
 			const $months = get(months);
 			const firstMonth = $months[0].value;
@@ -994,6 +995,9 @@ import { generateIds } from '../../internal/helpers/id'
 			 * month, and then starting at the beginning of that array,
 			 * shift focus by the nextIndex amount.
 			 */
+
+			// shift the calendar forward a month unless next month is disabled
+			if (get(isNextButtonDisabled)) return;
 
 			const $months = get(months);
 			const firstMonth = $months[0].value;
--- a/src/tests/calendar/Calendar.spec.ts
+++ b/src/tests/calendar/Calendar.spec.ts
@@ -410,32 +410,50 @@ describe('Calendar', () => {
 
 		// five keypresses to get to February 1980
 		await user.keyboard(`{${kbd.ARROW_DOWN}}`);
+		expect(getByTestId('month-0-date-8')).toHaveFocus();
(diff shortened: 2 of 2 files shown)


ollema:
also, I can add to test navigating upwards/back in time as well if needed :) assuming we can get the current test to work

huntabyte:
Keyboard tests that trigger reactive state updates are a bit tricky sometimes. I have a busy weekend but will try to find some time to investigate further. From the looks of it the test _should_ pass, however, we don't want to merge this until we have a working test of course!

huntabyte:
One thing I notice is that the calendar tests still have the bass ackwards way of setting up the `user` for most of the suite. It should be `const user = userEvent.setup()` before the render in each test, I'm not 100% sure if this has something to do with it, but testing library strongly recommends this approach.

We should also probably add a `setup` function and use the `testKbd` as well, basically similar to how we're doing things in: https://github.com/melt-ui/melt-ui/blob/develop/src/tests/collapsible/Collapsible.spec.ts

Don't feel obligated to take this on if you don't have the time @ollema , just adding some notes here for myself to clean up which may resolve this odd behavior.

ollema:
> Keyboard tests that trigger reactive state updates are a bit tricky sometimes. I have a busy weekend but will try to find some time to investigate further. From the looks of it the test should pass, however, we don't want to merge this until we have a working test of course!

sounds good. it's weird because the state is updating, until a certain point where it does not switch month for some unknown (to me) reason

> One thing I notice is that the calendar tests still have the bass ackwards way of setting up the `user` for most of the suite. It should be `const user = userEvent.setup()` before the render in each test, I'm not 100% sure if this has something to do with it, but testing library strongly recommends this approach.

the user is setup before the render in the test -  I copied from earlier examples

huntabyte:
Alright I'm starting to investigate this a bit further, it seems `month-0-date-5` is actually still in January as well.

huntabyte:
I'm a bit stumped on this one, mind putting a second set of eyes on this @TGlide? It does in fact function as expected in the preview if you apply those same props, however, the test doesn't seem to actually flip the month when navigating through them 😠

ollema:
did you manage to get it to work @huntabyte? amazing? what was the trick?

want me to add tests for the other directions as well?

huntabyte:
@ollema, actually @TGlide decided to do the obvious and visual test using the test component, and the issue was with how we were passing `undefined` props, rather than removing them.

It should all be cleaned up and good to go once it goes through CI 🥲

PR #757, opened 2023-11-21
fix: `ids` prop is not `Partial`

abdel-17:
To override ids on some components, you have to provide ids for all elements. This PR fixes that by making the `ids` prop `Partial`.

### DIFF
--- a/src/lib/builders/tooltip/types.ts
+++ b/src/lib/builders/tooltip/types.ts
@@ -32,7 +32,7 @@ export type CreateTooltipProps = {
 	/**
 	 * Optionally override the default ids we assign to the elements
 	 */
-	ids?: IdObj<TooltipIdParts>;
+	ids?: Partial<IdObj<TooltipIdParts>>;
 };
 
 export type Tooltip = BuilderReturn<typeof createTooltip>;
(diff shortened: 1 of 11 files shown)


src/lib/builders/tooltip/types.ts
```diff
@@ -32,7 +32,7 @@ export type CreateTooltipProps = {
 	/**
 	 * Optionally override the default ids we assign to the elements
 	 */
-	ids?: IdObj<TooltipIdParts>;
+	ids?: Partial<IdObj<TooltipIdParts>>;
```
TGlide, line 35: You can make the IdObj itself be a partial

abdel-17: Tried it. The entire project exploded with type errors everywhere. I tried to fix them, but the calendar builder relies on the ids being required, not partial.

abdel-17: You’ll find the original commit in the history. Try it for yourself.

huntabyte: Yeah @TGlide the issue with making the `IdObj` a partial is that the type is also used internally within the builders for the `ids` object so it expects all the properties to exist.

TGlide: Yeah, it's probbly not worth the hassle and confusion. Merging!

TGlide (approved):

Issue #750, opened 2023-11-20
🐛Bug: Dialog closes on mouseup outside dialog after mousedown inside dialog


DePasqualeOrg:
**Describe the bug**
The dialog closes when the mouse is released outside the dialog after clicking inside the dialog.

**To Reproduce**
[https://www.shadcn-svelte.com/docs/components/dialog](https://www.shadcn-svelte.com/docs/components/dialog)

In the dialog example, click "Edit profile", select text in one of the input elements, and release the mouse outside the dialog.

**Expected behavior**
The dialog should only close when the click occurred outside the dialog. For example, if the user selects text with the mouse inside the dialog and the mouse is released outside the dialog, closing the dialog is not the expected behavior.

**Desktop (please complete the following information):**
 - macOS 14
 - Safari 17.1

PR #762, opened 2023-11-23
Fix: clickoutside handler

huntabyte:
Closes: #750

Previously, we were triggering an outside click on `pointerup` instead of `pointerdown` causing unexpected behavior when doing things like selecting text within a dialog. If you clicked inside and drug outside to select a bunch of stuff, the dialog would close. This is no longer an issue.

### DIFF
--- a/.changeset/early-dolphins-provide.md
+++ b/.changeset/early-dolphins-provide.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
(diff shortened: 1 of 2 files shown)


TGlide:
We changed to pointerup to fix an issue where the touch event was passing through the overlay. Simply reverting will bring back that issue.

TGlide:
P.S. While I closed the PR, if we manage to make sure #468 does not happen with `pointerdown`, I'm fine with that!

Issue #763, opened 2023-11-23
🐛Bug: Popover trigger not toggling


angelvega93:
**Describe the bug**
When the popover is open and you click the trigger, it closes and immediately reopens.

**To Reproduce**
- Click the trigger element to open/close the popover.
- Notice that it only opens, but does not close it.

**Expected behavior**
If you press the trigger and the popover is open, it should close.

PR #765, opened 2023-11-24
fix: popover trigger toggle

huntabyte:
Closes: #763 

Test case added as well.

### DIFF
--- a/.changeset/clever-bananas-cheer.md
+++ b/.changeset/clever-bananas-cheer.md
@@ -2,4 +2,4 @@
 "@melt-ui/svelte": patch
 ---
 
-fix: popover trigger toggle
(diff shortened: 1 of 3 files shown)


.changeset/clever-bananas-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+fix: popover trigger toggle
```
TGlide, line 5:
```suggestion
Popover: Fix trigger not toggling the `open` state
```

TGlide (approved):
Awesome!

angelvega93:
The problem still there if you click a child element.

chrome_iKo8ia5mpU

PR #767, opened 2023-11-24
Slider: add `data-value` attribute to `tick` and `thumb`

abdel-17:
This PR adds the `data-value` attributes to the `tick`. This improves customizability, like only showing certain ticks, for example. I also added `data-value` to the `thumb` for consistency (I'm aware `aria-valuenow` exists, just thought it'd be nice to be consistent).


TGlide (approved):
Thank you!

Issue #713, opened 2023-11-08
🐛Bug: Combobox not selecting when filtered objects contain undefined property


jorri11:
**Describe the bug**
I have a problem with an item not being set as selected if the selected object has a property that is undefined.

**To Reproduce**
Reproduction link: https://stackblitz.com/edit/github-jjgjbl?file=src%2Froutes%2F%2Bpage.svelte 

1. Open the stackblitz
2. Select any option except from "5656"
3. Open the options again and observe none is styled as selected.
4. Select option "5656" witch has no elements as undefined.
5. Open options again and observe "5656" is selected.

**Expected behavior**
Any of the available options should be selected.

**Screenshots**
5656 styled as expected:
![image](https://github.com/melt-ui/melt-ui/assets/42845131/bf96ce9b-c231-439b-ac82-5f23ded73067)

2341 styled incorrectly:
![image](https://github.com/melt-ui/melt-ui/assets/42845131/34e0c539-1c6d-4366-a85c-4d41f95cf7fa)

 - OS: Windows
 - Browser: Brave
 - Version: Version 1.60.110 Chromium: 119.0.6045.105 

**Additional context**
Not sure if its relevant but it seems like $selected.value strips away the undefined properties

jorri11:
Issue Investigation Update:

After investigation, I think ive identified the root cause of the bug. In the listbox/create.ts file, the value undergoes stringification and is stored in the 'data-value' attribute. This process results in the loss of undefined properties.

The isSelected check employs deepEqual, which returns false in scenarios like `deepEqual({a: undefined}, {}) // => false`. To address this, I propose two potential solutions:

Preserve Undefined in JSON.stringify:
Adjust the stringification process to preserve undefined values. Refer to [this Stack Overflow discussion](https://stackoverflow.com/questions/26540706/preserving-undefined-that-json-stringify-otherwise-removes) for guidance.

Remove Undefined Properties Before deepEqual:
Alternatively, we can eliminate undefined properties before performing the deep equal check in isSelected().

I can contribute a pull request with either solution. Could you provide guidance on the preferred approach, or is there any other direction you'd like to take? I'm eager to assist in resolving this issue. 😊

TGlide:
Hmm, I don't think we should adjust to make undefined the same as null. Let's remove undefined before doing the deepEqual check

PR #770, opened 2023-11-25
Remove undefined properties before deepEquals

jorri11:
Fix for issue #713

### DIFF
--- a/src/lib/internal/helpers/object.ts
+++ b/src/lib/internal/helpers/object.ts
@@ -13,10 +13,18 @@ export function omit<T extends Record<string, unknown>, K extends keyof T>(
 	return result;
 }
 
-export function removeUndefinedValues(inputObject: { [key: string]: any }): { [key: string]: any } {
+type StrippedKeys<T extends Record<string, unknown>, ToStrip> = {
+  [K in keyof T]: T[K] extends ToStrip ? never : K
+}
+
+type StripValues<T extends Record<string, unknown>, ToStrip> = {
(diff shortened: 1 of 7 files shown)


src/lib/internal/helpers/object.ts
```diff
@@ -12,3 +12,11 @@ export function omit<T extends Record<string, unknown>, K extends keyof T>(
...
 	return result;
 }
+
+export function removeUndefinedValues(inputObject: { [key: string]: any }): { [key: string]: any } {
+	return Object.fromEntries(
+		Object.entries(inputObject)
+			.filter(([key, value]) => value !== undefined)
+			.map(([key, value]) => [key, value === Object(value) ? removeUndefinedValues(value) : value])
+	);
+}
```
TGlide, line 22:
```suggestion
type StrippedKeys<T extends Record<string, unknown>, ToStrip> = {
  [K in keyof T]: T[K] extends ToStrip ? never : K
}

type StripValues<T extends Record<string, unknown>, ToStrip> = {
  [K in StrippedKeys<T, ToStrip>[keyof T]]: T[K]
}

export function removeUndefinedValues<T extends Record<string, unknown>>(inputObject: T) {
	return Object.fromEntries(
		Object.entries(inputObject)
			.filter(([key, value]) => value !== undefined)
			.map(([key, value]) => [key, value === Object(value) ? removeUndefinedValues(value) : value])
	) as StripValues<T, undefined>
}
```

PR #788, opened 2023-12-01
fix: aria-modal attr type

### LINKED ISSUE #787: 🐛Bug: Types of property `'aria-modal'` are incompatible
**Describe the bug**
TypeScript type conflict when declaring `{...$content}` `use:content` of `createModal()`

```ts
Types of property ''aria-modal'' are incompatible.
Type 'string | undefined' is not assignable to type 'Booleanish | null | undefined'.
```

These are the type declarations which enter in conflict:
* https://github.com/melt-ui/melt-ui/blob/3557427f0e175765622257ef2eff1d2482dbfbb5/src/lib/builders/dialog/create.ts#L192
* https://github.com/sveltejs/svelte/blob/b7af2ffabdda5124a95c7e218d8a650c3a6d9cbf/packages/svelte/elements.d.ts#L521

**To Reproduce**
Similar to the example of https://github.com/melt-ui/melt-ui/issues/764 but let me know if an example is necessary and I will be happy to create one.

**Expected behavior**
There is no mismatch between internal types. svelte-check keeps succeeding.

**Additional context**
* `"svelte-check"`: `"3.6.1"`
* `"svelte"`: `"4.2.7"`
* `"@sveltejs/kit"`: `"1.27.6"`
* `"@melt-ui/svelte"`: `">=0.63.1"`

🔙 Regression introduced in https://github.com/melt-ui/melt-ui/pull/769

oscard0m:
# Description
fix: `aria-modal` attr TS type matches the type of a Svelte Element

# Context
Fixes #787

### DIFF
--- a/.changeset/silent-doors-impress.md
+++ b/.changeset/silent-doors-impress.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+- Dialog: fix `aria-modal` attribute type
--- a/src/lib/builders/dialog/create.ts
+++ b/src/lib/builders/dialog/create.ts
@@ -189,7 +189,7 @@ export function createDialog(props?: CreateDialogProps) {
 				role: get(role),
 				'aria-describedby': $descriptionId,
 				'aria-labelledby': $titleId,
-				'aria-modal': $isVisible ? 'true' : undefined,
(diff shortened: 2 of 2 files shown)


huntabyte:
Thank you!

PR #791, opened 2023-12-03
Slider: RTL support

abdel-17:
Really excited about this pr! 👀

Added two more orientations for sliders, `horizontal-rl` and `vertical-tb`, effectively adding RTL support for sliders with an added bonus of top-to-bottom orientation for vertical sliders. I also documented the two orientations.

I manually tested all 4 orientations with and without ticks. Everything seems to be placed correctly and keyboard shortcuts work. I haven't added any automated tests, though.

### DIFF
--- a/src/lib/builders/slider/create.ts
+++ b/src/lib/builders/slider/create.ts
@@ -129,7 +129,7 @@ export const createSlider = (props?: CreateSliderProps) => {
 				dir: $dir,
 				disabled: disabledAttr($disabled),
 				'aria-disabled': disabledAttr($disabled),
-				'aria-orientation': $orientation,
+				// TODO: Is aria-orientation missing here?
 				'data-orientation': $orientation,
 				style: $disabled
 					? undefined
--- a/src/lib/builders/slider/create.ts
+++ b/src/lib/builders/slider/create.ts
@@ -1,12 +1,12 @@
 import {
 	addEventListener,
 	addMeltEventListener,
-	ariaDisabledAttr,
 	builder,
 	createElHelpers,
 	disabledAttr,
 	effect,
 	executeCallbacks,
+	generateIds,
 	getElementByMeltId,
 	isBrowser,
 	isHTMLElement,
@@ -19,10 +19,9 @@ import {
 } from '$lib/internal/helpers/index.js';
 import type { MeltActionReturn } from '$lib/internal/types.js';
 import { derived, get, writable } from 'svelte/store';
-import { generateIds } from '../../internal/helpers/id';
 import type { SliderEvents } from './events.js';
 
-import type { CreateSliderProps } from './types.js';
+import type { CreateSliderProps, SliderOrientation } from './types.js';
 
 const defaults = {
 	defaultValue: [],
@@ -34,6 +33,13 @@ const defaults = {
 } satisfies CreateSliderProps;
 
 const { name } = createElHelpers('slider');
+
+function isHorizontal(orientation: SliderOrientation) {
+	return orientation === 'horizontal' || orientation === 'horizontal-rl';
+}
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+function assertNever(_: never) {}
 
 export const createSlider = (props?: CreateSliderProps) => {
 	const withDefaults = { ...defaults, ...props } satisfies CreateSliderProps;
@@ -127,11 +133,12 @@ export const createSlider = (props?: CreateSliderProps) => {
 		returned: ([$disabled, $orientation]) => {
 			return {
 				disabled: disabledAttr($disabled),
-				'aria-disabled': ariaDisabledAttr($disabled),
+				'aria-disabled': disabledAttr($disabled),
+				// TODO: Is aria-orientation missing here?
 				'data-orientation': $orientation,
 				style: $disabled
 					? undefined
-					: `touch-action: ${$orientation === 'horizontal' ? 'pan-y' : 'pan-x'}`,
+					: `touch-action: ${isHorizontal($orientation) ? 'pan-y' : 'pan-x'}`,
 
 				'data-melt-id': meltIds.root,
 			};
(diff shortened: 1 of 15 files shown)


abdel-17:
Hey @huntabyte, `dir="rtl"` makes sense for horizontal sliders, but what about vertical sliders? `dir="rtl"` doesn't make sense in that context. Should we add dir="ttb"? What if the user passes `orientation="horizontal"` and `dir="ttb"`? If you have any suggestions, I'm willing to modify the code.

huntabyte:
Hm, the slider is one of the more unique ones I suppose.

I know for a fact we should have either a `dir` or `rtl` prop, likely `rtl` to keep it simple which we can apply to basically every builder as we add RTL support and ensure a consistent DX across the builders.

As for the top to bottom, I imagine this is the only/one of the only builders that it makes sense for right? If that's the case, then maybe a prop for `topToBottom`? It sounds a bit odd though and I do see what you mean, I'm just trying to ensure we have some sort of standard, as RTL support is definitely something we want across _all_ builders.

Open to suggestions from the community on this one as well!

src/lib/builders/slider/create.ts
```diff
@@ -123,11 +123,13 @@ export const createSlider = (props?: CreateSliderProps) => {
...
-		stores: [disabled, orientation],
-		returned: ([$disabled, $orientation]) => {
+		stores: [disabled, orientation, dir],
+		returned: ([$disabled, $orientation, $dir]) => {
 			return {
+				dir: $dir,
 				disabled: disabledAttr($disabled),
-				'aria-disabled': ariaDisabledAttr($disabled),
+				'aria-disabled': disabledAttr($disabled),
+				// TODO: Is aria-orientation missing here?
```
TGlide, line 132: It is!

src/lib/builders/slider/create.ts
```diff
@@ -139,28 +141,38 @@ export const createSlider = (props?: CreateSliderProps) => {
...
+			} else if ($orientation === 'horizontal') {
+				style.left = `${minimum}%`;
+				style.right = `${maximum}%`;
+			} else if ($dir === 'rtl') {
+				style.top = `${minimum}%`;
+				style.bottom = `${maximum}%`;
+			} else {
+				style.bottom = `${minimum}%`;
+				style.top = `${maximum}%`;
+			}
```
TGlide, line 165:
```suggestion
			const style: Record<string, string> = {
                          position: 'absolute',
                        };
                        
                        if ($orientation === 'horizontal') {
                          style.left = $dir === 'rtl' ? `${maximum}%` : `${minimum}%`;
                          style.right = $dir === 'rtl' ? `${minimum}%` : `${maximum}%`;
                        } else {
                          style.top = $dir === 'rtl' ? `${maximum}%` : `${minimum}%`;
                          style.bottom = $dir === 'rtl' ? `${minimum}%` : `${maximum}%`;
                        }

```

src/lib/builders/slider/create.ts
```diff
@@ -174,27 +186,39 @@ export const createSlider = (props?: CreateSliderProps) => {
...
+				} else if ($orientation === 'horizontal') {
+					style.left = thumbPosition;
+					style.translate = '-50% 0';
+				} else if ($dir === 'rtl') {
+					style.top = thumbPosition;
+					style.translate = '0 -50%';
+				} else {
+					style.bottom = thumbPosition;
+					style.translate = '0 50%';
+				}
```
TGlide, line 206:
```suggestion
const style: Record<string, string> = {
  position: 'absolute',
};

if ($orientation === 'horizontal') {
  style.left = $dir === 'rtl' ? undefined : thumbPosition;
  style.right = $dir === 'rtl' ? thumbPosition : undefined;
  style.transform = $dir === 'rtl' ? 'translate(50%, 0)' : 'translate(-50%, 0)';
} else {
  style.top = $dir === 'rtl' ? thumbPosition : undefined;
  style.bottom = $dir === 'rtl' ? undefined : thumbPosition;
  style.transform = $dir === 'rtl' ? 'translate(0, -50%)' : 'translate(0, 50%)';
}
```

Now that I think about it, why would rtl switch vertical directions?

src/lib/builders/slider/create.ts
```diff
@@ -294,31 +320,38 @@ export const createSlider = (props?: CreateSliderProps) => {
...
-				// the first tick as it would be rendered outside the slider.
-				//
-				// As for the last tick, offset it by its full size rather than
-				// half also to prevent it from being rendered outside.
-				if (index === $ticks - 1) {
-					// Left is negative, down is positive.
-					style.translate = horizontal ? '-100% 0' : '0 100%';
-				} else if (index !== 0) {
-					style.translate = horizontal ? '-50% 0' : '0 50%';
+				if ($orientation === 'horizontal' && $dir === 'rtl') {
```
TGlide, line 343: These ifs can be a bit complex too, something similar to what i've shown before would be useful 😄

TGlide (changes requested):
Thank you so much! I left some comments 🙂

src/lib/builders/slider/create.ts
TGlide: I read the comments now, ignore me :)

src/lib/builders/slider/create.ts
abdel-17: So, about that. I did some research and apparently the `aria-disabled` and `aria-orientation` attributes should be on the thumb, not the root. That's a breaking change.

src/lib/builders/slider/create.ts
abdel-17: > These ifs can be a bit complex too, something similar to what i've shown before would be useful 😄

I'm not the biggest fan of ternaries, personally, but I'll fix the ifs. I wish JS had native pattern matching similar or switch statements with multiple values.

abdel-17: Hey! So I just pushed the changes. I chose a different approach from ternary statements. If you still prefer ternaries, I have no problem refactoring the code, but I think the new changes are much more readable than both the ternaries and the if statements.

src/lib/builders/slider/create.ts
TGlide: No worries, we can change these.  We can keep the data-attributes on the root as well though

src/lib/builders/slider/create.ts
```diff
@@ -123,11 +132,16 @@ export const createSlider = (props?: CreateSliderProps) => {
...
-		returned: ([$disabled, $orientation]) => {
+		stores: [disabled, orientation, dir],
+		returned: ([$disabled, $orientation, $dir]) => {
 			return {
+				dir: $dir,
 				disabled: disabledAttr($disabled),
-				'aria-disabled': ariaDisabledAttr($disabled),
+				// TODO: Consider removing `aria-disabled` from here.
+				// `aria-disabled` doesn't make sense on the root
+				// because the slider `role` is on the thumb.
```
TGlide, line 142: Let's do it, we can keep the data attribute

src/lib/builders/slider/create.ts
TGlide: I do prefer the ternaries, but the direction derived with the switch case is pretty nice too 😄 You can keep it!

Issue #803, opened 2023-12-06
🐛Bug: Listbox - Meta/Alt/Control keys used as filter 


huntabyte:
**Describe the bug**
Since e.key === "Meta", etc. we're filtering based on that, which should certainly be ignored.

reference: https://github.com/huntabyte/bits-ui/issues/216

PR #804, opened 2023-12-06
fix: Listbox - ignore certain typeahead keys

huntabyte:
Closes: #803

### DIFF
--- a/src/lib/internal/helpers/typeahead.ts
+++ b/src/lib/internal/helpers/typeahead.ts
@@ -27,6 +27,13 @@ export type HandleTypeaheadSearch = {
 };
 
 /**
+ * Keys to ignore for typeahead so we aren't matching things
+ * like `Shift menu item` or `Control center` or `Alt menu` when
+ * a user presses those keys.
+ */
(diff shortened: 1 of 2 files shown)


src/lib/internal/helpers/typeahead.ts
```diff
@@ -26,6 +26,13 @@ export type HandleTypeaheadSearch = {
 	 */
 };
 
+/**
+ * Keys to ignore for typeahead so we aren't matching things
+ * like `Shift menu item` or `Control center` or `Alt menu` when
+ * a user presses those keys.
+ */
+const ignoredKeys = new Set(['Shift', 'Control', 'Alt', 'Meta', 'CapsLock', 'NumLock']);
```
TGlide, line 34: I think it's better to have a allowList rather than a ignoreList, no? Even better, we can just regex and check if the key is a letter.

huntabyte: It can also be symbols though right?

TGlide: Yeah, playing around with native select, you're right.

TGlide (approved):

Issue #808, opened 2023-12-08
🐛Bug: Scroll not restored when listbox destroyed


huntabyte:
**Describe the bug**
When you have a listbox inside a conditionally rendered component like popover content with a transition, if you open the select and then outside click the popover, the scroll is never restored because the `destroy` action is never called.

**To Reproduce**
https://github.com/huntabyte/melt-select-repro

Open the popover, then open the select, then click outside the popover. You'll see the scroll is never restored, and the little dummy actions `destroy` is never called, which is where the scroll is readded for the listbox components.

**Expected behavior**
The scroll should be readded when the select's `open` is false.

https://github.com/melt-ui/melt-ui/assets/64506580/a1a63be2-b701-4dd9-a022-64e7094cdedc

Issue #733, opened 2023-11-14
🐛Bug: Listbox - when going from one listbox to another, scrollbar becomes visible


huntabyte:
**Describe the bug**

https://github.com/huntabyte/shadcn-svelte/assets/44144724/0d861df8-a260-4a3b-8b89-068dd77820f7

huntabyte:
Any ideas for how we may tackle this @TGlide? I am more than happy to investigate/attempt to implement.

A wild idea I have but have yet to explore before hearing your thoughts is tracking the latest component actively removing the scroll, likely based on some internal `id` that represents the entire builder/component.

We could then update some dummy element to have that ID as part of a `data-attr,` when we close a piece, we check if that dummy element's ID matches the `scrollLockId` of the current builder, and if so, we re-add the scroll. Otherwise, we don't.

TGlide:
I wonder if this was always the case? Or if it was implemented with the transition support. Tough nut to crack... I'll have to play around. I have loads of issues to address and want to work through them, so we no longer have 72 issues 😅 then I can play with this too. Ofc feel free to play around yourself too!

huntabyte:
Actually realizing it now this is only happening with Listbox components.

PR #809, opened 2023-12-08
fix: Listbox `removeScroll` bug

huntabyte:
Closes: #808 #733

### DIFF
--- a/src/lib/builders/listbox/create.ts
+++ b/src/lib/builders/listbox/create.ts
@@ -456,22 +456,12 @@ export function createListbox<
 		},
 		action: (node: HTMLElement): MeltActionReturn<ListboxEvents['menu']> => {
 			let unsubPopper = noop;
-			let unsubScroll = noop;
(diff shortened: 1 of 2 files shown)


src/lib/builders/listbox/create.ts
```diff
@@ -661,6 +646,20 @@ export function createListbox<
...
+		let unsubScroll = noop;
+
+		if (get(preventScroll) && $open) {
+			unsubScroll = removeScroll();
+		}
+
+		return () => {
+			unsubScroll();
+		};
+	});
```
TGlide, line 661:
```suggestion
	effect([open], ([$open]) => {
		if (!isBrowser || !get(preventScroll)) return;
		
		if ($open) {
			return removeScroll();
		}
	});
```

huntabyte: Damn, you make my code look 🤮 😂

PR #811, opened 2023-12-08
fix: PIN Input - fix `delete` key event handling and add tests

ollema:
I think this is what is causing my test to fail in:
https://github.com/huntabyte/bits-ui/pull/227

edit: yep, seems like this works!

also added tests to test this and test other PIN Inputs features


TGlide (approved):
Awesome, thanks a ton!

PR #825, opened 2023-12-17
fix(slider): remove `aria-disabled` from root

abdel-17:
The `aria-disabled` prop should be only be present on the thumb, where the "slider" role exist.


TGlide (approved):
Thank you!

Issue #828, opened 2023-12-18
📁Docs: Tags Input mentions that `addOnPaste` is `true` by default but it doesn't work


mcmxcdev:
When searching on https://melt-ui.com/docs/builders/tags-input for `addOnPaste` it mentions that the default value is `true`. Trying to paste anything into the component input doesn't tagify the content though.

I assume this is a docs only issue rather than a bug because it seems to work fine when adding `addOnPaste: true` manually to `createTagsInput` locally.

PR #833, opened 2023-12-19
docs: Tags Input mentions that addOnPaste is true by default but it doesn't work

mcmxcdev:
## Changes
- change wrong default of `true` to `false` for `addOnPaste` in `tags-input.ts`

## Motivation

The default value of `addOnPaste` is defined as `false` here: https://github.com/melt-ui/melt-ui/blob/26e41134169528e22451782259be0b2506bdaec0/src/lib/builders/tags-input/create.ts#L34

BUT it was showing as `true` in the documentation on the website.

Closes https://github.com/melt-ui/melt-ui/issues/828


TGlide:
Thank you!

PR #841, opened 2023-12-20
docs: Update PP installation guide

AdrianGonz97:
A custom version of `sequence` (to fix source maps) is now built into the `@melt-ui/pp` package, so we don't need `svelte-sequential-preprocessor` anymore.


TGlide:
You're the GOAT!

PR #845, opened 2023-12-21
fix(escape-keydown): remove `capture` listener

### LINKED ISSUE #842: 🐛Bug: `actions/escape-keydown` does not ignore prevented/stopped events
**Describe the bug**

If I have my own `escape` keydown listener inside a component, a parent Melt/Bits component that attaches a `escape-keydown` listener will execute, despite my own component's attempts to stop bubbling.

For example, from Bits:

```html
<Sheet.Root bind:open>
  <Sheet.Contentclass="sm:max-w-3xl overflow-y-scroll overscroll-y-contain">
  
    <CustomInput />
  
  </Sheet.Content>
</Sheet.Root>
```

When `CustomInput` has a keydown listener that looks like this:

```js
function onkeydown(ev: KeyboardEvent) {
  let key = ev.which;

  // ESCAPE
  if (key === 27) {
    ev.preventDefault();
    ev.stopPropagation();
    ev.stopImmediatePropagation();
    open = false;
    return;
  }

  // ...
}
```

... then the parent Sheet should not also close from the <kbd>ESC</kbd> keypress.

lukeed:
> A boolean value indicating that events of this type will be dispatched to the registered listener before being dispatched to any EventTarget beneath it in the DOM tree. If not specified, defaults to false.

This means that the top-level listener was running before any listener(s) that actually should be run. Because of this, it was impossible for a "child" listener to intercept an `ESC` keydown event & have it not affect the rest of the page.

Closes #842

### DIFF
--- a/.changeset/slow-apples-fry.md
+++ b/.changeset/slow-apples-fry.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+escape-keydown: remove `capture` listener (fixes #842)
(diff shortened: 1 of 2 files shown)


TGlide:
Thank you! 🎉

PR #846, opened 2023-12-21
DateField/DateRangeField: add `readonlySegments` prop

ksmoore17:
Given segments can be set as readonly to control them only through the value store


ksmoore17:
I just force pushed to change my mentions of `SegmentPart` in the docs into `EditableSegmentPart`

ksmoore17:
Any idea how to solve that eslint error in the css previews? I know I could use classes instead of the `[data-melt-datefield-*]` selectors, but it's the same error then with the `[data-invalid]` selectors. I see that the pagination css example  [uses `:global`](https://github.com/melt-ui/melt-ui/blob/013a3b7dddaf9a66d69c1e0086e8e1d22ee485fd/src/docs/previews/pagination/main/css/index.svelte#L68) to get around the same issue. Should I just remove the css examples?

Interestingly, the file has no lint errors for any selector (even a truly unused class) if I don't use the preprocessor for just one of the elements:
```
<div>
	<span {...$label} use:label class="label">Appointment Date</span>
	<div use:melt={$field} class="field">
		{#each $segmentContents as seg, i (i)}
			<div use:melt={$segment(seg.part)}>
				{seg.value}
			</div>
		{/each}
	</div>
</div>

<style>
	[data-melt-datefield-label] {
		...
	}
	[data-melt-datefield-label][data-invalid] {
		...
	}
	[data-melt-datefield-field] {
		...
	}
	[data-melt-datefield-field][data-invalid] {
		...
	}
	...
</style>
```

huntabyte:
Try adding `lang="postcss"` to the `<style>` tag.

huntabyte:
We don't want to expose the `readonlySegments` props to the `DatePicker` & `DateRangePicker` components. It doesn't make much sense there considering they could be overwritten by the calendar, so I've omitted them from those exposed props for those builders.

huntabyte (approved):
Thanks a ton for this, the tests are greatly appreciated!

Issue #830, opened 2023-12-18
🐛Bug: Tags Input should clear input after pasting value into it 


mcmxcdev:
**Describe the bug**
When using Tags Input with `addOnPaste: true` and copying a value, that value is tagified but also left behind in the input. This definitely isn't expected behavior because by default unique values are not allowed in the Tags Input.

**Expected behavior**
Tags Input should be cleared after pasting value into it .

**Screenshots**

[Screencast from 2023-12-18 09-58-32.webm](https://github.com/melt-ui/melt-ui/assets/16797721/6874451b-8711-4933-b218-9f94eb06ee84)

**Additional context**
https://stackblitz.com/edit/melt-ui-playground-hujcj8?file=package.json,src%2Flib%2Fcomponents%2FAccordion.svelte

PR #847, opened 2023-12-22
Fix/bug tags input should clear

SikandarJODD:
Solved Issue #830  
Describe the bug 
When using Tags Input with addOnPaste: true and copying a value, that value is tagified but also left behind in the input. This definitely isn't expected behavior because by default unique values are not allowed in the Tags Input.

Expected behavior
Tags Input should be cleared after pasting value into it .

Solved by adding 
```
 e.preventDefault():
 ```
 on Line no : 328 at : <code>src/lib/builders/tags-input/create.ts</code>
 
 Now it works Properly : 

https://github.com/melt-ui/melt-ui/assets/93428946/ece1ab08-bc6a-4b3c-bc28-a6f1f1a35f5a

### CODE REVIEW

File: src/lib/builders/tags-input/create.ts [hunk 1]
```diff
@@ -325,6 +325,7 @@ export function createTagsInput(props?: CreateTagsInputProps) {
 				}),
 				addMeltEventListener(node, 'paste', async (e) => {
 					// Do nothing when there is nothing to paste
+					e.preventDefault();
```
TGlide, line 328: We don't want to prevent default if `addOnPaste` is set to `false`

TGlide (changes requested):
Thank you for the contribution!

Left a comment. Also, please remove the `package-lock.json`, as this project uses `pnpm`, not `npm`.

src/lib/builders/tags-input/create.ts
SikandarJODD: yes i changed it to : on Line No 328 
```
     if (addOnPaste) {
	e.preventDefault();
     }
```

TGlide: That will always be true, as that is the store, not the value of the store. We're also already checking for the value a few lines below.

SikandarJODD: Can i do this instead 
```
                     if (!get(addOnPaste)) return;
                     else e.preventDefault();
```

PR #849, opened 2023-12-22
Slider: Refactor Thumb and Tick

abdel-17:
This pr changes and `thumb` and `tick` elements from being functions to just arrays of builders.

```html
<!-- the old API -->
{#each { length: $ticks } as _}
  <span use:melt={$tick()} />
{/each}

{#each $value as _}
  <span use:melt={$thumb()} />
{/each}

<!-- the new API -->
{#each $thumbs as thumb}
  <span use:melt={thumb} />
{/each}

{#each $ticks as tick}
  <span use:melt={tick} />
{/each}
```

Limitations:
The melt pp does not seem to work with the new loop syntax. I added TODOs to replace the loops later on when this issue is fixed.

Update:
The pp works fine. The problem is in the docs itself.

### DIFF
--- a/.changeset/lazy-zoos-itch.md
+++ b/.changeset/lazy-zoos-itch.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": minor
+---
+
+Slider: Refactor thumb and tick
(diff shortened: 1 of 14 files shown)


abdel-17:
Fixed the docs' pp issue 👍🏼

PR #850, opened 2023-12-23
Right arrow in tree view focuses next element if not expanding

LoganDark:
If you have an element without children, pressing right arrow focuses the next item.

This behavior is consistent with IDEs and allows expanding an entire directory tree simply by holding right-arrow.

### DIFF
--- a/.changeset/loud-radios-hug.md
+++ b/.changeset/loud-radios-hug.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+Tree: Right arrow focuses next element if not expanding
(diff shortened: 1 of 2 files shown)


TGlide:
(writing down what I wrote on Discord)

Thanks for the PR! But again I worry about accessibility. The [APG pattern for trees](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/) describes the interaction when using the right arrow key, and specifies that when the focus is on an end node (one that does not have child nodes), the right arrow key does nothing.

In my experience, that's how I've seen it work too. I've sent a video where I hold right arrow on a few folders through VSCode.

https://github.com/melt-ui/melt-ui/assets/26071571/f3680567-9faa-4d5d-9786-dc0a11b4a881

In any case, behaviour like this is easily overrideable in melt, be it by overriding the keydown handler, or just providing an extra one on top.

LoganDark:
Closing this since it seems I don't have a compelling enough reason to disobey the spec, which doesn't mention this as an optional behavior

PR #859, opened 2023-12-27
fix(Toast): use as const for `'assertive' | 'polite'` aria-live type

ksmoore17:
aria-live property on toast `content` store was returning as `string` where ts expects `"off" | "assertive" | "polite"`. Similar to #839.

fixes ex:
image

### DIFF
--- a/src/lib/builders/toast/create.ts
+++ b/src/lib/builders/toast/create.ts
@@ -111,7 +111,7 @@ export function createToaster<T = object>(props?: CreateToasterProps) {
 					role: 'alert',
 					'aria-describedby': toast.ids.description,
 					'aria-labelledby': toast.ids.title,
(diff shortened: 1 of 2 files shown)


src/lib/builders/toast/create.ts
```diff
@@ -111,7 +111,7 @@ export function createToaster<T = object>(props?: CreateToasterProps) {
 					role: 'alert',
 					'aria-describedby': toast.ids.description,
 					'aria-labelledby': toast.ids.title,
-					'aria-live': toast.type === 'foreground' ? 'assertive' : 'polite',
+					'aria-live': toast.type === 'foreground' ? ('assertive' as const) : ('polite' as const),
```
TGlide, line 114:
```suggestion
					'aria-live': (toast.type === 'foreground' ? 'assertive' : 'polite') as const,
```

TGlide (approved):
Thank you!

PR #860, opened 2023-12-28
[Table of Contents] fix smooth scroll and scroll offset

MadeBaruna:
Setting location hash will trigger browser navigation jump


TGlide:
What issue does this solve?

MadeBaruna:
On the table of content, if you click the link, it will jump directly to the heading, ignoring the smooth scroll and the scroll offset

TGlide:
This brings a regression, as even when setting scroll-behavior to auto, the scroll is now smooth.

TGlide:
Thank you! 🎉

PR #861, opened 2023-12-28
Added a Calendar builder prop to output months as an array of strings for multiple months.

lolcabanon:
I added a `createCalendar()` prop : `monthsHeaderType: 'string' || 'array'`.

Useful to display calendar header with different elements over each month instead of just a month range string.

Example :

```svelte
<div use:melt={$calendar}>
	<header>
		<button use:melt={$prevButton}>
			<ChevronLeft size=".7em" />
		</button>

               <!-- THIS HERE -->
		{#if $headingValue instanceof Array}
			<div use:melt={$heading} title={$headingValue.join(' - ')}>
				{#each $headingValue as h}
					<span>{h}</span>
				{/each}
			</div>
		{:else}
			<div use:melt={$heading} title={$headingValue}>
				<span>{$headingValue}</span>
			</div>
		{/if}


		<button use:melt={$nextButton}>
			<ChevronRight size=".7em" />
		</button>
	</header>
<!-- ... -->
</div>

<style lang='scss'>
[data-melt-calendar-heading] {
    display: flex;
    gap: 1em;
    width: 100%;

    span {
	flex: 1;
	text-align: center;
    }
}
</style>
```

Alternatively, it could be a boolean like `displayAllMonths: boolean = false` or another version that would always return `string` instead of `string[]`, something like `multipleMonthsHeader: 'range' || 'all' = 'range'`. I'm open to design suggestion to optimize compatibility and simplicity.

For the moment, I do `{#each $headingValue.split(' - ') as h}` to simulate the same behaviour, but it works only with two months.

~~**Note :**~~
~~I cannot figure out why, it seems to work in prerendering, but not in the frontend with `vite dev` server.~~

~~When I add a `console.log($headingValue, $monthsHeaderType)` on the top level of my component, it is logged like this `[ 'décembre 2023', 'janvier 2024', 'février 2024' ] array` in my terminal, but like this `décembre 2023 - février 2024 array` in the browser.~~

~~Maybe something is cached in my pc... I deleted my `.sveltekit` folder, hard refreshed the browser, disabled cache, but nothing changed.~~


lolcabanon:
To illustrate my use case :

This is not so intuitive :

image

This is much more readable to me :

image

lolcabanon:
> **Note :** I cannot figure out why, it seems to work in prerendering, but not in the frontend with `vite dev` server.

Ok, I just figured it out. I did not `pnpm link --global` correctly, so I was keeping the old version. Tested and it works like expected.

Now I am waiting for your feedback to see if you have an interest in this feature and how to implement it!

Happy new year! :partying_face:

huntabyte:
Hey @lolcabanon, I completely understand your demonstrated use-case, but could this not be accomplished outside of Melt, given you have access to the `months` array?

The entire goal of the builder patterns and Melt UI in general is to give you the tools you need to build your components however you see fit. While we do provide some "starter" helpers, like the `headingValue` store to get your component off the ground, we don't want to end up in a state of "prop hell" where every tiny alteration for every possible use case is exposed as a prop. We're already slipping into that trap in other builders where every individual use-case is added as a prop, when it would take the user 15 seconds to implement it outside of Melt.

Although this may seem like a small addition, over time and with a bunch of these, we end up with a massive maintenance burden and increases the API surface area which once we're 1.0 will have to remain stable. This makes it more difficult to release new features in minor releases as we have a lot of "little" things to step around to avoid introducing a major breaking change.

lolcabanon:
Hey, totally makes sense. I'll keep that in mind if I have other contribution propositions to melt!

Thanks for your review and your awesome work on MeltUI and in the Svelte community. :)

PR #870, opened 2024-01-01
fix(progress): allow null value for indeterminate state

tmarnet:
The purpose of this PR is to update the types of the `value` prop used by the `Progress` builder to allow for passing a `null` value to effectively be able to set the progress in the `indeterminate` state, as the implementation suggests:

https://github.com/melt-ui/melt-ui/blob/74e868522ec869ed8a1cdb2192f41a9bd881bd8e/src/lib/builders/progress/create.ts#L38


TGlide (approved):
Thank you!

PR #871, opened 2024-01-02
fix(menu): focus on the trigger when rootActiveTrigger is defined

### LINKED ISSUE #867: 🐛Bug: Dropdown Menu - Focus on trigger button after page load
**Describe the bug**
Immediately after the page loads, it automatically focuses on the element that opens and closes the Dropdown menu.

**To Reproduce**
1. Go to https://melt-ui.com/docs/builders/dropdown-menu

**Expected behavior**
It should not be in focus immediately after the page loads

**Screenshots**

 - OS: [macOS Sonoma 14.2.1]
 - Browser [chrome]
 - Version [120.0.6099.129]

**Additional context**

To reproduce using SvelteLab, use the [Melt UI template](https://www.sveltelab.dev/?t=melt_ui) and modify the `src/routes/+page.svelte` file.

azukiazusa1:
fix #867.

When the page is first loaded, there is no need to focus on the trigger even if $rootOpen is false.

However, when rootActiveTrigger is set at the time it is mounted, the effect will be executed and the trigger will be focused.

Since we only need to pay attention to whether the menu is closed or not, we did not need to monitor the rootActiveTrigger.

### DIFF
--- a/.changeset/seven-pillows-turn.md
+++ b/.changeset/seven-pillows-turn.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+fix(menu): focus on the trigger when rootActiveTrigger is defined
--- a/src/lib/builders/menu/create.ts
+++ b/src/lib/builders/menu/create.ts
@@ -1138,55 +1138,50 @@ export function createMenuBuilder(opts: _MenuBuilderOptions) {
 		}
 	});
 
-	effect(
-		[rootOpen, rootActiveTrigger, preventScroll],
-		([$rootOpen, $rootActiveTrigger, $preventScroll]) => {
-			if (!isBrowser) return;
-
-			const unsubs: Array<() => void> = [];
-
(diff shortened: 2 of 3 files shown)


huntabyte:
> Since we only need to pay attention to whether the menu is closed or not, we did not need to monitor the rootActiveTrigger.

This isn't true if using `forceVisible` and conditionally rendering the menu, as you can see from the test I just added that is failing.

Instead, I think we should use a `closeMenu` function which would handle the focus restoration once the menu has closed, rather than focus being handled within the `effect`.

huntabyte:
Let me confirm that this won't impact the auto refocus of the trigger when a user closes the menu using the keyboard. I added a test which is currently failing for it. There's probably a better way to handle the focus restoration rather than simply reacting to the `open`. Will do some explorations cause always focusing the button is def not the intended behavior.

huntabyte (approved):
Thank you!

azukiazusa1:
Thanks for the review!

PR #877, opened 2024-01-04
chore: update issue template

huntabyte:
Improve the issue template. Follows a similar format as https://github.com/sveltejs/svelte which gives us a good amount of information to debug (versions, reproduction, etc.) so we aren't wasting energy running down non-issue issues 🫡


TGlide (approved):
Awesome!

PR #879, opened 2024-01-05
chore: Revert `vite-plugin-svelte` upgrade

AdrianGonz97:


huntabyte (approved):
We aren't using sveltekit v2 yet, so we shouldn't be using this plugin! Nice catch!

Issue #865, opened 2024-01-01
🐛Bug: Dialog Overlay Issue on Mobile like-devices


FortyGazelle700:
**Describe the bug**
When you are on a mobile device, or using the Device Emulation Tools to simulate a mobile device, you can click past the dialog overlay onto the elements below.

**To Reproduce**
1. Open [MeltUI - Dialog](https://melt-ui.com/docs/builders/dialog)
2. Open DevTools (`F12`)
3. Use Device Emulation Tool (`Ctrl`+`Shift`+`M`)
4. In Dimensions Dropdown (at the top of your window, just below address bar) select: `iPhone Se` (or similar)
5. Click on `New API Key`
6. While the dialog is open, click in the top left corner, where you'd expect the MeltUI's hamburger icon to be

**Expected behavior**
Just the dialog element to close

**Desktop (please complete the following information):**
 - OS: Using Manjaro (Arch Linux) to emulate
 - Browser: Edge
 - Version: 119.0.2151.58

**Smartphone (please complete the following information):**
 - Device: Using S22 Ultra (Samsung) to emulate
 - OS: Android
 - Browser: Chrome
 - Version: 120.0.6099.144

Issue #832, opened 2023-12-18
🐛Bug: Dialog component overlay does not block touch events for elements beneath it


luisfontes:
**Describe the bug**
The dialog component overlay does not block touch events for elements beneath it.

**To Reproduce**
1. Go to [https://melt-ui.com/docs/builders/dialog](https://melt-ui.com/docs/builders/dialog)
2. Toggle device emulation on Chrome/Edge
3. Click on the "Open Dialog" button
4. Click on any bottom nav buttons under the overlay
5. Dialog closes and right after the action of the nav button is fired

**Expected behavior**
No event fired for anything beneath the dialog overlay.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Edge
 - Version: 120.0.2210.77

luisfontes:
This seems to have been fixed in one of the recent releases. Closing the issue.

huntabyte:
I'm still able to reproduce this!

huntabyte:
If you go to https://melt-ui.com/docs/builders/dialog (on a mobile device), open the dialog, and then click on the little toggle button behind the dialog, you'll see the event passes through to the toggle.

luisfontes:
You're right, I forgot to toggle device emulation 🤦

PR #883, opened 2024-01-05
fix: improve outside interactions (click outside)

huntabyte:
Closes: #865 #750 #832

This is the initial introduction to improving interactions within Melt. This closes a few different outstanding bugs, so I think we should review/release this bit here to take care of those.

In the near future, I'd like to work with you, @TGlide to see how we can better encapsulate the "modal" behavior across the various builders that use it. We should be able to incorporate all of the things that modals handle:
- closing on outside interactions/clicks
- preventing scroll
- handling escape keydown events
- focus traps
- etc...

Essentially, the goal is to make this a single action that can handle all the above cases, with their respective overrides, handler functions, conditions, etc. so we can fix something in one place and it becomes a fix across all builders. We can then expose simple/consistent APIs to builder consumers without having to put too much thought into it, and behavior is familiar/consistent across every builder.


src/lib/builders/dialog/create.ts
```diff
@@ -228,17 +206,16 @@ export function createDialog(props?: CreateDialogProps) {
 				}),
 
 				effect([closeOnOutsideClick, open], ([$closeOnOutsideClick, $open]) => {
-					return useClickOutside(node, {
-						enabled: $open,
-						handler: (e: PointerEvent) => {
+					return useModal(node, {
```
TGlide, line 209: Could we maybe have it receive the stores directly? That way we don't need to do an effect here.

src/lib/internal/actions/modal/action.ts
```diff
@@ -0,0 +1,73 @@
...
+		const index = visibleModals.indexOf(node);
+		if (index >= 0) {
+			visibleModals.splice(index, 1);
+		}
+	}
+
+	function update(config: ModalConfig) {
+		const { open, onClose, shouldCloseOnInteractOutside, closeOnInteractOutside } = config;
+
+		sleep(100).then(() => {
```
TGlide, line 21: Doesn't tick or sleep(1) suffice here?

huntabyte: Nice catch, let me take another look to ensure the tests still pass without this. I know it was added during debugging the nested dialogs, so I may have let it sneak in by accident!

huntabyte: Actually, I left it as it was originally in your implementation of handling the nested dialogs:

CleanShot 2024-01-11 at 10 27 02@2x

We can see if the tests still pass with it being 1 or less than 100 though!

src/lib/builders/dialog/create.ts
huntabyte: So you're thinking we pass in the stores and just move the `effect`s into the `useModal`? I think that's a good idea and would clean this up quite a bit.

src/lib/internal/actions/modal/action.ts
huntabyte: Okay I removed this and it passed all the tests 🫡

src/lib/builders/dialog/create.ts
huntabyte: Consolidated the effect into the action

src/lib/internal/actions/interact-outside/action.ts
```diff
@@ -0,0 +1,96 @@
...
+import type { InteractOutsideConfig, InteractOutsideEvent } from './types.js';
+
+export function useInteractOutside(node: HTMLElement, config: InteractOutsideConfig) {
+	let unsub = noop;
+	function update(config: InteractOutsideConfig) {
+		unsub();
+		const { onInteractOutside, onInteractOutsideStart, enabled } = config;
+
+		if (!enabled) {
+			return {
```
TGlide, line 11: Why are you returning destroy here? You should make unsub = noop, and then do an early return, no?

src/lib/internal/actions/interact-outside/types.ts
```diff
@@ -0,0 +1,7 @@
+export type InteractOutsideEvent = PointerEvent | MouseEvent | TouchEvent;
+
+export type InteractOutsideConfig = {
+	onInteractOutside?: (e: InteractOutsideEvent) => void;
```
TGlide, line 4: If the other one is start, what is this one? End? Maybe some JSDoc here would help clear things up

babakfp:
It seems like #832 and #865 are not fixed:

https://github.com/melt-ui/melt-ui/assets/44144724/9dc9ff52-b334-4b7c-99bc-11474ffc3980

I tested it in https://huntabyte-fix-outside-intera.melt-ui.pages.dev/docs/builders/dialog

I'm not sure if this PR is sposed to fix the issue in the above URL or not.

huntabyte:
Putting this back into draft while I figure out where I introduced the regression because it should be preventing the behavior being displayed above.

huntabyte:
@TGlide I'm having some issues with moving the effect into the action itself, seems to have something to do with cleanup as the events aren't unregistering, so I reverted it back to the effect within the `create` dialog.

@babakfp, the browser's mobile emulation won't cut it here as we're detecting the possibility of `PointerEvent`s, which your PC's browser still has. If you clone the repo serve it via ngrok and access it via your phone, you'll see that it does in fact close those issues.

babakfp:
> @babakfp, the browser's mobile emulation won't cut it here as we're detecting the possibility of `PointerEvent`s, which your PC's browser still has. If you clone the repo serve it via ngrok and access it via your phone, you'll see that it does in fact close those issues.

Testing it with my phone:

https://github.com/melt-ui/melt-ui/assets/44144724/21006703-380a-43cc-895c-688949230cdb

huntabyte:
Same for this? https://a71888c7.melt-ui.pages.dev/

babakfp:
> Same for this? https://a71888c7.melt-ui.pages.dev/

Yep:


https://github.com/melt-ui/melt-ui/assets/44144724/b68ab5b1-76f5-420c-9843-850aca1f95ee

huntabyte:
Browser and OS? @babakfp

babakfp:
> Browser and OS? @babakfp

Chrome. Android.

IMG_20240119_054226
IMG_20240119_054205
IMG_20240119_054141

huntabyte:
Latest preview I can't reproduce

https://github.com/melt-ui/melt-ui/assets/64506580/94880b72-f8b1-4807-93f8-9904ecb32715

Issue #831, opened 2023-12-18
🌟Feature request: Combobox to have grouped options


pricetula:
**Is your feature request related to a problem? Please describe.**
Would have been great to have grouped options on the combobox component

**Describe the solution you'd like**
Replicate builder store used to create a select group.

**Describe alternatives you've considered**
N/A

**Additional context**
Similar implementation of grouped select options
[
<img width="326" alt="Screenshot 2023-12-19 at 00 50 37" src="https://github.com/melt-ui/melt-ui/assets/3857337/d675e588-8676-4a36-b91a-ab65d8b9106e">
](https://melt-ui.com/docs/builders/select#group)

PR #891, opened 2024-01-09
Combobox: add group and groupLabel elements

ksmoore17:
Closes: #831 

creates group and group label in listbox for both select and combobox builders. Also adds some missing entries in the select and combobox api reference


huntabyte (approved):
Thanks for tackling this! Looks great!

PR #893, opened 2024-01-10
allow 00 as values in minute and second segment for date field

sakibulalam:
Previously when User typed in `00` as the value for minute and second segment it would not get registered. This fixes the issue and allows `00` as a valid input for minute and second segment but disallowing for the other fields.

### DIFF
--- a/.changeset/warm-donkeys-eat.md
+++ b/.changeset/warm-donkeys-eat.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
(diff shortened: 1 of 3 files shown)


huntabyte (approved):
Awesome job, thanks for this!

luhagel:
Heya, this issue is currently affecting us, would it be possible to get this merged into a release in the next few days? :)

sakibulalam:
@huntabyte @TGlide can we merge this?

PR #897, opened 2024-01-11
allow changing dayPeriod regardless of state of other segments

sakibulalam:
it seems `dayPeriod` was hardcoded to be a `defaultParts` and the updated value was not reflected. Also, there was a bug in the `hour` update when `dayPeriod` is toggled.


huntabyte:
This is failing CI locally for me now, so not sure how it's passing here.

Ideally we shouldn't update the hour segment content when the dayPeriod is changed unless an hour has been set already. I can try to look at this in more detail later!

sakibulalam:
> Ideally we shouldn't update the hour segment content when the dayPeriod is changed unless an hour has been set already. I can try to look at this in more detail later!

That makes sense. Previous implementation set the hour segment when it's not set. I will push changes to keep the hour segment as is if it's not set.

This cases should suffice
- If `defaultValue` is not set, and `dayPeriod` is updated, `hour` segment remains empty
- If `defaultValue` is set, and `dayPeriod` is updated,  true `hour` is updated to match the `dayPeriod`
- If `hour` segment is set, and `dayPeriod` is updated,  true `hour` is updated to match the `dayPeriod`

TGlide:
Closing due to inactivity

Issue #573, opened 2023-10-01
✨ Builder: Scroll Area


TGlide:

huntabyte:
My best wishes to you on this one buddy!

huntabyte:
> My best wishes to you on this one buddy!

Oh how the turns have tabled 😂

shyakadavis:
![turntables](https://github.com/melt-ui/melt-ui/assets/87414827/eb750b1b-0f27-406b-a4b3-6712681f9445)

PR #908, opened 2024-01-13
New Builder: Scroll Area

huntabyte:
Closes: #573


anatolzak:
Hey! Thanks for working on this! 

Just wanted to mention in case it wasn't noticed that for the scroll area of type `scroll`, the scroll thumb jumps when you start scrolling slowly.

https://github.com/melt-ui/melt-ui/assets/53095479/e263c0e7-524a-4be3-b137-52f1cf8f5054

It is more difficult to notice this behavior when scrolling a little faster

https://github.com/melt-ui/melt-ui/assets/53095479/156bdad3-b6de-429b-a162-2068745c2f96

Also, if you press the space bar while the scroll bar is not showing, it messes up its position when it is visible again

https://github.com/melt-ui/melt-ui/assets/53095479/3bbd50b4-21ee-4dd0-8bbd-46da8ec113b7

TGlide:
@anatolzak thanks for pointing it out! Fixed :)

anatolzak:
@TGlide when using the scroll area of type `scroll`, if you hover over the scroll thumb while it is visible, it shouldn't disappear. Currently, it disappears if you hover over the scroll thumb, and it only becomes visible when you start scrolling again.

With Radix UI Scroll Area of type `scroll`, the scroll thumb doesn't disappear if you hover over it while it is visible. And this is the same with the macOS scroll bar.

anatolzak:
Also, there is an issue on Firefox with the scroll area of type `hover`. When hovering over the scroll area, leaving the tab/window using the keyboard without moving the cursor, and coming back to the tab, the scroll thumb is not shown even though the cursor is over the scroll area.

I tried attaching the `onmouseover` event, and it seems like that would solve this issue. 

https://github.com/melt-ui/melt-ui/assets/53095479/60236923-8959-49f3-89ff-c4110104cbf7

huntabyte:
Alright @anatolzak - once this latest preview deploys check it out and see if those issues are resolved!

shyakadavis:
Hi;

Firefox anarchist here confirming for both cases 😅 


https://github.com/melt-ui/melt-ui/assets/87414827/84c25b38-7aee-4c0d-affe-a9647244c20e


https://github.com/melt-ui/melt-ui/assets/87414827/2229c445-9599-44fe-8da6-454f9fcd94f0

anatolzak:
@huntabyte I can confirm the issues I described earlier do not happen anymore. 

One new issue I found is with the `hover` type. If the cursor is already over the scroll area before the JS loads, the scroll thumb doesn't show. I assume this can't be solved as we add event listeners after the JS loads, and if the user doesn't move the cursor after the event listeners are added, there is no way to know that the cursor is over the scroll area. 

And Radix appears to have the same issue.


https://github.com/melt-ui/melt-ui/assets/53095479/5d66cac9-fc12-45ae-8b63-12b8673d2914



However, in firefox, if your cursor was already over the scroll area before the event listeners are added, and then you start moving your cursor, the scroll thumb still doesn't show. This is because the `mouseenter` never fires on firefox. A fix for this is to either listen to `mousemove` or `pointerenter` for this issue to be resolved in firefox.

https://github.com/melt-ui/melt-ui/assets/53095479/cae493bd-1038-43f4-a547-b51009aa3d32

huntabyte:
Awesome! Thank you @anatolzak! Yeah the pre-JS issue is unavoidable I think, unless there is a way we can get a reference to the users pointer position on load (no clue if possible). 

@TGlide all the glaring issues are cleaned up if you have time to give her a peek!

anatolzak:
@huntabyte I also fixed a padding issue in the docs that prevents scroll when the pointer is over the root container padding

https://github.com/huntabyte/melt-ui/pull/2

huntabyte:
> @huntabyte I also fixed a padding issue in the docs that prevents scroll when the pointer is over the root container padding

You're awesome, thank you @anatolzak!

src/lib/builders/scroll-area/create.ts
```diff
@@ -0,0 +1,524 @@
...
+				}),
+				id: $rootId,
+			};
+		},
+		action: (node: HTMLElement) => {
+			scrollAreaEl.set(node);
+			return {
+				destroy() {
+					scrollAreaEl.set(null);
+				},
```
TGlide, line 142: Maybe an update checking if the currEl is the root being destroyed would be better

src/lib/builders/scroll-area/create.ts
```diff
@@ -0,0 +1,524 @@
...
+			};
+		},
+		action: (node: HTMLElement): MeltActionReturn<ScrollAreaEvents['thumb']> => {
+			scrollbarState.thumbEl.set(node);
+
+			let effectHasRan = 0;
+			let unsubViewportScroll = noop;
+
+			effect([scrollbarState.sizes], ([_]) => {
+				if (effectHasRan === 2) return;
```
TGlide, line 430: Why do we need this?

src/lib/builders/scroll-area/create.ts
```diff
@@ -0,0 +1,524 @@
...
+				const $viewportEl = rootState.viewportEl.get();
+				if ($viewportEl) {
```
TGlide, line 432: Use an early return here

src/lib/builders/scroll-area/helpers.ts
```diff
@@ -0,0 +1,125 @@
...
+	const thumbOffsetFromEnd = thumbSizePx - offset;
+	const minPointerPos = sizes.scrollbar.paddingStart + offset;
+	const maxPointerPos = sizes.scrollbar.size - sizes.scrollbar.paddingEnd - thumbOffsetFromEnd;
+	const maxScrollPos = sizes.content - sizes.viewport;
+	const scrollRange = dir === 'ltr' ? [0, maxScrollPos] : [maxScrollPos * -1, 0];
+	const interpolate = linearScale([minPointerPos, maxPointerPos], scrollRange as [number, number]);
+	return interpolate(pointerPos);
+}
+
+function clamp(value: number, [min, max]: [number, number]): number {
```
TGlide, line 107: Don't we already have this util elsewhere?

src/lib/builders/scroll-area/scrollbars.ts
```diff
@@ -0,0 +1,425 @@
...
+ * The other scrollbar actions will extend this one, preventing a ton
+ * of code duplication.
+ */
+export function createBaseScrollbarAction(state: ScrollAreaState) {
+	const { rootState, scrollbarState } = state;
+	scrollbarState.isVisible.set(true);
+
+	function handleDragScroll(e: MouseEvent) {
+		const $domRect = scrollbarState.domRect.get();
+		if ($domRect) {
```
TGlide, line 38: Use an early return here

src/lib/builders/scroll-area/scrollbars.ts
```diff
@@ -0,0 +1,425 @@
...
+		if (!isHTMLElement(currentTarget)) return;
+		scrollbarState.domRect.set(currentTarget.getBoundingClientRect());
+		scrollbarState.prevWebkitUserSelect.set(document.body.style.webkitUserSelect);
+		document.body.style.webkitUserSelect = 'none';
+		const $viewportEl = rootState.viewportEl.get();
+		if ($viewportEl) {
+			$viewportEl.style.scrollBehavior = 'auto';
+		}
+		handleDragScroll(e);
+	}
```
TGlide, line 65:
```suggestion
	function handlePointerDown(e: PointerEvent) {
		if (e.button !== 0) return;
		
		const target = e.target;
		if (!isHTMLElement(target)) return;
		target.setPointerCapture(e.pointerId);
		
		const currentTarget = e.currentTarget;
		if (!isHTMLElement(currentTarget)) return;
		scrollbarState.domRect.set(currentTarget.getBoundingClientRect());
		scrollbarState.prevWebkitUserSelect.set(document.body.style.webkitUserSelect);
		document.body.style.webkitUserSelect = 'none';
		
		const $viewportEl = rootState.viewportEl.get();
		if ($viewportEl) {
			$viewportEl.style.scrollBehavior = 'auto';
		}
		
		handleDragScroll(e);
	}
```

src/lib/builders/scroll-area/scrollbars.ts
```diff
@@ -0,0 +1,425 @@
...
+ */
+export function createAutoScrollbarAction(state: ScrollAreaState) {
+	// always create the base action first, so we can override any
+	// state mutations that occur there
+	const baseAction = createBaseScrollbarAction(state);
+	const { rootState, scrollbarState } = state;
+
+	const handleResize = debounceCallback(() => {
+		const $viewportEl = rootState.viewportEl.get();
+		if ($viewportEl) {
```
TGlide, line 144: Early return

TGlide (changes requested):

src/lib/builders/scroll-area/create.ts
huntabyte: What do you mean by this? The `scrollAreaEl` is scoped to this specific instance so if this one is destroyed it would be the root right?

TGlide (approved):

Issue #909, opened 2024-01-13
Svelte component libraries should use "NodeNext" for moduleResolution, not "bundler"


rsweeneydev:
### Describe the bug

TypeScript types are not resolved property for a couple melt-ui components. This means intellisense doesn't work. For example, DatePicker component. This is caused by missing fully specified import specifiers. 

For example, in date-picker/index.ts 

```
export * from './create';
export * from './types';
```

Should be 

```
export * from './create.js';
export * from './types.js';
```

To prevent these types of issues, and support the widest range of uses, the authors of TypeScript and SvelteKit recommend that library authors *not* use moduleResolution "bundler". melt-ui is using moduleResolution "bundlers", so these issues are not detected at build and VS Code doesn't add file extensions when auto adding import statements.

Here are the relevant links to the documentation:

* [TypeScript guidance for library authors](https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution-for-libraries)
* [SvelteKit guidance for component library authors](https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution-for-libraries)

To fix this issue, melt-ui should add `"module": "NodeNext"` to it's tsconfig.json and remove `"moduleResolution": "bundler"`. ("module": "NodeNext" implies "moduleResolution": "NodeNext").

This will identify all the places in the library where import specifiers are invalid for modern node.

Note: I've filed a similar issue in bits-ui

### Reproduction

1. Create a new SvelteKit project
2. Set "module": "NodeNext" in tsconfig.json
3. Add melt-ui DatePicker
4. run svelte-check

Result:

```
/Users/robert/code/bits-ui/src/lib/bits/range-calendar/ctx.ts:1:15
Error: Module '"@melt-ui/svelte"' has no exported member 'CreateRangeCalendarProps'. 
import { type CreateRangeCalendarProps, createRangeCalendar } from "@melt-ui/svelte";
import { getContext, setContext } from "svelte";
```

### System Info

```bash
System:
    OS: macOS 14.1.2
    CPU: (10) arm64 Apple M1 Max
    Memory: 5.62 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.17.1 - ~/.nvm/versions/node/v18.17.1/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 9.6.7 - ~/.nvm/versions/node/v18.17.1/bin/npm
[...]

PR #910, opened 2024-01-13
Migrate library to use module: "NodeNext" as recommended by TypeScript and SvelteKit docs

rsweeneydev:
Fixes: #909


rsweeneydev:
Here's more detail on why this change is necessary. The tldr is that TypeScript libraries should use `"module": "NodeNext"` to ensure they are the most compatible. NodeNext enforces the strictest possible settings and will enable the library to be used in node.js, the browser, or a bundler.

I'm intentionally overriding both`module` and `moduleResolution` to align the library with the TypeScript and SvelteKit best practices for library authors. `module: NodeNext` must be used with `moduleResolution: NodeNext`. Since both `module` and `moduleResolution` are set in .svelte-kit/tsconfig.json, we have to override both.

Here's more detail on why libraries should be NodeNext. From the TypeScript handbook:

> When compiling an app, you choose the moduleResolution option for a TypeScript project based on who the module resolution [host](https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution-is-host-defined) is. When compiling a library, you don’t know where the output code will run, but you’d like it to run in as many places as possible. Using "module": "nodenext" (along with the implied ["moduleResolution": "nodenext"](https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-nodenext)) is the best bet for maximizing the compatibility of the output JavaScript’s module specifiers, since it will force you to comply with Node.js’s stricter rules for import module resolution.
>
>...
>
> In short, "moduleResolution": "bundler" is infectious, allowing code that only works in bundlers to be produced. Likewise, "moduleResolution": "nodenext" is only checking that the output works in Node.js, but in most cases, module code that works in Node.js will work in other runtimes and in bundlers.

https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution-for-libraries

They are a little more direct in the reference section of the TypeScript docs.

> A common misconception is that node16 and nodenext only emit ES modules. In reality, node16 and nodenext describe versions of Node.js that support ES modules, not just projects that use ES modules. Both ESM and CommonJS emit are supported, based on the [detected module format](https://www.typescriptlang.org/docs/handbook/modules/reference.html#module-format-detection) of each file. Because node16 and nodenext are the only module options that reflect the complexities of Node.js’s dual module system, **they are the only correct module options for all apps and libraries that are intended to run in Node.js v12 or later, whether they use ES modules or not.**

https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-nodenext

It doesn't seem possible to combine `module: esnext` with `moduleResolution: NodeNext`.

> --moduleResolution node16 and nodenext must be paired with their [corresponding module value](https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-nodenext).

https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-nodenext-1

The inverse is also true. The bundler option only works with esnext.

> --moduleResolution bundler must be paired with the --module esnext option.

And finally, in the Guide, there is a section titled: "[I'm writing a library](https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-writing-a-library)", it says to use `module: node16` and explains why this is the best choice for compatibility. Node16 = NodeNext

TGlide (commented):
Thank you so much!

package.json
AdrianGonz97, line 1: `@sveltejs/package` needs to be updated to the latest for the same reason provided in the Bits PR: https://github.com/huntabyte/bits-ui/pull/277#discussion_r1451766367

AdrianGonz97 (changes requested):

package.json
rsweeneydev: Thanks for catching this. I upgraded @sveltejs/package to 2.2.5 with the latest commit.

TGlide:
Some type errors that'll probably get fixed with #918. So will wait on that for now

PR #912, opened 2024-01-14
Fix: use `moduleResolution: NodeNext`

abdel-17:
Fixes [909](https://github.com/melt-ui/melt-ui/issues/909)


abdel-17:
Nevermind, a PR already exists and this one is not in sync with "develop".

PR #918, opened 2024-01-15
Refactor/withGet

TGlide:
This adds a `withGet` method, that overrides `writable`s set and update methods to update a local `value` variable, that can be accessed with `.get()`.

e.g.
```ts
const original = writable(1);
const awesome = withGet(original);

console.log(awesome.get()) // 1

awesome.set(2)
console.log(awesome.get()) // 2

original.set(3)
console.log(awesome.get()) // 3
```

It also works with `readable` and `derived` stores, but those currently require subscriptions, which then must be manually cleaned up. I'm working on improving this.

**TODO**
- [X] `withGet`
- [X] `withGet.writable`
- [x] `withGet.derived`


huntabyte:
This is gonna be sick!!

Issue #920, opened 2024-01-18
[Table of Contents] when clicking on item, page scrolls to top without triggering scroll function


anatolzak:
### Describe the bug

When clicking on an item in the table of contents, the page immediately scrolls to the html element with the corresponding ID. This behavior is actually wrong, because according to the code, if no custom `scrollFn` is passed to the table of contents builder, then `window.scrollTo({ top: offsetPosition, behavior: scrollBehaviour })` is called and the default behavior is smooth as per the docs, which should result in a smooth scroll. 

However, the reason the scroll is not smooth even though we are calling `window.scrollTo()` as said above is because after that we set `window.location.hash = id` which immediately changes the page scroll position to the item with the ID. 

In order to prevent the browser from scrolling to the element so that we can manually do so, we need to call `event.preventDefault()` when the anchor element is clicked, which is already done. In addition, we also need to ```history.pushState(null, '', `#${id}`);``` which doesn't change the page's scroll position, contrary to setting `window.location.hash` which does cause a page scroll jump.

So in order to fix this issue, we need to replace `window.location.hash = id` with ```history.pushState(null, '', `#${id}`);```

### Reproduction

This issue also becomes obvious when in the docs example of table of contents, we pass a `scrollFn` to the builder which according to the code comments should:
```typescript
/**
 * Here we're overwriting the default scroll function
 * so that we only scroll within the ToC preview
 * container, instead of the entire page.
 */
```
https://github.com/melt-ui/melt-ui/blob/c52c664b5d4e7605d6ea5e07371199fb7fe78c7b/src/docs/previews/table-of-contents/main/tailwind/index.svelte#L14C3-L29C5

but the page scroll jumps instead of performing the desired behavior of scrolling the preview container instead of the page.

And here is a video of the table of contents in the docs not smoothly scrolling which is the default behavior

### System Info

```bash
System:
    OS: macOS 13.4.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 1.69 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v20.9.0/bin/yarn
    npm: 10.1.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
[...]

PR #921, opened 2024-01-18
[table of contents] fix page scroll jump on item click

anatolzak:
closes https://github.com/melt-ui/melt-ui/issues/920

When clicking on an item in the table of contents, the page immediately scrolls to the html element with the corresponding ID. This behavior is actually wrong, because according to the code, if no custom `scrollFn` is passed to the table of contents builder, then `window.scrollTo({ top: offsetPosition, behavior: scrollBehaviour })` is called and the default behavior is smooth as per the docs, which should result in a smooth scroll. 

However, the reason the scroll is not smooth even though we are calling `window.scrollTo()` as said above is because after that we set `window.location.hash = id` which immediately changes the page scroll position to the item with the ID. 

In order to prevent the browser from scrolling to the element so that we can manually do so, we need to call `event.preventDefault()` when the anchor element is clicked, which is already done. In addition, we also need to ```history.pushState(null, '', `#${id}`);``` which doesn't change the page's scroll position, contrary to setting `window.location.hash` which does cause a page scroll jump.

So in order to fix this issue, we need to replace `window.location.hash = id` with ```history.pushState(null, '', `#${id}`);```

This issue also becomes obvious when in the docs example of table of contents, we pass a `scrollFn` to the builder which according to the code comments should:
```typescript
/**
 * Here we're overwriting the default scroll function
 * so that we only scroll within the ToC preview
 * container, instead of the entire page.
 */
```
but the page scroll jumps instead of performing the desired behavior of scrolling the preview container instead of the page.

### Docs Table of Contents After
https://b005dbbf-685b-42e9-9961-5b74e58b0921.s3.us-east-1.amazonaws.com/example-after.mp4

### Docs Table of Contents Before
https://b005dbbf-685b-42e9-9961-5b74e58b0921.s3.us-east-1.amazonaws.com/exaple-before.mp4

### Table of contents used in docs After
https://b005dbbf-685b-42e9-9961-5b74e58b0921.s3.us-east-1.amazonaws.com/docs-after.mp4

### Table of contents used in docs Before
https://b005dbbf-685b-42e9-9961-5b74e58b0921.s3.us-east-1.amazonaws.com/docs-before.mp4

### DIFF
--- a/src/lib/builders/table-of-contents/create.ts
+++ b/src/lib/builders/table-of-contents/create.ts
@@ -372,7 +372,7 @@ export function createTableOfContents(args: CreateTableOfContentsArgs) {
 
 					// Add items hash to URL
 					if (id) {
-						window.location.hash = id;
+						history.pushState(null, '', `#${id}`);
 					}
 				})
 			);


TGlide:
Thanks, but this has already been addressed by #860 :)

PR #932, opened 2024-01-26
docs: fix wrong description of ChangeFn

macmillen:
The docs include the wrong param names. Instead of the `next` param it mentions `prev`.

``` ts
// melt-ui/melt-ui/src/lib/internal/helpers/overridable.ts
export type ChangeFn<T> = (args: { curr: T; next: T }) => T;
```


TGlide (approved):
Thank you!

PR #943, opened 2024-02-01
Add onHighlightedItemChange to listbox builder

### LINKED ISSUE #944: `$highlighted` store doesn't have correct `Value` type inference
### Describe the bug

The `$highlighted` store's value in combobox is always set to `unknown`. The whole store should essentially have a type similar to something like `ComboboxOptionProps<Value=unknown> = ListboxOptionProps<Value>`.

### Reproduction

This is a type inference issue. I can get a repro if necessary, but it is reproducible always on the current package version.

image

### Logs

_No response_

### System Info

```bash
MacOS latest, melt latest, working in latest melt-ui repo with nvm node version. No browsers.
```

### Severity

annoyance

wesharper:
Adds onHighligthedItemChange to listbox builder to support advanced control over highlighting for complex comboboxes with multifaceted UIs.

Observe the following behavior:
https://github.com/melt-ui/melt-ui/assets/13283469/a4f416c7-b8e6-4d0a-90ab-346f11e79d27

In the example, leaving the menu component with the cursor will change the `$highlighted` and `$highlightedItem` stores to `null`, which breaks the expectations for the details panel on the right side. Adding `onHighlightedItemChange` allows the end user to prevent the value from being set to `null` and navigate the details pane.

Additionally, since the `$highlighted` store continues to be derived from the value of `$highlightedItem`, highlighted will continue to stay in sync with the intended value.

**Also resolves #944**


TGlide:
I think this is a good idea. But I also want to override how these types of props work overall (#948). Let me think a bit on this

TGlide (approved):
Could you add it to select too? Thanks!

wesharper:
You just meant the type right? Done!

TGlide:
Closing this in favor of #973

PR #946, opened 2024-02-01
fix: structuredClone polyfill

aqmebrak:
As per this [Caniuse](https://caniuse.com/?search=structuredClone), the structuredClone API is available for around 93.36% of the global browser use. 
Many browser version are not supporting it and are still used by a small percentage of people as it's relatively a recent addition (many last unsupported version are from around 2021-2022).

I propose to use a polyfill to resolve the issue for those browser. The polyfill is loaded as a global polyfill. 

Not sure if this is the correct approach for loading a polyfill, help is much appreciated.


src/lib/builders/date-field/_internal/structuredClone.ts
```diff
@@ -0,0 +1,8 @@
+import structuredClonePolyfill from "@ungap/structured-clone";
+
+// Export existing implementation if available.
+if (!("structuredClone" in globalThis)) {
+	// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+	// @ts-ignore
+	globalThis.structuredClone = structuredClonePolyfill;
```
TGlide, line 7: Rather than overriding global this, I believe having a `clone` method which calls either structuredClone or the polyfill would be best. Unless the issue is coming from an internal library, in which case a PR should be open there instead.

TGlide:
> Unless the issue is coming from an internal library, in which case a PR should be open there instead.

It seems this is indeed the case. Closing

aqmebrak:
I might not understand it correctly but, from what I could see the use of structuredClone are coming from the date-field builder ?
https://github.com/melt-ui/melt-ui/pull/946/files/00bc174fbc33783699b299e26c6e171a4763f226#diff-e05dbf36d185cd38a7371cc84a8bb8aaa529f10734c60c29c572bd66b726e5fbR1892
https://github.com/melt-ui/melt-ui/pull/946/files/00bc174fbc33783699b299e26c6e171a4763f226#diff-e05dbf36d185cd38a7371cc84a8bb8aaa529f10734c60c29c572bd66b726e5fbL161

PR #947, opened 2024-02-01
docs: Fix 'denied' prop name on Tags Input builder docs

devjume:
In the Tags Input builder docs `denied` prop was still referenced as `disallowed` even though it was changed on: https://github.com/melt-ui/melt-ui/commit/072e30053a4b35a39cce5e5a7f8234fe2e7da14a


TGlide (approved):
Thank you!

PR #955, opened 2024-02-05
fix: Close tooltip on scroll

### LINKED ISSUE #945: Tooltips stays open after scrolling away from trigger without moving cursor
### Describe the bug

Tooltips stays open after scrolling away from trigger while maintaining same cursor position

### Reproduction
1. Hover over trigger, and the tooltip will open
2. Start scrolling without moving the cursor position and the tooltip will stay open, which is not the expected behavior. 

https://github.com/melt-ui/melt-ui/assets/53095479/8ad4657f-4bc3-4231-a17b-88e31aaf37e7

Here is a comparison to Radix tooltip closing if you scroll away from the trigger, which is the expected bahviour

https://github.com/melt-ui/melt-ui/assets/53095479/467f5fe9-9963-4f6e-aa75-62bd074ef9a2

### Logs

_No response_

### System Info

```bash
  System:
    OS: macOS 14.3
    Memory: 333.43 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    bun: 1.0.22 - ~/.bun/bin/bun
  Browsers:
    Chrome: 120.0.6099.109
    Edge: 114.0.1823.82
    Safari: 17.3
  npmPackages:
    @sveltejs/kit: ^2.0.0 => 2.5.0 
    bits-ui: ^0.16.0 => 0.16.0 
    svelte: ^4.2.7 => 4.2.9 
```

### Severity

annoyance

huntabyte:
Closes #945 

This PR addresses a bug where the tooltip content would remain open if a user scrolls outside of it, which is unexpected behavior. I've added a listener to the `window` when the content is mounted to handle scroll events that occur outside the content. If the tooltip content is scrollable, we don't close when scrolling that content, only other scroll targets.

I've added tests for this functionality & patched some of the tooltip tests that are currently being skipped to now work. A `setup` test helper for the tooltip tests.

### DIFF
--- a/.changeset/silly-bananas-train.md
+++ b/.changeset/silly-bananas-train.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+Tooltip: fixed a bug where the content would remain open when scrolling outside (Closed: #945)
--- a/src/lib/builders/tooltip/create.ts
+++ b/src/lib/builders/tooltip/create.ts
@@ -7,6 +7,8 @@ import {
 	executeCallbacks,
 	getPortalDestination,
 	isBrowser,
+	isDocument,
+	isElement,
 	isTouch,
 	kbd,
 	makeHullFromElements,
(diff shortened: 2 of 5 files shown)


usahai:
> Closes #945
> 
> This PR addresses a bug where the tooltip content would remain open if a user scrolls outside of it, which is unexpected behavior. I've added a listener to the `window` when the content is mounted to handle scroll events that occur outside the content. If the tooltip content is scrollable, we don't close when scrolling that content, only other scroll targets.
> 
> I've added tests for this functionality & patched some of the tooltip tests that are currently being skipped to now work. A `setup` test helper for the tooltip tests.

Can confirm that this is a bug - thanks for the fix. Looking forward to having the fix.

TGlide (approved):
godsend

PR #958, opened 2024-02-05
Tree View: Multi Select

abdel-17:
This pr adds the option to select multiple items in the tree builder, and some extra goodies.

# Features:
- Support multi select with the `multiple` prop.
- Support passing a custom id to the root.
- `forceVisible` is now a modifiable option.
- All states and options now have a `get()` method attached.
- Type-ahead now uses locale-insensitive string comparison under the hood.
- The `withGet` helper now works with readable stores.

# Fixes:
- Focus not being set correctly to the first selected item.

# Breaking Changes:
- The `data-melt-id`, `data-id`, and `data-group-id` props have been replaced with the `id` prop.
- The `selectedItem` state has been replaced with `selected`, which store the ids rather than the HTML element itself, and a derived `firstSelected` for convenience.

# Issues
Currently, there is a minor issue where the (expanded, selected) states get out of sync, so the user has to manually press Space to select the item without expanding.

https://github.com/melt-ui/melt-ui/assets/88583085/bf09e48f-c7b2-41be-91da-9ab91aea57e1


src/docs/data/builders/tree.ts
```diff
@@ -28,43 +36,37 @@ const builder = builderSchema(BUILDER_NAME, {
...
 	],
 	helpers: [
 		{
-			name: 'isCollapsedGroup',
+			name: 'isExpanded',
 			description:
-				'A derived store that returns a function that returns whether or not the item is collapsed.',
-			type: 'Readable<(itemId: string) => boolean>',
+				'A derived store that returns a function that returns whether or not the item is expanded.',
+			type: 'WithGet<Readable<(itemId: string) => boolean>>',
```
huntabyte, line 56: Do we want to expose the `WithGet` to the consumer? We don't have this documented anywhere and it adds an element of WTF is this to someone reading the API ref.

It also would cause issues with other non-melt dependencies that may expect just a `Readable` type.

IMO we return the stores without the `get` methods and keep it simple.,

src/lib/builders/tree/create.ts
```diff
@@ -250,86 +213,134 @@ export function createTreeView(args?: CreateTreeViewProps) {
...
 		// Select all 'treeitem' li elements within our root element.
-		items = Array.from(rootEl.querySelectorAll('[role="treeitem"]')).filter(
-			(el) => !isHidden(el as HTMLElement)
-		) as HTMLElement[];
+		return Array.from(rootEl.querySelectorAll('[role="treeitem"]')).filter(
+			(el): el is HTMLElement => isHTMLElement(el) && !isHidden(el)
+		);
+	}
+
+	function focusItemStartingWith(letter: string, items: HTMLElement[], nodeIdx: number) {
```
huntabyte, line 261: This can be moved outside the closure so the function is only created once rather than on each instantiation since it receives everything it needs as args.

src/lib/builders/tree/create.ts
```diff
@@ -250,86 +213,134 @@ export function createTreeView(args?: CreateTreeViewProps) {
...
+			if (!item.id) return;
+			item.focus();
+		}
 	}
 
-	function getElementAttributes(el: HTMLElement) {
-		const hasChildren = el.hasAttribute(ATTRS.EXPANDED);
-		const expanded = el.getAttribute(ATTRS.EXPANDED);
-		const dataId = el.getAttribute(ATTRS.DATAID);
+	const collator = Intl.Collator(undefined, {
```
huntabyte, line 294: This should also be able to be moved outside the closure

src/lib/builders/tree/create.ts
```diff
@@ -250,86 +213,134 @@ export function createTreeView(args?: CreateTreeViewProps) {
...
+		usage: 'search',
+		sensitivity: 'base',
+	});
 
-		return {
-			hasChildren,
-			expanded,
-			dataId,
-		};
+	function localeStartsWith(item: HTMLElement, letter: string) {
```
huntabyte, line 299: Same here (outside closure)

src/lib/builders/tree/create.ts
```diff
@@ -250,86 +213,134 @@ export function createTreeView(args?: CreateTreeViewProps) {
...
-			expanded.update((prev) => [...prev, dataId]);
+		if (elementIsExpanded(el)) {
+			expanded.update((prev) => prev.filter((id) => id !== el.id));
 		} else {
-			expanded.update((prev) => prev.filter((item) => item !== dataId));
+			expanded.update((prev) => [...prev, el.id]);
 		}
 	}
 
+	function isTreeItem(el: HTMLElement) {
```
huntabyte, line 315: This can be moved as well

src/lib/builders/tree/create.ts
```diff
@@ -250,86 +213,134 @@ export function createTreeView(args?: CreateTreeViewProps) {
...
+	}
+
 	function elementHasChildren(el: HTMLElement) {
-		return el.hasAttribute(ATTRS.EXPANDED);
+		return el.hasAttribute('aria-expanded');
 	}
 
 	function elementIsExpanded(el: HTMLElement) {
-		return el.getAttribute(ATTRS.EXPANDED) === 'true';
+		return el.ariaExpanded === 'true';
```
huntabyte, line 324: This can also be moved outside

src/lib/builders/tree/create.ts
```diff
@@ -250,86 +213,134 @@ export function createTreeView(args?: CreateTreeViewProps) {
...
+	}
+
+	function elementIsSelected(el: HTMLElement) {
```
huntabyte, line 327: Same for this one

src/lib/builders/tree/types.ts
```diff
@@ -26,16 +26,24 @@ export type CreateTreeViewProps = {
 	 * @see https://melt-ui.com/docs/controlled#change-functions
 	 */
 	onExpandedChange?: ChangeFn<string[]>;
-};
 
-export type TreeParts = 'label' | 'item' | 'group';
+	/**
+	 * Pass `true` to enable multi select.
```
huntabyte, line 31:
```suggestion
	 * If `true`, multiple items can be selected
```

huntabyte:
> Currently, there is a minor issue where the (expanded, selected) states get out of sync, so the user has to manually press Space to select the item without expanding.

I think this should get addressed before merge so we aren't introducing introducing a bug to introduce a feature.

src/docs/data/builders/tree.ts
abdel-17: We can keep it out of the documentation sure. I forgot about it being internal. But, having a get() method attached shouldn’t cause any issues. You can still pass it to functions that expect a Readable.

TGlide: Agree with removing it out of the documentation, but keeping the get method.

TGlide:
> Currently, there is a minor issue where the (expanded, selected) states get out of sync

What exactly is the issue? They don't need to be the same value.

APG recommends two models for selection in multi-trees. We can be opinionated and choose one *or* allow both models, changeable via a prop.

Click should probably follow the same behaviour as the `space` key. Unfortunately, they don't have an example component with multi-select, so we'll have to try it out to see if it makes sense.

mustafa0x:
I feel that click to select is a bit unintuitive. This isn't the behavior of Windows explorer, Mac Finder, nor Google Drive, to name a few. To select while clicking it is required to press ctrl/cmd.

abdel-17:
Maybe we can, instead, have a `multiSelect` prop with three options.

- false (default)
- true
- "cmd-click"

mustafa0x:
$0.02: `mouseMultiSelect='meta-click' | 'click' | false`

abdel-17:
So you suggest separate prop for controlling mouse selection?

mustafa0x:
Yes I think that makes more sense, due to the high likelihood that it won't be the only aspect  of multi-select that is customizable.

TGlide:
Closing, lets add this in Runes version

PR #968, opened 2024-02-07
Docs: use pagefind's built in debouncer instead of our own

Hugos68:
# Why?

Easier to reason about and it's build into pagefind meaning less bugs, more defined behaviour.

See: https://pagefind.app/docs/api/#debounced-search

PS: I see `pnpm format && pnpm lint` edited a bunch of files, probably from someone who pushed without running those, the actual changes are found [here](src/routes/(landing-ui)/search.svelte).


TGlide (approved):
Thank you!

PR #984, opened 2024-02-15
New Builder: Spin Button

abdel-17:
This PR adds a new **Stepper**  builder, also known elsewhere as Spinbutton.

https://github.com/melt-ui/melt-ui/assets/88583085/3cdc9c44-6f94-4881-8d44-bbb849496f5c

Follows the [APG Spinbutton Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/spinbutton/#:~:text=A%20spinbutton%20is%20an%20input,the%20minute%20of%20an%20hour.)

## Minor Caveats
When VoiceOver announces a stepper, it suggesting controlling it by pressing Control-Option-Up/Down, but this functionality only works with the native `<input type="number">`.

voiceover-bug

For some reason, I wasn't able to listen to the Option and Up/Down keydown events to manually get around it. Only the Control key emits a `keydown` event when VoiceOver is active and the stepper is in focus. We should at least mention this issue in the docs.

Related issue: https://github.com/microsoft/fluentui/issues/13606


abdel-17:
So, there's a small bug with `loop` and `step`.

Consider the case where `step: 2` and `loop: true`. If `value` is currently `max` and I increment it, it should be `min + 1`, but with the current implementation, it's `min`. Trying to fix this now.

huntabyte:
Are we sure we want to call this a Stepper? I know that's what it's announced as to screen readers, but when I see "Stepper" as a component I think of something like this: https://www.skeleton.dev/components/steppers

abdel-17:
Prior to researching implementations, I had never heard the term "spinbutton". Maybe it's because English is not my first language, but "stepper" to me sounds more intuitive. If you google stepper input, it shows what you would call a "spinbutton" rather than Skeleton's stepper. Maybe we can call it StepperInput?

image

huntabyte:
Yeah when you put "input" alongside it, it makes more sense. I'll defer to @TGlide on this one b/c I believe he wanted to call it a SpinButton originally

TGlide:
Let's call it `spinbutton`. Both the APG pattern and the aria role is called like so, and as @huntabyte said, Stepper may be confusing.

abdel-17:
Yeah sure

huntabyte:
Reminder to myself to look into the VO issues for this one

huntabyte:
Hey @TGlide, once you give the implementation a review/approval I'll add in unit tests for this. Just don't want to have to change a bunch of tests if you propose an architectural change!

TGlide:
Closing, lets add this with Runes

PR #988, opened 2024-02-16
Fix: `usePortal` Action

abdel-17:
# Problem
If `usePortal` is used as a Svelte action and the `update()` function receives an `undefined` target, it incorrectly throws an error rather than fallback to `document.body`.

# Fix
This PR correctly handles this edge case by first checking if the portal target is `undefined`.


TGlide (approved):

TGlide:
Rejecting because of changes upstream. Will fix it

Issue #815, opened 2023-12-11
🐛Bug: Datepicker `$trigger` should not be triggering when `createDatePicker` has `disabled` set to true


mcmxcdev:
**Describe the bug**
When using `createDatePicker` with `disabled: true`, I would expect the datepicker to not allow any interaction.

Turns out the calendar can still be opened and closed though through click and tabbing to the calendar button.

**Expected behavior**

Calendar should not be openable. Either add `disabled` or `tabindex=-1` to the `$trigger` button.

**Video showcase**

[Screencast from 2023-12-11 12-05-06.webm](https://github.com/melt-ui/melt-ui/assets/16797721/963fec02-ec31-4c32-b7ed-17540e5b1b3f)

PR #989, opened 2024-02-16
fix: disabled date picker triggers

huntabyte:
Closes: #815


lolcabanon:
Hey! I was debugging some issue related to trigger not being disabled.

**Happy to see it's fixed already!** :fire:

PR #991, opened 2024-02-16
Fix: Generic Type Defaults

### LINKED ISSUE #907: `Select` doesn't allow `true` to be set for `multiple` prop
### Describe the bug

We are updating our custom `Select` component to work with single and multiple selections.

When trying to provide `true` as default value for `multiple`, we encounter the following TS issue:

Screenshot from 2024-01-12 10-55-06

`CreateSelectProps['multiple']` should allow `boolean | undefined` instead of `false | undefined`

### Reproduction

N/A

### Logs

_No response_

### System Info

```bash
N/A
```

### Severity

annoyance

abdel-17:
Fixes #907

### DIFF
--- a/src/docs/components/nav/theme-switch/theme-switch.svelte
+++ b/src/docs/components/nav/theme-switch/theme-switch.svelte
@@ -3,10 +3,10 @@
 	import type { Theme } from './types.js';
 
 	export const getThemeCtx = () => {
-		return getContext<Select<Theme>>('theme');
+		return getContext<Select<Theme, false>>('theme');
 	};
 
-	export const setThemeCtx = (ctx: Select<Theme>) => {
+	export const setThemeCtx = (ctx: Select<Theme, false>) => {
 		setContext('theme', ctx);
 	};
 
--- a/src/docs/previews/calendar/main/tailwind/LocaleCombobox.svelte
+++ b/src/docs/previews/calendar/main/tailwind/LocaleCombobox.svelte
@@ -16,9 +16,15 @@
 		}),
 	);
 
-	export let onSelectedChange: CreateComboboxProps<string>['onSelectedChange'] =
-		undefined;
-	export let defaultSelected: CreateComboboxProps<string>['defaultSelected'] = {
+	export let onSelectedChange: CreateComboboxProps<
+		string,
+		false
+	>['onSelectedChange'] = undefined;
+
+	export let defaultSelected: CreateComboboxProps<
+		string,
+		false
+	>['defaultSelected'] = {
 		value: 'en-US',
 		label: 'English (US)',
 	};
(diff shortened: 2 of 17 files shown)


huntabyte:
CleanShot 2024-02-16 at 15 26 24@2x

This change now requires an explicit `true` in the type argument as well as in the form of a `multiple` prop.

abdel-17:
This problem exists in the "develop" branch as well. It happens when you specify the generic type. It doesn't infer them from the props anymore, so you must pass a generic explicitly.

Screenshot 2024-02-16 at 10 34 05 PM

abdel-17:
I agree that discriminated unions are much nicer. I'll explore that solution.

huntabyte:
@abdel-17 I see, I didn't notice that introduced the requirement of the multiple generic as well. I still stand by my comment of it being a regression in terms of DX vs just requiring the `type` prop to discriminate, which is more intuitive than having to figure out that on top of specifying `multiple: true` you also have to now pass in your type _and_ the fact that its multiple in the form of a generic. It's a bit bass ackwards.



```ts
const stuff = createSelect<string, true>({
	multiple: true;
})
```

vs

```ts
const stuff = createSelect<string>({
	type: 'multiple'
})
```

huntabyte:
> I agree that discriminated unions are much nicer. I'll explore that solution.

I know when I discussed this with @TGlide in the past he was uneasy about the idea of requiring the `type` prop but maybe his mind has changed now that all these issues around types keep popping up 😂

abdel-17:
> > I agree that discriminated unions are much nicer. I'll explore that solution.
> 
> I know when I discussed this with @TGlide in the past he was uneasy about the idea of requiring the `type` prop but maybe his mind has changed now that all these issues around types keep popping up 😂

You can definitely do discriminated unions without requiring the type prop! I'll commit changes in a moment.

huntabyte:
> > > I agree that discriminated unions are much nicer. I'll explore that solution.
> > 
> > 
> > I know when I discussed this with @TGlide in the past he was uneasy about the idea of requiring the `type` prop but maybe his mind has changed now that all these issues around types keep popping up 😂
> 
> You can definitely do discriminated unions without requiring the type prop! I'll commit changes in a moment.

Discriminated unions without a discriminator... I'm interested to see this one 👨‍🍳

abdel-17:
@huntabyte The basic idea is to make the discriminator optional in the "single" case and required in the "multiple" case.
```ts
type SingleAccordionProps = {
	multiple?: false;
	defaultValue?: string;
};

type MultipleAccordionProps = {
	multiple: true;
	defaultValue?: string[];
};

export type CreateAccordionProps = SingleAccordionProps | MultipleAccordionProps;

// Doesn't error
const _1: CreateAccordionProps = {
	defaultValue: 'a',
};

// Doesn't error
const _2: CreateAccordionProps = {
	multiple: false,
	defaultValue: 'a',
};

// Errors
const _3: CreateAccordionProps = {
	defaultValue: ['a'],
};

// Doesn't error
const _4: CreateAccordionProps = {
	multiple: true,
	defaultValue: ['a'],
};
```

I'm facing some issues with type inference in the builder function itself, but I'll work my way through it.

TGlide:
Closing because of internal discussions on Discord. A select overhaul is coming soon

PR #993, opened 2024-02-16
fix(Toast): use as const` for close element `type` attribute

ksmoore17:
type property on toast close element was returning as string where ts expects `'button' | 'reset' | 'submit' | null | undefined`. Similar to https://github.com/melt-ui/melt-ui/pull/859.

fixes ex:
image


huntabyte (approved):
Thank you!

PR #995, opened 2024-02-17
fix: portal issues

huntabyte:
I'm not sure how an issue hasn't been created for this yet, but we were handling nested portals incorrectly and doing some unnecessary searching when a custom prop was provided.

### DIFF
--- a/src/lib/builders/dialog/create.ts
+++ b/src/lib/builders/dialog/create.ts
@@ -42,7 +42,7 @@ const defaults = {
 	closeOnOutsideClick: true,
 	role: 'dialog',
 	defaultOpen: false,
-	portal: 'body',
+	portal: undefined,
 	forceVisible: false,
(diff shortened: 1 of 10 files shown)


src/lib/builders/dialog/create.ts
```diff
@@ -256,11 +256,11 @@ export function createDialog(props?: CreateDialogProps) {
 	const portalled = makeElement(name('portalled'), {
 		stores: portal,
 		returned: ($portal) => ({
-			'data-portal': $portal ? '' : undefined,
+			'data-portal': $portal !== null ? '' : undefined,
```
TGlide, line 259: We should probably standardize this in an util

huntabyte: Good call!

TGlide (approved):

Issue #997, opened 2024-02-21
🌟Feature request: `createToaster()` prop to pause all toast closing delay on hover  


lolcabanon:
I'd like to be able to pause **all** toasts when hovering on any toast.

**Describe the solution you'd like**

A `createToaster()` prop :

```js
const toaster = createToaster({
  pauseAllOnHover: true // false by default
})
```

**Describe alternatives you've considered**

Custom event listener?

I am also working on a PR, it does not seem complicated to add.

PR #998, opened 2024-02-21
Added prop to createToaster() to pause all timers on hover

lolcabanon:
See issue #997.

A `createToaster()` prop to pause all timers on any toast hover :

```js
const toaster = createToaster({
  pauseAllOnHover: true // false by default
})
```

Let me know if you'd want this feature, and I'll update the doc accordingly.

:)


TGlide:
Maybe a `group` option, similar to `createTooltip` could make more sense... Or, instead of a `pauseAllOnHover`, maybe a `hover` prop, with threeavailable options `pause`, `pause-all` and `null`, the default being `pause`.

lolcabanon:
> Maybe a `group` option, similar to `createTooltip` could make more sense... Or, instead of a `pauseAllOnHover`, maybe a `hover` prop, with threeavailable options `pause`, `pause-all` and `null`, the default being `pause`.

`hover` prop makes sense to me. I'll update and finish the PR :smile:

Issue #996, opened 2024-02-21
Calendar $prevButton and $nextButton trigger a form submit when clicked.


lolcabanon:
### Describe the bug

Calendar `$prevButton` and `$nextButton` trigger a form submit when clicked.

The solution is to add `type=button` to the button. Shouldn't this be in the props of the builders? Maybe you did not include it in case one would use another element than a button?

### Reproduction

- Create a DatePicker or Calendar component
- Place inside form
- Click to open calendar, then click next or previous button
- Form is submitted unexpectedly

### System Info

```bash
System:
    OS: Linux 6.1 Manjaro Linux
    CPU: (8) x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
    Memory: 3.89 GB / 15.50 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    npm: 9.9.2 - ~/.local/share/pnpm/npm
    pnpm: 8.14.3 - ~/.local/share/pnpm/pnpm
  npmPackages:
    @melt-ui/svelte: ^0.74.1 => 0.74.1 
    @sveltejs/kit: ^2.5.0 => 2.5.0 
    svelte: ^4.2.11 => 4.2.11
```

### Severity

annoyance

PR #1003, opened 2024-02-23
fix(Calendar/Range Calendar): use `'button'` for button elements' `type`

ksmoore17:
Closes: #996


huntabyte:
is this logic already applied to the `Range Calendar`? If not, we should include that in here as well.

ksmoore17:
Good call

TGlide (approved):

PR #1004, opened 2024-02-23
fix(Toolbar): set correct tabindex for buttons and links

fbbdev:
At present, the tabindex attribute of toolbar items is initialized only for group items, never for simple buttons or links.

As a result, when there is no group item, or the first item is a button/link, the toolbar becomes inaccessible by keyboard navigation.

This patch fixes the problem and introduces a comprehensive suite of regression tests, both by changing the order of items and by removing group items entirely.


huntabyte (approved):
Thank you!

Issue #1010, opened 2024-02-25
[Tooltip]: content styles reset when open state changes given content is always mounted and forceVisible is true


anatolzak:
### Describe the bug

If you pass `forceVisible: true` and always keep the content mounted, whenever `$open` changes, the content styles reset (`top: auto; left: auto`) and quickly go back to the trigger's position.

https://github.com/melt-ui/melt-ui/assets/53095479/ce93277a-6ecd-4d95-9a9e-35bf5cfdec8c

### Reproduction

https://github.com/anatolzak/melt-ui-tooltip-content-floating-styles-overriding-bug
https://melt-ui-tooltip-content-floating-styles-overriding-bug.vercel.app/

### System Info

```bash
System:
    OS: macOS 14.3
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 608.16 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.15.0 - ~/.nvm/versions/node/v18.15.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v20.9.0/bin/yarn
    npm: 9.5.0 - ~/.nvm/versions/node/v18.15.0/bin/npm
    pnpm: 8.11.0 - /usr/local/bin/pnpm
    bun: 1.0.22 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 122.1.63.161
    Chrome: 120.0.6099.109
    Edge: 114.0.1823.82
    Safari: 17.3
  npmPackages:
    @sveltejs/kit: ^2.0.0 => 2.5.0 
    svelte: ^4.2.9 => 4.2.9
```

### Severity

annoyance

anatolzak:
@mcmxcdev what is the melt-ui version you are using?
you might be experiencing this issue https://github.com/melt-ui/melt-ui/issues/1005 that got fixed recently

mcmxcdev:
> @mcmxcdev what is the melt-ui version you are using? you might be experiencing this issue #1005 that got fixed recently

I upgraded to latest just before making this issue comment. I think my issue is actually more related to the fact that the popover closes and then the tooltip loses its anchor point, so I might just need to disable closeOnOutsideClick for popover.

anatolzak:
@mcmxcdev your issue is actually a different one that what this one was about.
[...]

PR #1011, opened 2024-02-25
fix(tooltip): overriding floating styles when forceVisible true

anatolzak:
fixes https://github.com/melt-ui/melt-ui/issues/1010

If you pass `forceVisible: true` and always keep the content mounted, whenever `$open` changes, the content styles reset `(top: auto; left: auto)` and quickly go back to the trigger's position.

The reason for this is because of the following lines of code

https://github.com/melt-ui/melt-ui/blob/39a62bfad15af2e0b9f689afee36e845ac5acd84/src/lib/builders/tooltip/create.ts#L201C5-L203C8

```typescript
style: styleToString({
}),
```

If the content is always mounted, then floating-ui already set the `top` and `left` style attributes on the first render. However, when `$open` updates, the returned props get recalculated and since`$isVisible` is `true` (because `forceVisible` is `true`), then the style attribute gets reset to an empty string. 

So if `$isVisible` is `true`, we need to make sure that we do not return an empty string for the `style` property because that would remove any style properties set by vanilla js code in the tooltip

```typescript
// src/lib/internal/actions/floating/action.ts
Object.assign(floating.style, {
	position: options.strategy,
	top: `${y}px`,
	left: `${x}px`,
});
```

AND we need to make sure that we do not set the style property to undefined because svelte will remove any style properties set by vanilla js code. 

So the solution is to remove undefined values to not mess up properties set on an element by vanilla js.

So reseting the style property causes the tooltip content to jump to the left part of the page and then back to being anchored to the trigger.

### Before (I added `transition-property: all` to demonstrate the issue)

https://github.com/melt-ui/melt-ui/assets/53095479/709af6a4-81a5-49c9-9a17-8e2214be9538

### After 
the reason there are no logs is because floating-ui doesn't need to recompute the position of the content since it was already calculated in the initial render (since the content is always mounted)

https://github.com/melt-ui/melt-ui/assets/53095479/2d5b49be-82e8-4fdb-8fa9-2f9e91edb763


anatolzak:
Hey @huntabyte, could you please take a quick look at this PR, it was a pretty simple fix to the tooltip issue where we accidentally overrode the styles set by floating-ui on the tooltip content. This caused the tooltip content to jump across the page if it is always mounted. Thanks! :)

huntabyte (approved):
Thank you!

huntabyte:
Ah, nice find! I actually ran into this issue with Bits and implemented by own workaround which basically looked like what you did here.

PR #1015, opened 2024-02-26
fix: type narrowing for builder props

anatolzak:
added `as const` to every object returned from...
```typescript
makeElement(name('group'), {
	returned: () => {
		return {
			...
		} as const; // <------
	},
});
```

### Before
Untitled 2

### After
Untitled 1


TGlide:
I wonder if we can be smarter with this. Using something like this: https://xebia.com/blog/typescript-5-0-and-the-new-const-modifier-on-type-parameters/

anatolzak:
I tried using the const modifier and was almost able to do it but it seems like TS has a weird bug that breaks its behavior when used in more complex situations.

I had to create a way smaller and simpler version of the makeElement function to test the const modifier without too many flowing variables to first prove a way to do it before trying to apply it to a much more complex scenario. But again, TS seems to have a weird issue that prevents me right now from moving forward with it.

[TS Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAysD2AnCBnKBeKBXAdgax3gHccBtAXQG4Aoa0SKAUQBsIBbCHYAYQENnmAI14BjPAB4YUCAA9gnACZo4SVABooABWlzFaeIIBWEEcAB8GKAAoAbvyyoAXLACUGC9oA+1qwDp-vIgA5ijOvDggFG7oHi40dODQALK8eBAs7JzAAILBKJI68jhKsAjIKBraskUlBsamGgBKhXpMrBxcfALCYpKVZhaYAN7UUGNQyMBYiDgQCs6NNONQKGVOsDQAvrQi8DirUGyp6e1ZluKj47v7wLAtxcprFVAA9C8rT5dj1wdVug9QRomJAKcSrRAASxwQQ0uAIxBwZg0bygYEQ8DAKC+UB+t2a1VaGQ6PH4QlEEhg-Ve70m01mCmoZisgRCzhSaSJWVyIT6WiaZhcCwgUxmABVEuJGoMLCNlrioEMJsK6XMNKtVGhNpYWSgluNaTMlSL6VZ1eU3Lw0EDjeLIJKzFt4i8AFTUeUKXjAXgARmcsv1EF4Cj2zBAUAAFhABPBnAAiIhIZgKWNbajOl5uvYHD1e72WI4c05cKz+sZm9akb0aADk3urGmAiAc5DU2IN9OcVmiMuxy3bCt7y3GkejzmrCcQSfrg7Gmz1s9bmzitBdmZuUBzvAATH7sSPmDGPpDoan02vs57t-njpzi6WPhrnJWa3WG02IC228qZnNO+WUN2A5DgGxpAcBQ77oe46Jgo07gVAc7Ypsi7LtQq7upeADMu7LJBzjglCQSnhmGFeph16Fpkd7Yv+T5VlAtb1lAjbNq2fbfh21i0VAz5QLG3qxm+zaAfe7GgaJ4F4QxE5Tmx4GIcsyHUEulBAA)

The idea behind what we are trying to accomplish is to get the props returned from `makeElement` to be interpreted as const. We get the props from the `returned` method, which takes in a parameter--the store values that are passed to the makeElement function. 

I was able to get the props to be interpreted as const but only if I didn't declare the stores parameter in the `returned` method.

image

However, for some reason, when I declare the stores parameter in the `returned` method, TS doesn't  return the object as const anymore.

image

The weird thing is that TS knows the type of the stores parameter from the generics, but if you explicitly add the type for the stores parameter, the returned props are then interpreted as const. But the whole point was to not have to write "unnecessary" code.

image

image

anatolzak:
@TGlide Unless I am missing something with the const modifier, it seems like it's a little broken and so maybe we should just use `as const` for now, as it is actually already used in many builders, and then if someone can figure out a better way to do this, they can create a PR

TGlide (approved):

PR #1025, opened 2024-02-29
feat: Popover overlay

### LINKED ISSUE #1023: Feature Request: Overlays for all modal-like builders
Currently, we have components like the Menus, Combobox, Select, etc. that act similar to modals by default in the sense that they prevent scroll when open, close on escape, and close on outside clicks. What these don't do however is prevent and disable interactions outside of the menu/content.

For example, here's the currently default behavior of the Select:

https://github.com/melt-ui/melt-ui/assets/64506580/5dbc203f-2bbb-4a16-b17e-571666bce687

Since the scroll is disabled, it appears as though I shouldn't be able to do anything outside the content, but I can still interact with things outside of the "modal", which creates this weird in-between of "I can interact with things only within the current viewport", which feels a bit sloppy.

Perhaps the simplest way to solve this problem would be to add an additional `overlay` element to the various builders where we support modal behaviors, similar to the dialog. This wouldn't be a breaking change and opting out would be as simple as not using the overlay.

This could be handled by the user manually today, but only if they are using `forceVisible` and can stick the overlay behind the synced `open` state, so I think adding a 1st party element would be beneficial.

huntabyte:
This PR adds an (optional) `overlay` element builder to the Popover to enable modal behavior. To simplify the review process, I'll create separate PRs to add this functionality to the other floating builders and once they are all merged/released will close #1023 

It will be portalled to the same destination as the `content`.

- [x] Update `data-portal` prop once #1018 is merged and before this is merged
- [x] Add example preview component to docs


TGlide:
Added a comment in #1023

Issue #985, opened 2024-02-16
Rendering days of month changes to default calendar after selecting a day from calendar


B-Esmaili:
### Describe the bug

I setup the date-picker with `persian-calendar`  and `fa-IR` locale. In First render it works as expected , but when i select a day (which is in other gregorian month than current month) it renders the days based on gregorian calendar (not persian clendar). To reproduce the issue just use following code and click over day 1.

### Reproduction

[https://stackblitz.com/edit/github-tsc74y?file=src%2Flib%2Fcomponents%2FAccordion.svelte](https://stackblitz.com/edit/github-tsc74y?file=src%2Flib%2Fcomponents%2FAccordion.svelte)

### Severity

annoyance

B-Esmaili:
> Please provide a [reproduction](https://melt-ui.com/repro).
> 
> More info

I updated the post with link for playground.

huntabyte:
Thank you for taking the time to do that! We'll look into it!

B-Esmaili:
I investigated the issue , apparently the problem is the fact that [parseStringToDateValue](https://github.com/melt-ui/melt-ui/blob/c5cf4aa07222e574a14598960be239000426b383/src/lib/internal/helpers/date/utils.ts#L70C17-L70C39) function does not take into account the `Calendar` specified by default `DateValue`. The revamped version of this function could be something like following :

``` typescript
export function parseStringToDateValue(dateStr: string, referenceVal: DateValue): DateValue {
	let dateValue: DateValue;
	if (referenceVal instanceof ZonedDateTime) {
		dateValue = parseZonedDateTime(dateStr);
	} else if (referenceVal instanceof CalendarDateTime) {
		dateValue = parseDateTime(dateStr);
	} else {
		dateValue = parseDate(dateStr);
	}
	
	return toCalendar(dateValue, referenceVal.calendar);	 
}
```
This is my speculation , maybe i am wrong but this is based on my understanding of the code. I hope this would be a help for you to further investigate the issue.

B-Esmaili:
Yes definitely i'll submit a PR in my next free time.

PR #1026, opened 2024-02-29
fix: use calendar of reference value when parsing date

B-Esmaili:
Closes: #985
This `PR` fixes the issue in `parseStringToDateValue` function which parses date strings to `DateValue` but currently ignores the `calendar` used in original value.


huntabyte (approved):
Thank you!

B-Esmaili:
> Thank you!

One more thing worth mention is i found another bug in `@internationalized\date` package and submitted another `PR` there as well. I just wanted to let you know in case you encounter wrong rendering of days it is because of that bug and not related to this library.

PR #1035, opened 2024-03-02
feat: Menu Overlays

### LINKED ISSUE #959: 🐛 Bug: Unexpected Menu behaviors
The melt-ui context menu currently has a few different behaviors from the default browser context menu:
- Pointer events (hover, mousedown, click, etc) still register for elements outside of the context menu
- Clicking outside of the context menu both closes the menu and registers the click event to the clicked element
- Pressing tab closes the menu

These differences make the melt context menu feel unintuitive in our application.  It'd be great if we had options to configure these behaviors to feel closer to the native context menu.  Potentially a  `modal: boolean` would be good enough?

huntabyte:
This PR adds an optional `overlay` builder element to the Dropdown & Context Menus to enable modal behavior where outside interactions can be blocked by the overlay while the menu is open.

Closes #959

### DIFF
--- a/src/docs/content/builders/dropdown-menu.md
+++ b/src/docs/content/builders/dropdown-menu.md
@@ -115,6 +115,9 @@ The [Modal Dropdown](#modal-dropdown) example below demonstrates this behavior i
 
 ### Modal Dropdown
 
+An example using the `overlay` builder element to prevent interaction with the rest of the page
+while the menu is open.
+
 <Preview code={snippets.modal}>
     <svelte:component this={previews.modal} />
 </Preview>
--- a/src/docs/content/builders/dropdown-menu.md
+++ b/src/docs/content/builders/dropdown-menu.md
@@ -5,9 +5,11 @@ description:
 ---
 
 <script>
-    import { KbdTable, APIReference } from '$docs/components'
+    import { KbdTable, APIReference, Preview } from '$docs/components'
     export let schemas
     export let keyboard
+	export let snippets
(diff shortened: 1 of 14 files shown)


huntabyte:
Related #1023

src/docs/content/builders/dropdown-menu.md
```diff
@@ -84,6 +88,40 @@ occurring.
 </div>
 ```
 
+### Modal Behavior
+
+To give the dropdown menu complete modal behavior, where an overlay is rendered behind the menu to
+prevent interaction with the rest of the page while open, use the `overlay` builder element.
```
TGlide, line 94: I don't think we should requuire an `overlay` to block interaction outside. A `modal` prop would make more sense. 

Also, I've never seen an overlay for menu components like ours, is it really needed?

TGlide (changes requested):

src/docs/content/builders/dropdown-menu.md
huntabyte: A `modal` prop would make sense if we were a component library because then we could conditionally render an overlay (as is the case with Radix and others). Unless we want to inject a node into the DOM and take the ability to control the overlay from the user I'm not sure it makes more sense.

A `modal` prop would only make sense if we removed `closeOnOutsideClick`, `preventScroll`, and `closeOnEscape` in favor of a single prop. Not sure if users would like this since it removes flexibility.

Issue #1024, opened 2024-02-29
🌟Feature request: Combobox Trigger


huntabyte:
**Is your feature request related to a problem? Please describe.**
The combobox builder doesn't expose a trigger except for the input. Oftentimes you want the input _and_ a button (think arrow down icon button) to act as triggers.

**Describe the solution you'd like**
The listbox trigger to be exposed from the combobox to (optionally) use in addition to the input.

**Describe alternatives you've considered**
Manually wiring it up. Cons are that the functionality already exists for the listbox, it's just a matter of returning it and updating conflicting attrs with input.

**Additional context**
This feature would enable users to easily create something like this: https://headlessui.com/react/combobox

silenceofthewans:
Another use case for this would be a Todoist like sheet that presents combobox-like functionality inside a sheet:

https://github.com/melt-ui/melt-ui/assets/61865590/a464a9e4-e8b7-44e0-bcd9-62f978319b05

huntabyte:
@silenceofthewans This could be accomplished today using a dialog with the combobox open by default inside the dialog!

PR #1038, opened 2024-03-03
feat: Combobox `trigger` element builder 

huntabyte:
This PR adds the `trigger` element builder to the `createCombobox` return elements.

It is optional and can be used to toggle the menu open and closed.

Closes: #1024

### DIFF
--- a/src/lib/builders/combobox/create.ts
+++ b/src/lib/builders/combobox/create.ts
@@ -13,7 +13,7 @@ import {
 	omit,
 } from '$lib/internal/helpers/index.js';
 import type { MeltActionReturn } from '$lib/internal/types.js';
-import { writable } from 'svelte/store';
+import { get, writable } from 'svelte/store';
 import { createListbox } from '../listbox/create.js';
 import type { ComboboxEvents } from './events.js';
 import type { ComboboxSelected, CreateComboboxProps } from './types.js';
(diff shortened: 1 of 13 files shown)


TGlide:
We should probably make it so clicking on the input does not close combobox, as part of this.

src/lib/builders/combobox/create.ts
```diff
@@ -94,6 +94,32 @@ export function createCombobox<
...
+			return {
+				...omit($trigger, 'action', 'id'),
+			};
+		},
+		action: (node: HTMLElement): MeltActionReturn<ComboboxEvents['trigger']> => {
+			const unsubEvents = executeCallbacks(
+				addMeltEventListener(node, 'click', () => {
+					const inputEl = document.getElementById(get(listbox.elements.trigger).id);
+					if (inputEl) {
+						inputEl.focus();
```
TGlide, line 109: You should probably only focus when it's being opened, I'd gather 🤔

huntabyte: ✅

huntabyte:
> We should probably make it so clicking on the input does not close combobox, as part of this.

Alright so I've added that part (along with tests), but if you notice in the preview I'm not able to move the cursor by clicking the various positionings in the input. I'm not seeing any preventDefaults being called so one would think it would work like a normal input. Am I missing something here? @TGlide

mustafa0x:
> I'm not able to move the cursor by clicking the various positionings in the input.

related issue (not specific to this pr): if the combobox is open, it required two clicks to focus another input. the first closes the combobox, but keeps the input focused.

i don't think this behavior is correct. compare with https://react-spectrum.adobe.com/react-aria/examples/user-combobox.html

anatolzak:
> related issue (not specific to this pr): if the combobox is open, it required two clicks to focus another input. the first closes the combobox, but keeps the input focused.

@mustafa0x pretty sure https://github.com/melt-ui/melt-ui/pull/1114 fixed this. Can you verify if this still happens?

mustafa0x:
Still seems to happen


https://github.com/melt-ui/melt-ui/assets/1473102/1df129de-16a4-4e63-8148-bf2818aa963b

anatolzak:
> Still seems to happen (sorry for the derail)

@mustafa0x it was actually fixed in the mentioned PR but it hasn't been released yet and the main website hasn't been updated. When testing this locally, I don't experience the issue. I have also [added tests](https://github.com/melt-ui/melt-ui/pull/1148) to prevent regressing the issue you described and they tests are passing. 

https://github.com/melt-ui/melt-ui/assets/53095479/2255cab4-0f4b-4c0b-8bc1-3dcf6e7c2c78

lumosminima:
👀

TGlide:
Closing, will add in Melt for Runes

PR #1042, opened 2024-03-03
fix: data-state consistency and floating-ui styles

### LINKED ISSUE #1041: Data State Inconsistency
### Describe the bug

There is currently an inconsistency in how we calculate the `data-state` attribute.

With Dialog Content and Overlay, Tooltip Content:
```typescript
'data-state': $open ? 'open' : 'closed'
```

With the rest of the components (Link Preview Content, Menu Content, Popover Content, etc.):
```typescript
'data-state': $isVisible ? 'open' : 'closed'
```

I believe that the `data-state` should be based on `$open`, because if we use `forceVisible: true` and always keep the content mounted, `data-state` will always be `open`.

The only place where this decision between `$open` and `$isVisible` matters, is when the user uses `forceVisible: true`, in which case if they unmount the component when `$open` is `false` and use a svelte transition, `data-state` will actually stay `open` because of how svelte internally freezes the component to transition it out. But if the user always leaves it as mounted and just changes the content's visibility, `data-state` should be `closed`, which is not currently the case with the codebase.

Now, if we agree that we should change the `data-state` based on `$open`, we need to make sure that we don't override styles set by floating-ui in certain components.

```typescript
'data-state': $open ? 'open' : 'closed',
style: styleToString({
}),
```

The above code would "rerender" every time the `$open` state changes, and if the user never unmounts the content, the styles will get reset to an empty string, thereby removing the styles set by floating-ui, which causes the component to jump. This was already fixed in the case of the tooltip content here https://github.com/melt-ui/melt-ui/issues/1010

The solution was to use the below code
```typescript
'data-state': $open ? 'open' : 'closed',
style: $isVisible ? undefined : styleToString({ display: 'none' }),
```
... and to remove properties with undefined values returned from the tooltip content, because if we set style to undefined, svelte will also reset the styles set by floating-ui. So we needed to remove the style property altogether.

So since this will be a problem in all builders, we should remove properties with undefined values in the `makeElement` function.

### Reproduction

-

### Logs

_No response_

### System Info

```bash
-
```

### Severity

annoyance

anatolzak:
closes #1041

There is currently an inconsistency in how we calculate the `data-state` attribute.

With Dialog Content and Overlay, Tooltip Content:
```typescript
'data-state': $open ? 'open' : 'closed'
```

With the rest of the components (Link Preview Content, Menu Content, Popover Content, etc.):
```typescript
'data-state': $isVisible ? 'open' : 'closed'
```

I believe that the `data-state` should be based on `$open`, because if we use `forceVisible: true` and always keep the content mounted, `data-state` will always be `open`.

The only place where this decision between `$open` and `$isVisible` matters, is when the user uses `forceVisible: true`, in which case if they unmount the component when `$open` is `false` and use a svelte transition, `data-state` will actually stay `open` because of how svelte internally freezes the component to transition it out. But if the user always leaves it as mounted and just changes the content's visibility, `data-state` should be `closed`, which is not currently the case with the codebase.

Now, if we agree that we should change the `data-state` based on `$open`, we need to make sure that we don't override styles set by floating-ui in certain components.

```typescript
'data-state': $open ? 'open' : 'closed',
style: styleToString({
}),
```

The above code would "rerender" every time the `$open` state changes, and if the user never unmounts the content, the styles will get reset to an empty string, thereby removing the styles set by floating-ui, which causes the component to jump. This was already fixed in the case of the tooltip content here https://github.com/melt-ui/melt-ui/issues/1010

The solution was to use the below code
```typescript
'data-state': $open ? 'open' : 'closed',
style: $isVisible ? undefined : styleToString({ display: 'none' }),
```
... and to remove properties with undefined values returned from the tooltip content, because if we set style to undefined, svelte will also reset the styles set by floating-ui. So we needed to remove the style property altogether.

So since this will be a problem in all builders, we should remove properties with undefined values in the `makeElement` function.


anatolzak:
In the link preview builder, we used to set the following properties on the content when `$isVisible` is false.
`hidden: true, style: { opacity: 0, pointer-events: none }`

I replaced `opacity: 0, pointer-events: none` with `display: none` because when we set `hidden: true`, the element is already removed from document flow, so there is no reason to set the opacity and pointer events, and can instead just set `display: none`.

Also, we always set `userSelect: none` on the link preview content. It seems like it's unnecessary since we only need to worry about that during user selection which is already handled by

```typescript
effect([containSelection], () => {
    ...
    contentElement.style.userSelect = 'text';
    contentElement.style.webkitUserSelect = 'text';
})
```

The reason I removed all these styles and only kept `style: $isVisible ? undefined : styleToString({ display: 'none' })` is because if we set any styles when the `$isVisible` is true and when `forceVisible` is `false`, we will rerender the styles every time `$open` changes and thus we reset the styles set by floating-ui which was causing the following issue 

https://github.com/melt-ui/melt-ui/assets/53095479/709af6a4-81a5-49c9-9a17-8e2214be9538

TGlide (approved):

TGlide:
Hey! Thanks again for all your work. if you could resolve conflicts, should be good to go!

anatolzak:
Hey @TGlide! I resolved the conflicts. Thanks so much for your time! :)

PR #1049, opened 2024-03-04
fix(tooltip, popover, listbox, link-preview) incorrect arrow placement provided forceVisible false

### LINKED ISSUE #1048: [Tooltip, Popover, Listbox] Incorrect Arrow Placement provided forceVisible false
### Describe the bug

For the following components, tooltip, popover, and listbox, when used with `forceVisible: false`, the arrow position is wrong.

### Tooltip
https://github.com/melt-ui/melt-ui/assets/53095479/4bb9364c-ef62-4586-8dc2-8b664d92a2c5

### Listbox (Combobox, Select etc.)
https://github.com/melt-ui/melt-ui/assets/53095479/c8efd5f6-4e9c-4603-95c6-b71527b55a1d

### Popover
https://github.com/melt-ui/melt-ui/assets/53095479/6652fa7c-34ee-4cf1-ad63-623463b7b793

### Reproduction

https://github.com/anatolzak/melt-ui-arrow-bug
https://melt-ui-arrow-bug.vercel.app/

### Logs

_No response_

### System Info

```bash
System:
    OS: macOS 14.3
    Memory: 243.90 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    bun: 1.0.22 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 122.1.63.161
    Chrome: 120.0.6099.109
    Edge: 114.0.1823.82
    Safari: 17.3
  npmPackages:
    @melt-ui/svelte: ^0.75.2 => 0.75.2 
    svelte: ^4.2.7 => 4.2.12
```

### Severity

annoyance

anatolzak:
closes #1048

For the following components, tooltip, popover, and listbox, when used with `forceVisible: false`, the arrow position is wrong.

This is easily fixed by adding `tick()`. 

The reason for this is because when `forceVisible` is `false` and `$open` is `false`, we set `display: none` on the content, and so in that moment when `$open` gets switched to `true`, the content's client rect is `{ top: 0, left: 0, ... }`, so the arrow offset ends up being `0`, which is why it screws the visual position it's supposed to have. So we just need to wait for svelte to rerender and remove `display: none` at which point the content has valid client rect and the arrow offset can be derived from it.

### Tooltip Before

https://github.com/melt-ui/melt-ui/assets/53095479/d3f0d0d3-0772-4fd9-bad8-3f376514185d

### Tooltip After

https://github.com/melt-ui/melt-ui/assets/53095479/742f4b8e-2304-4e86-b327-08dc7be33c10

### Popover Before

https://github.com/melt-ui/melt-ui/assets/53095479/6bacbdad-60e5-4654-b1ff-466298f87aac

### Popover After

https://github.com/melt-ui/melt-ui/assets/53095479/82905e4e-0fd1-477e-88c5-7f842de69a0c

### Listbox Before (Combobox, Select, etc.)

https://github.com/melt-ui/melt-ui/assets/53095479/932951b8-b313-4331-b709-90a70a545e83

### Listbox After

https://github.com/melt-ui/melt-ui/assets/53095479/8ec0fe7e-778a-4ab9-82d0-8690be00a91f

### Link Preview Before
Technically it works right now because we don't set `display: none` on the content. However we set `hidden: true, style: { opacity: 0, pointer-events: none }`.  If we set `hidden: true`, the element is removed from document flow. So we might as well set `display: none` instead of `opacity: 1, pointer-events: none`. So if we set `display: none`, the same issue occurs. I am going to create a separate PR for `display: none` on link preview content

https://github.com/melt-ui/melt-ui/assets/53095479/166ff006-7698-405d-b733-966aa7fc1a85

### Link Preview After

https://github.com/melt-ui/melt-ui/assets/53095479/ee56b96e-2de3-4653-8ed4-10ea3d64bb14


anatolzak:
Also to address the arrow positioning issue in the tooltip content builder, I adjusted the function invocation sequence. Previously, `useFloating` was called before `usePortal`, but reversing this order—calling `usePortal followed by `useFloating`—resolved the problem.

anatolzak:
@TGlide when you can, do you mind giving this a quick look? Currently, arrows don't render correctly when using `forceVisible: false`. The solution is to use `tick()` which is already used to solve this issue in multiple other components. Thanks!

PR #1050, opened 2024-03-04
fix: listbox arrowSize jsdoc default

anatolzak:
In the listbox builder, there is no default value for `arrowSize`. However the JSDoc claims the default is 8. 
I set the default in the JSDoc to `undefined`.

image


huntabyte:
I want to say this gets set by the default "popper"/floating configuration. Is this not true?

anatolzak:
> I want to say this gets set by the default "popper"/floating configuration. Is this not true?

@huntabyte it doesn't. The popper and floating don't take any `arrowSize` option. floating ui relies on the arrow dom element for its calculations.

huntabyte (approved):
Thank you!

PR #1051, opened 2024-03-04
refactor dialog

anatolzak:
- setting dialog `activeTrigger` in trigger action instead of in `handleOpen`


huntabyte:
This change isn't necessary and will impact how we handle multiple ways of triggering a dialog while restoring focus to the element that triggered it.

PR #1052, opened 2024-03-04
refactor listbox

anatolzak:
Listbox Refactor:
- setting `activeTrigger` in trigger action instead of in `openMenu` and in `safeOnMount` and in `effect([open])`
- using early return in `effect([open])` for cleaner code

### DIFF
--- a/src/lib/builders/listbox/create.ts
+++ b/src/lib/builders/listbox/create.ts
@@ -39,7 +39,6 @@ import {
 	toggle,
 	withGet,
 } from '$lib/internal/helpers/index.js';
-import { safeOnMount } from '$lib/internal/helpers/lifecycle.js';
 import type { Defaults, MeltActionReturn } from '$lib/internal/types.js';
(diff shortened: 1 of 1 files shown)


huntabyte:
Same comments as #1055

src/lib/builders/listbox/create.ts
```diff
@@ -292,6 +285,7 @@ export function createListbox<
 			} as const;
 		},
 		action: (node: HTMLElement): MeltActionReturn<ListboxEvents['trigger']> => {
+			activeTrigger.set(node);
```
TGlide, line 288: It should reset on destroy

TGlide (commented):
Thank you, just one small change needed!

src/lib/builders/listbox/create.ts
anatolzak: Done!

TGlide (approved):

PR #1054, opened 2024-03-04
refactor context menu

anatolzak:
Context Menu Refactor:
- using `isElement` helper
- setting `rootActiveTrigger` in trigger action instead of in `handleOpen`
- using nullish coalescing operator


huntabyte:
This PR will get similar feedback as #1055 - the `activeTrigger` needs to remain composable. By setting it in the action we aren't able to use an element not using the action as an active trigger.

TGlide (approved):

TGlide:
> This PR will get similar feedback as #1055 - the `activeTrigger` needs to remain composable. By setting it in the action we aren't able to use an element not using the action as an active trigger.

Why not? If that's the goal, then we need to expose the activeTrigger store, which we currently do not. And even then, it shouldn't be an impediment, right?

If you agree, feel free to merge this later :)

TGlide (approved):

PR #1055, opened 2024-03-04
refactor popover

anatolzak:
- settings popover `activeTrigger` in trigger action instead of in `safeOnMount` and in `toggleOpen` and in `effect([open, activeTrigger, preventScroll])`
- using early return in `effect([open, activeTrigger, preventScroll])` for cleaner code

### DIFF
--- a/src/lib/builders/popover/create.ts
+++ b/src/lib/builders/popover/create.ts
@@ -18,6 +18,7 @@ import {
 	toWritableStores,
 	portalAttr,
 	generateIds,
+	withGet,
 } from '$lib/internal/helpers/index.js';
 
 import {
@@ -77,7 +78,7 @@ export function createPopover(args?: CreatePopoverProps) {
 	const openWritable = withDefaults.open ?? writable(withDefaults.defaultOpen);
 	const open = overridable(openWritable, withDefaults?.onOpenChange);
 
-	const activeTrigger = writable<HTMLElement | null>(null);
+	const activeTrigger = withGet.writable<HTMLElement | null>(null);
 
 	const ids = toWritableStores({ ...generateIds(popoverIdParts), ...withDefaults.ids });
 
@@ -139,7 +140,7 @@ export function createPopover(args?: CreatePopoverProps) {
 									: {
 											returnFocusOnDeactivate: false,
 											clickOutsideDeactivates: true,
-											escapeDeactivates: true,
+											escapeDeactivates: $closeOnEscape,
 									  },
 								modal: {
 									shouldCloseOnInteractOutside: shouldCloseOnInteractOutside,
--- a/src/lib/builders/popover/create.ts
+++ b/src/lib/builders/popover/create.ts
@@ -16,7 +16,6 @@ import {
 	removeScroll,
 	styleToString,
 	toWritableStores,
-	sleep,
 	portalAttr,
 	generateIds,
 } from '$lib/internal/helpers/index.js';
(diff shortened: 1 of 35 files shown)


src/lib/builders/popover/create.ts
```diff
@@ -213,14 +202,13 @@ export function createPopover(args?: CreatePopoverProps) {
 			} as const;
 		},
 		action: (node: HTMLElement): MeltActionReturn<PopoverEvents['trigger']> => {
+			activeTrigger.set(node);
```
huntabyte, line 205: While this seems like a nice approach, it's limiting. Currently, we can use any arbitrary element as the `activeTrigger` using the `id`, this now turns using the builder into a requirement which is unideal for composition.

src/lib/builders/popover/create.ts
```diff
@@ -316,26 +304,14 @@ export function createPopover(args?: CreatePopoverProps) {
...
-			if ($preventScroll) {
-				unsubs.push(removeScroll());
-			}
-
-			const triggerEl = $activeTrigger ?? document.getElementById(ids.trigger.get());
-			handleFocus({ prop: openFocus.get(), defaultEl: triggerEl });
+		if ($preventScroll) {
+			unsubs.push(removeScroll());
 		}
+		handleFocus({ prop: openFocus.get(), defaultEl: $activeTrigger });
```
huntabyte, line 314: We "re-get" the trigger because it could change throughout use of the popover (via the ID prop mentioned in my previous comment)

huntabyte (commented):
Is this PR solving any actual issues or just refactoring? Most of the things exist for a reason so it's best to have an issue discussing why these changes are required so we can also consider the tradeoffs / implications of them.

src/lib/builders/popover/create.ts
```diff
@@ -180,13 +174,8 @@ export function createPopover(args?: CreatePopoverProps) {
 		},
 	});
 
-	function toggleOpen(triggerEl?: HTMLElement) {
-		open.update((prev) => {
-			return !prev;
-		});
-		if (triggerEl) {
-			activeTrigger.set(triggerEl);
```
anatolzak, line 188: removing this line, fixed the tests in https://github.com/melt-ui/melt-ui/pull/1080
for some reason, in a recent PR https://github.com/melt-ui/melt-ui/pull/1049, we added `tick()` to solve the issue mentioned in the PR. But for some reason, that started failing the tests in https://github.com/melt-ui/melt-ui/pull/1080 and by removing the highlighted line of code, the tests started passing.

src/lib/builders/popover/create.ts
anatolzak: If the trigger changed throughout the use of the popover, the presupposition is that there was a trigger previously so `activeTrigger` is truthy. By that token we would never reach `document.getElementById`. Am I wrong?

src/lib/builders/popover/create.ts
anatolzak: I feel like letting people use the library without using the trigger builder is weird. The point of using melt is to have all the accessibility and functionality taken care of by the library. If you choose to not use the trigger builder, you are not going to have all the correct aria attributes, etc. so it seems awkward to be building the codebase around that use-case. of course, I may be totally wrong

anatolzak: Letting people use any arbitrary element as the trigger without using our builder introduces so many edge cases. We always want to make sure we have the up-to-date reference to the trigger for the floating positioning. 

If we require using our trigger builder, we easily set the active trigger on trigger mount and set it to null on destroy. However, if we allow any arbitrary element to be used, how do we know if the element is still in the DOM? What if the trigger was unmounted and remounted. This will get very ugly very quickly if we want to actually fully support that and fix all the issues that will come with it.

TGlide: I agree and disagree with both of you 😄 

I agree with @anatolzak in the sense that making the id the implicit way of using it is weird. Two ways of doing things and all that.

However, Melt was indeed made to be super flexible. It provides accessible defaults, but allows you to step out of the reigns here and there. So if we want to support this, which I agree with @huntabyte that we should, we should just expose the `activeTrigger` store and document it, say it's dangerous, etc. This way its the users responsibility that things are breaking.

TGlide (approved):

Issue #1039, opened 2024-03-03
[Link Preview, Menu, Tooltip] Interact Outside Transition Out Issue During Component Unmount


anatolzak:
### Describe the bug

This was previously fixed in https://github.com/melt-ui/melt-ui/pull/1006 using `sleep(0)`, however this seems to have been removed during a refactor since then in the link preview and menu builders.

As far as the tooltip, it seems like `sleep(0)` does not suffice and from experimenting, it seems like we have to do `sleep(20)` in the tooltip content blur handler before calling `closeTooltip()` to prevent this issue.

### Menu
https://github.com/melt-ui/melt-ui/assets/53095479/1bf2370b-483c-4feb-b1d0-b5b52559dde3

### Link Preview
https://github.com/melt-ui/melt-ui/assets/53095479/cb4815ba-bcae-4cb5-ab7e-37db1fa440ea

### Tooltip
https://github.com/melt-ui/melt-ui/assets/53095479/a1ee3f6c-badf-4fac-8049-8a90ea73249c

### Reproduction

https://melt-interact-outside-out-transition-bug.vercel.app/
https://github.com/anatolzak/melt-interact-outside-out-transition-bug

### System Info

```bash
System:
    OS: macOS 14.3
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 420.84 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v20.9.0/bin/yarn
    npm: 10.1.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
    pnpm: 8.11.0 - /usr/local/bin/pnpm
    bun: 1.0.22 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 122.1.63.161
    Chrome: 120.0.6099.109
    Edge: 114.0.1823.82
    Safari: 17.3
  npmPackages:
    @melt-ui/svelte: ^0.74.4 => 0.74.4 
    @sveltejs/kit: ^2.0.0 => 2.5.1 
    svelte: ^4.2.7 => 4.2.12
```

### Severity

annoyance

anatolzak:
This was solved by https://github.com/melt-ui/melt-ui/pull/1059

PR #1059, opened 2024-03-05
fix: clean dead reference to floating content

### LINKED ISSUE #1058: [Tooltip] content jumps to top left of page during unmounting when using out transition and else if block to mount content
### Describe the bug

https://github.com/melt-ui/melt-ui/assets/53095479/a1ee3f6c-badf-4fac-8049-8a90ea73249c

This has been an ongoing issue with multiple components--the content jumps to the top left of the page when you unmount the component externally while the content is open when using an out transition on the content and when using an else if block to render the content.

The workaround that solved this issue for all components but the tooltip was to introduce `sleep(0)` before setting the `$open` state to `false`. So when the component get unmounted externally while the content is open, svelte can unmount the whole component instead of, for some reason, starting the transition out of the content, at which stage the content loses its anchor and causes the content to jump to the top left of the page.

However, for some reason, adding `sleep(0)` to the tooltip as a workaround does not work. It requires adding a pretty significant delay (~80ms) to prevent this issue. However adding a significant delay is not preferable.

### Reproduction

https://melt-interact-outside-out-transition-bug.vercel.app/tooltip

https://github.com/anatolzak/melt-interact-outside-out-transition-bug/blob/main/src/routes/tooltip/%2Bpage.svelte

### Logs

_No response_

### System Info

```bash
System:
    OS: macOS 14.3
    Memory: 243.90 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    bun: 1.0.22 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 122.1.63.161
    Chrome: 120.0.6099.109
    Edge: 114.0.1823.82
    Safari: 17.3
  npmPackages:
    @melt-ui/svelte: ^0.75.2 => 0.75.2 
    svelte: ^4.2.7 => 4.2.12
```

### Severity

annoyance

anatolzak:
closes #1058, #1039

https://github.com/melt-ui/melt-ui/assets/53095479/a1ee3f6c-badf-4fac-8049-8a90ea73249c

This has been an ongoing issue with multiple components--the content jumps to the top left of the page when you unmount the component externally while the content is open when using an out transition on the content and when using an else if block to render the content.

The previous workaround that solved this issue for all components but the tooltip was to introduce `sleep(0)` before setting the `$open` state to `false`. So when the component get unmounted externally while the content is open, svelte can unmount the whole component instead of, for some reason, starting the transition out of the content, at which stage the content loses its anchor and causes the content to jump to the top left of the page.

However, for some reason, adding `sleep(0)` to the tooltip as a workaround does not work. It requires adding a pretty significant delay (~80ms) to prevent this issue. However adding a significant delay is not preferable.

I dug deeper into this issue and found out that the reason the content jumps to the top left of the page is because in `useFloating`, the `compute` function reruns to set the new position of the content as it is transitioning out. 

I found out that the client rect of the reference (the anchor) is `{ top: 0, left: 0 }` during the unmounting of the content. The reason for that is that we are holding a reference to the trigger which was already unmounted, but in JavaScript, you can still hold a reference to an html element that is not in the document anymore. An html element that is not the document has client rect of `0`s.

So our code that recomputes the position of the content is rerun with a trigger that is "at the top left of the page", but in reality it's no longer in the document. 

So we should check if the reference is in the document before computing a new position for the content.

And so this finally solves this issue without having to add `delay(0)` to so many builders in addition to the delay not working in the case of the tooltip.

Finally :)

### DIFF
--- a/.changeset/strange-actors-guess.md
+++ b/.changeset/strange-actors-guess.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+fix(Tooltip, Link Preview, Menu, Popover, Listbox) fixed bug where content jumps to top left of page during external unmounting when using out transition on the content and else if block to render the content (closes #1058, #1039)
--- a/src/lib/builders/listbox/create.ts
+++ b/src/lib/builders/listbox/create.ts
@@ -32,7 +32,6 @@ import {
 	prev,
 	removeHighlight,
 	removeScroll,
-	sleep,
 	stripValues,
 	styleToString,
 	toWritableStores,
@@ -227,8 +226,7 @@ export function createListbox<
 	}
 
 	/** Closes the menu & clears the active trigger */
-	async function closeMenu() {
-		await sleep(0);
+	function closeMenu() {
 		open.set(false);
 		highlightedItem.set(null);
 	}
--- a/src/lib/builders/popover/create.ts
+++ b/src/lib/builders/popover/create.ts
@@ -17,7 +17,6 @@ import {
 	removeScroll,
 	styleToString,
 	toWritableStores,
-	sleep,
 	portalAttr,
 	generateIds,
 } from '$lib/internal/helpers/index.js';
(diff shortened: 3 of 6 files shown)


anatolzak:
For some reason a test in Combobox wasn't passing after removing `sleep(0)` from `closeMenu` in `listbox/create.ts`, even though I just recently added the delay in a PR. I fixed the tests in question

huntabyte:
This makes complete sense!

Issue #1062, opened 2024-03-05
[Table of Contens] Browser Back Forward Scroll


anatolzak:
### Describe the bug

When using native html elements with ids and clicking on multiple anchor tags with a hash in the current page, when clicking the back and forward navigation buttons in the browser, the page scrolls back to that element.

When using melt-ui's table of contents and clicking the back and forward navigation buttons in the browser, the page does not scrolls back to the correct element.

### Here is a video showing native behavior:

https://github.com/melt-ui/melt-ui/assets/53095479/6242eab2-3e9e-41c3-b65b-a48a20d740a7

### And here is Melt-UI Table of Contents

https://github.com/melt-ui/melt-ui/assets/53095479/104617f7-76d0-4af0-8958-dc99bd6b706f

Also, I am getting the following warning in the console
Avoid using `history.pushState(...)` and `history.replaceState(...)` as these will conflict with SvelteKit's router. Use the `pushState` and `replaceState` imports from `$app/navigation` instead.

### Reproduction

-

### System Info

```bash
System:
    OS: macOS 14.3
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 46.34 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v20.9.0/bin/yarn
    npm: 10.1.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
    pnpm: 8.11.0 - /usr/local/bin/pnpm
    bun: 1.0.22 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 122.1.63.161
    Chrome: 120.0.6099.109
    Edge: 114.0.1823.82
    Safari: 17.3
  npmPackages:
    @melt-ui/svelte: ^0.75.2 => 0.75.2 
    @sveltejs/kit: ^2.0.0 => 2.5.2 
    svelte: ^4.2.7 => 4.2.12
```

### Severity

annoyance

PR #1063, opened 2024-03-05
fix(table of contents) browser back forward scroll

anatolzak:
closes https://github.com/melt-ui/melt-ui/issues/1062

When using native html elements with ids and clicking on multiple anchor tags with a hash in the current page, when clicking the back and forward navigation buttons in the browser, the page scrolls back to that element.

As of right now, when using melt-ui's table of contents and clicking the back and forward navigation buttons in the browser, the page does not scrolls back to the correct element.

This PR fixes this by using `pushState` from `$app/navigation` instead of `history.pushState(...)`, which is recommended from the warning I was previously getting in the console with our current implementation.

> Avoid using history.pushState(...) and history.replaceState(...) as these will conflict with SvelteKit's router. Use the pushState and replaceState imports from $app/navigation instead.

### Before

https://github.com/melt-ui/melt-ui/assets/53095479/8c709046-aece-4c84-b423-87137525d0be

### After

https://github.com/melt-ui/melt-ui/assets/53095479/b3336e85-27bf-4023-bd4d-84d7011657f3

### DIFF
--- a/src/lib/builders/table-of-contents/create.ts
+++ b/src/lib/builders/table-of-contents/create.ts
@@ -18,6 +18,7 @@ import type {
 	HeadingParentsLU,
 	TableOfContentsItem,
 } from './types.js';
+import { pushState } from '$app/navigation';
 
 const defaults = {
 	exclude: ['h1'],
@@ -373,7 +374,7 @@ export function createTableOfContents(args: CreateTableOfContentsArgs) {
 
 					// Add items hash to URL
 					if (id) {
-						history.pushState({}, '', `#${id}`);
+						pushState(`#${id}`, {});
 					}
 				})
 			);
--- a/.changeset/curvy-gorillas-talk.md
+++ b/.changeset/curvy-gorillas-talk.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+Fixed bug in Table of Contents where navigating with the browser's back and forward buttons failed to automatically scroll to the element specified by the `location.hash`. (closes #1062)


src/lib/builders/table-of-contents/create.ts
```diff
@@ -373,7 +374,7 @@ export function createTableOfContents(args: CreateTableOfContentsArgs) {
 
 					// Add items hash to URL
 					if (id) {
-						history.pushState({}, '', `#${id}`);
+						pushState(`#${id}`, {});
```
huntabyte, line 377: We avoid using Svelte**Kit**-specific APIs to support non-kit projects.

Perhaps we keep the default at `history.pushState({}, '', `#${id}`)` and expose a prop that receives the `id` as an arg that the user could override if they wanted to.

Something like:

```ts
const defaults = {
	// ...
	pushState: (id) => {
			history.pushState({}, '', `#${id}`
	}
}
```

and then in the click handler:

```ts

if (id) {
	options.pushState.get()(id)
}
```

Or something along those lines. Wdyt @TGlide? Prop name subject to bikeshedding of course (can't think of a nice name atm), but the overall idea.

huntabyte (changes requested):
Let's discuss this

src/lib/builders/table-of-contents/create.ts
anatolzak: We should provide a default option as @huntabyte suggested, however I would argue that we should mark that option as required. If it's optional, no one is going to provide the svelte-kit specific pushState which is more correct and also prevents the issue for back and forward navigation. 

I assume most people are using melt with svelte-kit, so marking the option as optional, will guarantee that most people will still encounter #1062. 

Of course marking the option as required seems undesired, however it's a difficult challenge to build a library that is mostly used by svelte-kit users while the library tries to not use svelte-kit specific API's which then breaks the apps of most users using this library.

AdrianGonz97: I don't think this is good idea. Unless the intention is to become a SK only library, we shouldn't be using/importing any SK specific dependencies, even if we mark them as optional, and _especially_ if they're Vite virtual modules. The bundler will certainly throw a fit when it runs into `await import('$app/navigation')` when it's used in vanilla Vite project and will probably require extra plumbing to not error out for non-SK users.

TGlide: Well this sucks.

Would it be possible to modify the url, and navigate to the element, instead of using pushState?

anatolzak: > Would it be possible to modify the url, and navigate to the element, instead of using pushState?

@TGlide I am pretty sure doing that was breaking the smooth scroll https://github.com/melt-ui/melt-ui/pull/860 if using `window.location.href = ...` instead of `pushState`

huntabyte: You're thinking putting this into a `runAfterTransition` @TGlide?

TGlide: @huntabyte exactly

PR #1064, opened 2024-03-05
fix interfact outside interaction in different portal

### LINKED ISSUE #1036: Tooltip Loses Anchor and Stays Open When Clicked inside Popover
### Describe the bug

If you mount a tooltip inside a popover, and then click on the tooltip content, the popover gets unmounts and the tooltip content jumps to the top left of the page (because of losing its anchor).

https://github.com/melt-ui/melt-ui/assets/53095479/6f464c03-c62d-4d40-88d6-e65ec61ddfd7

The reason this happens is because when the popover is open, we are listening for outside clicks to close the popover. However if you have a tooltip mounted inside a popover, and you click inside the tooltip content, you would expect nothing to happen.

I tried adding a listener for `pointerup` on the tooltip content and then calling `e.stopPropagation()` to prevent the popover `pointerup` from running. 

This didn't work since the popover listens for `pointerup` on the document with `capture: true`.

And from what I understand, there is no way to have the tooltip content add another `pointerup` listener that will be able to run before the popover one and be able to stop propagation since we are using `capture: true`.

I also tried adding `sleep(0)` before running the `pointerup` event of the popover and then calling `e.preventDefault` in the event added by the tooltip content, however this breaks some behavior of being able to select text in the tooltip content.

### Reproduction

https://github.com/anatolzak/melt-ui-popover-tooltip-interaction-bug
https://melt-ui-popover-tooltip-interaction-bug.vercel.app/

### Logs

_No response_

### System Info

```bash
System:
    OS: macOS 14.3
    Memory: 374.56 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    bun: 1.0.22 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 122.1.63.161
    Chrome: 120.0.6099.109
    Edge: 114.0.1823.82
    Safari: 17.3
  npmPackages:
    @melt-ui/svelte: ^0.74.4 => 0.74.4 
    svelte: ^4.2.7 => 4.2.12
```

### Severity

annoyance

anatolzak:
closes #1036

as @huntabyte said, in interact outside logic, we should ignore events coming from a different portal

### Before

https://github.com/melt-ui/melt-ui/assets/53095479/f6953850-1fe4-4e5f-af73-daabfb550a99

### After

https://github.com/melt-ui/melt-ui/assets/53095479/65056a17-402b-416e-bcad-3e600b6de40d

### DIFF
--- a/.changeset/warm-bees-tell.md
+++ b/.changeset/warm-bees-tell.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+Fixed Outside Interaction Handling Across Portals: Modified the interact outside logic to ignore events originating from different portals. (closes #1036)
--- a/src/lib/internal/actions/interact-outside/action.ts
+++ b/src/lib/internal/actions/interact-outside/action.ts
@@ -109,7 +109,13 @@ function isValidEvent(e: InteractOutsideEvent, node: HTMLElement): boolean {
 		return false;
 	}
 
-	return node && !isOrContainsTarget(node, target);
+	return node && !isOrContainsTarget(node, target) && !isEventInsideDifferentPortal(e, node);
+}
+
+function isEventInsideDifferentPortal(e: InteractOutsideEvent, node: HTMLElement): boolean {
+	const targetPortal = isElement(e.target) && e.target.closest('[data-portal]');
+	const nodePortal = node.closest('[data-portal]');
+	return !!targetPortal && targetPortal !== nodePortal;
 }
 
 function isOrContainsTarget(node: HTMLElement, target: Element) {


huntabyte:
The tooltip shouldn't be in a different portal unless specified. `portal` should be set to `null` automatically if it finds an existing portal ancestor. Is the tooltip content becoming a sibling of the popover content?

anatolzak:
> Is the tooltip content becoming a sibling of the popover content?

@huntabyte yes, the tooltip is mounted in a different portal than the popover

huntabyte:
Okay that's an issue that needs to be addressed then and will remove the need for this PR to exist.

anatolzak:
@huntabyte so you will never be able to have a popover with a tooltip inside that is mounted in a different portal?

huntabyte:
> @huntabyte so you will never be able to have a popover with a tooltip inside that is mounted in a different portal?

What would the use-case for this be? If you had a hard requirement for this you could prevent default in `onOutsideClick` handler. By default if a "portalled" element is a descendant of another "portalled" element, the portal will automatically be set to `null` so it remains a descendant and functions as you would expect (from a DOM structure perspective).

huntabyte:
Closing this as the issue was identified and is resolved in #1066

anatolzak:
> @huntabyte so you will never be able to have a popover with a tooltip inside that is mounted in a different portal?

> What would the use-case for this be?

@huntabyte One of the reasons for using portals in the first place--preventing clipping problems when using `overflow: hidden`. If for some reason you need to set `overflow: hidden` on the popover content, the tooltip gets clipped.

image

anatolzak:
> If you had a hard requirement for this you could prevent default in `onOutsideClick` handler.

@huntabyte This would break a lot of behavior. 
1. Clicking outside the popover wouldn't close it
2. Clicking inside the tooltip will not be possible

https://github.com/melt-ui/melt-ui/assets/53095479/a18e29c4-c1b5-4d5f-b351-754c66c8ac02

Issue #725, opened 2023-11-09
🐛Bug: Ordering of days of week when updating locale


huntabyte:
When the locale is updated, if `weekStartsOn` is not set by the user, should update to whatever is standard for that locale.

PR #1070, opened 2024-03-06
fix: months not updating when changing options

### LINKED ISSUE #1069: [Calendar & related] Updating options doesn't update values
### Describe the bug

When you try to update some value of some option that would change the calendar (specifically `numberOfMonths`, `fixedWeeks`, `weekStartsOn`) the calendar is not updated until the month or the date is changed.

I've already investigated this and it all boils down to 

https://github.com/melt-ui/melt-ui/blob/746ee7c65dea907f64af3314d51d7096fa0c0b7f/src/lib/builders/calendar/create.ts#L524-L530

the effect is responsible to recalculate the months but it bails out if the current placeholder is already in the visibleMonths. Removing this solves the issue but i suspect will do a bit more work in general. I think this check should be in place only if the placeholder is the only thing that changed.

I can also work on this and open a PR, just wanted a bit more insight on this if (wink wink @huntabyte )

### Reproduction

To test the fixed weeks not working obviously set to a month were fixed would work

https://www.sveltelab.dev/44fmd3k81k63jen

### Logs

_No response_

### System Info

```bash
  System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.3 - /usr/local/bin/pnpm
  npmPackages:
    @melt-ui/svelte: latest => 0.75.3 
    svelte: ^4.1.2 => 4.2.12 
```

### Severity

blocking an upgrade

paoloricciuti:
Closes #1069

### DIFF
--- a/src/lib/builders/calendar/create.ts
+++ b/src/lib/builders/calendar/create.ts
@@ -514,7 +514,7 @@ export function createCalendar<
 	 * Updates the displayed months based on changes in the placeholder value,
 	 * which determines the months to show in the calendar.
 	 */
-	effect([placeholder, weekStartsOn, locale, fixedWeeks, numberOfMonths], ([$placeholder]) => {
+	effect([placeholder], ([$placeholder]) => {
 		if (!isBrowser || !$placeholder) return;
 
 		const $visibleMonths = visibleMonths.get();
--- a/src/lib/builders/calendar/create.ts
+++ b/src/lib/builders/calendar/create.ts
@@ -514,20 +514,48 @@ export function createCalendar<
 	 * Updates the displayed months based on changes in the placeholder value,
 	 * which determines the months to show in the calendar.
 	 */
+	effect([placeholder, weekStartsOn, locale, fixedWeeks, numberOfMonths], ([$placeholder]) => {
+		if (!isBrowser || !$placeholder) return;
+
+		const $visibleMonths = visibleMonths.get();
+
+		/**
+		 * If the placeholder's month is already in the visible months,
+		 * we don't need to do anything.
+		 */
+		if ($visibleMonths.some((month) => isSameMonth(month, $placeholder))) {
+			return;
+		}
+
+		const $weekStartsOn = weekStartsOn.get();
+		const $locale = locale.get();
+		const $fixedWeeks = fixedWeeks.get();
+		const $numberOfMonths = numberOfMonths.get();
+
+		const defaultMonthProps = {
+			weekStartsOn: $weekStartsOn,
+			locale: $locale,
+			fixedWeeks: $fixedWeeks,
+			numberOfMonths: $numberOfMonths,
+		};
+
+		months.set(
+			createMonths({
+				...defaultMonthProps,
+				dateObj: $placeholder,
+			})
+		);
+	});
+
+	/**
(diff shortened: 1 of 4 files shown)


huntabyte:
Also closes #725

src/lib/builders/calendar/create.ts
```diff
@@ -514,20 +514,48 @@ export function createCalendar<
 	 * Updates the displayed months based on changes in the placeholder value,
 	 * which determines the months to show in the calendar.
 	 */
-	effect(
-		[placeholder, weekStartsOn, locale, fixedWeeks, numberOfMonths],
-		([$placeholder, $weekStartsOn, $locale, $fixedWeeks, $numberOfMonths]) => {
-			if (!isBrowser || !$placeholder) return;
+	effect([placeholder, weekStartsOn, locale, fixedWeeks, numberOfMonths], ([$placeholder]) => {
```
huntabyte, line 517:
```suggestion
	effect([placeholder], ([$placeholder]) => {
```

huntabyte: We need to remove these so this only fires when the `placeholder` changes and we aren't doubling with the following effect.

huntabyte (changes requested):

paoloricciuti:
> Also closes #725

Are you sure about that? It seems that is asking of dynamically set the first week of the day based on the locale...is that automatic?

src/lib/builders/calendar/create.ts
paoloricciuti: Duh I'm stupid lol...I removed initially, then CMD+Z to copy and forgot to re-remove it lol

huntabyte:
> > Also closes #725
> 
> Are you sure about that? It seems that is asking of dynamically set the first week of the day based on the locale...is that automatic?

We can add a test to confirm but I believe it's because the month wasn't being recreated (which handles that automatically if not specified)

paoloricciuti:
> > > Also closes #725
> > 
> > 
> > Are you sure about that? It seems that is asking of dynamically set the first week of the day based on the locale...is that automatic?
> 
> We can add a test to confirm but I believe it's because the month wasn't being recreated (which handles that automatically if not specified)

Do you know of a locale that has a different starting day?

paoloricciuti:
> > > Also closes #725
> > 
> > 
> > Are you sure about that? It seems that is asking of dynamically set the first week of the day based on the locale...is that automatic?
> 
> We can add a test to confirm but I believe it's because the month wasn't being recreated (which handles that automatically if not specified)

Ok checked and it's true...adding a test for that too

huntabyte (approved):
You're incredible! Can we make the same changes to the Range Calendar? If you don't have time I can take care of it!

paoloricciuti:
> You're incredible! Can we make the same changes to the Range Calendar? If you don't have time I can take care of it!

I should be able to do it

PR #1076, opened 2024-03-07
fix: datepicker not syncing with calendar

### LINKED ISSUE #1074: [Datepicker] options stores not syncronizing with underlying calendar
### Describe the bug

I tought #1069 was all the problem i was facing however there's another bug related to this specifically for the DatePicker (and i would guess for other composed builders too).

Basically here
https://github.com/melt-ui/melt-ui/blob/9dd987f366b32f12bd883e14ffe007cbc304093b/src/lib/builders/date-picker/create.ts#L198-L202
the options for the datepicker override the ones for dateField and calendar. This means that if you pass a default option (in my case was `numberOfMonths`) when setting the store you are actually setting the store in `options.numberOfMonths` and not the underlying `calendarOptions.numberOfMotnhs`.

This is why we have this series of effects
https://github.com/melt-ui/melt-ui/blob/9dd987f366b32f12bd883e14ffe007cbc304093b/src/lib/builders/date-picker/create.ts#L110-L132
This is easily fixable by adding another effects that sync `numberOfMonths` too but i think the same should apply for `fixedWeeks`, `weekStartsOn` and all the other properties.

I do wonder however if there's a way to automate this to make it more reliable in the future (and not forcing contributors to remember to update both places when adding a new prop for example).

I can prepare a PR for the simple case but would love to know if we want to tackle the overall approach (i also kinda need this so maybe i can put up the quick PR and then we can discuss if we want to automate it? 🥺 )

### Reproduction

- go to https://www.sveltelab.dev/ruu6h9eabwthdpr (p.s. i'm using tarball to install melt since it's the version that has the fixes to calendar and range calendar which is still to publish)
- open Calendar.svelte (i'm using the datepicker builder which is the one affected but used only the calendar because that's where the bug is and i already had the component setup)
- comment the option `numberOfMonths`
- try to click on the button to add number of months
- observe that the calendar actually update
- uncomment the option
- try to click on the button to add number of months
- observe that the calendar no longer update

### Logs

_No response_

### System Info

```bash
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.3 - /usr/local/bin/pnpm
  npmPackages:
    @melt-ui/svelte: latest => 0.75.3 
    svelte: ^4.1.2 => 4.2.12
```

### Severity

blocking an upgrade

paoloricciuti:
Closes #1074 

I initially added the code to sync the options in the range date picker to but i just now realized while writing the tests that the range date picker doesn't expose the calendar options like `numberOfMonths`, `fixedWeeks` or `weekStartsOn`. I have everything stashed locally so if you prefer that i expose those props in another PR and then sync them i can do it. In the meantime here it is.

EDIT: i've also fixed a couple of test names in my two old PRs

### DIFF
--- a/src/lib/builders/date-picker/create.ts
+++ b/src/lib/builders/date-picker/create.ts
@@ -163,7 +163,7 @@ export function createDatePicker(props?: CreateDatePickerProps) {
 		'minValue',
 		'maxValue'
 	);
-	let calendarOptions = omit(
+	const calendarOptions = omit(
 		calendar.options,
 		'locale',
 		'disabled',
--- a/src/lib/builders/date-picker/create.ts
+++ b/src/lib/builders/date-picker/create.ts
@@ -131,6 +131,30 @@ export function createDatePicker(props?: CreateDatePickerProps) {
 		calendar.options.maxValue.set($maxValue);
 	});
 
+	if (options.numberOfMonths != undefined) {
+		effect([options.numberOfMonths], ([$numberOfMonths]) => {
+			if ($numberOfMonths != undefined) {
+				calendar.options.numberOfMonths.set($numberOfMonths);
+			}
+		});
+	}
+
+	if (options.fixedWeeks != undefined) {
+		effect([options.fixedWeeks], ([$fixedWeeks]) => {
+			if ($fixedWeeks != undefined) {
+				calendar.options.fixedWeeks.set($fixedWeeks);
+			}
+		});
+	}
+
+	if (options.weekStartsOn != undefined) {
+		effect([options.weekStartsOn], ([$weekStartsOn]) => {
+			if ($weekStartsOn != undefined) {
+				calendar.options.weekStartsOn.set($weekStartsOn);
+			}
+		});
+	}
+
 	const dateFieldOptions = omit(
 		dateField.options,
 		'locale',
@@ -139,7 +163,7 @@ export function createDatePicker(props?: CreateDatePickerProps) {
 		'minValue',
 		'maxValue'
 	);
-	const calendarOptions = omit(
+	let calendarOptions = omit(
 		calendar.options,
 		'locale',
 		'disabled',
--- a/src/lib/builders/date-picker/create.ts
+++ b/src/lib/builders/date-picker/create.ts
@@ -16,6 +16,7 @@ import { pickerOpenFocus } from '$lib/internal/helpers/date/focus.js';
 import { createFormatter, dateStore, getDefaultDate } from '$lib/internal/helpers/date/index.js';
 import type { MeltActionReturn } from '$lib/internal/types.js';
 import type { DatePickerEvents } from './events.js';
+import { defaults as calendarDefaults } from '../calendar/create.js';
 
 const defaults = {
 	isDateDisabled: undefined,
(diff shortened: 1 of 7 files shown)


src/lib/builders/date-picker/create.ts
```diff
@@ -131,6 +131,30 @@ export function createDatePicker(props?: CreateDatePickerProps) {
 		calendar.options.maxValue.set($maxValue);
 	});
 
+	if (options.numberOfMonths != undefined) {
```
huntabyte, line 134: Why do we need this conditional here but not for the others? This _should_ always be defined considering we're creating a store for each prop.

Does this not work if its the same as the ones above it?

huntabyte: Ahh I see, makes sense!

paoloricciuti: I don't have a super clear idea of the implications creating the default all the times means that there's an extra step that always need to happen (the synchronization) even if someone doesn't pass a default value for it (currently they interact directly with the calendar store in that case)

paoloricciuti: Oh and btw this was sort of a question...like: what would you like to do? 😁

huntabyte: I think for now we apply the defaults to keep it in-line with the rest of it in the event we add more props in the future or change a specific prop for the calendar that only would apply as the default on the date picker. I should have had this in there from the beginning the more that I think about it!

paoloricciuti: Gotcha boss...will work on this tomorrow.

paoloricciuti: Uh now that I think about it...this might be a change in the typings (I have to verify) and thus a breaking change (number of months is possibly undefined even in the front facing builder). Could that be considered a breaking change?

paoloricciuti: @huntabyte i made the change and i can confirm that it will change the frontfacing types. Technically this is a breaking change but should be a small one since something that could've been undefined it's now never undefined. What do we want to do about this?

cc. @TGlide

TGlide: Breaking changes are fine, we just mark them as a minor patch.

paoloricciuti: Gotcha...just pushed the changes then 😄

huntabyte: Do you mean return type differences? I don't see a change to the `types.ts` file here! @paoloricciuti

paoloricciuti: > Do you mean return type differences? I don't see a change to the `types.ts` file here! @paoloricciuti

Yeah return type differences...if you use `createDatePicker` and destructure `numberOfMonths` now is of type `number` before it was `number | undefined`

PR #1078, opened 2024-03-08
fix not reusing portals

anatolzak:
We should not reuse portals by default.
For example, if we have a tooltip inside a popover, if we reuse portals and set `overflow: hidden` on the popover, the tooltip gets clipped.

Portals solve many issues but the benefits of them disappear if you reuse portals for multiple components.

Radix UI doesn't reuse portals. A tooltip inside a popover is rendered in a new portal.

### DIFF
--- a/.changeset/tough-bees-hunt.md
+++ b/.changeset/tough-bees-hunt.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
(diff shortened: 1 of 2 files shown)


huntabyte:
This breaks more things than it doesn't. If you don't see the benefit in it then you can manually override it and send it to the body on an as-needed basis.

Issue #1077, opened 2024-03-08
Interact Outside Issue When Mounting Tooltip in Popover in separate portals


anatolzak:
### Describe the bug

If you have a tooltip inside a popover and you mount the tooltip in a separate portal than the popover, if you click inside the tooltip, the popover closes while the tooltip stays open.

The reason to mount the tooltip in a separate portal is for example to prevent clipping issues if `overflow: hidden` is applied on the popover.

The reason this happens is because when clicking inside the tooltip, interact outside triggers and closes the tooltip. We should not trigger interact outside when the click happens inside portals.

https://github.com/melt-ui/melt-ui/assets/53095479/610e949e-5e78-4ad3-8cb1-4c73cb685efb

And adding an `onOutsideClick` to popover with `e.preventDefault()` doesn't help and introduces the following issues:
1. Interacting outside the popover doesn't work
2. Text selection inside the tooltip is disabled

https://github.com/melt-ui/melt-ui/assets/53095479/4d056081-531c-4019-aad1-9f8d4945eba5

### Reproduction

https://stackblitz.com/edit/github-kwo5ue?file=src%2Froutes%2F%2Bpage.svelte

### System Info

```bash
System:
    OS: Linux 5.0 undefined
    CPU: (6) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.3 - /usr/local/bin/pnpm
  npmPackages:
    @melt-ui/svelte: latest => 0.75.3 
    @sveltejs/kit: ^2.0.0 => 2.5.2 
    svelte: ^4.1.2 => 4.2.12
```

### Severity

annoyance

anatolzak:
This is fixed in this PR https://github.com/melt-ui/melt-ui/pull/1064

huntabyte:
We need to have further discussion on this. The PR submitted reintroduces this problem: https://github.com/melt-ui/melt-ui/issues/1036 which should have never existed as it was a bug (solved by the PR that closed #1036).

[...]

PR #1080, opened 2024-03-10
Feat: add sibling portal support & implement interact outside the right way

anatolzak:
closes https://github.com/melt-ui/melt-ui/issues/1077, https://github.com/melt-ui/melt-ui/issues/1153, https://github.com/melt-ui/melt-ui/issues/1167

This PR adds support for sibling portals by correctly implementing outside interactions using a stack. Previously outside interactions did not work correctly when using sibling portals.

Also, please don't be intimidated by the "size" of this PR. It's mostly tests and other than that, this PR is 70 additions and 161 deletions.

Throughout this PR, I will mention the term "floating element", what I mean by that is anything that can be mounted into a portal (i.e. dialog, tooltip, popover, link preview, menu, tooltip, listbox etc.).

The correct way to implement outside interactions is by only closing the top-most floating element given the top-most element has `closeOnOutsideClick: true`. If the top-most element has `closeOnOutsideClick: false`, no elements should close since the top-most element "asked" to not close on an outside interaction.

I also fixed an issue with the Tags Input when it's nested inside another floating element like a Popover. When editing a tag, if you click outside, we should just cancel editing. However, the popover would also close. This is solved by using the new `useInteractOutside()` on the `edit` builder to prevent other floating elements from reacting to outside interactions.

Here is a video showing nested floating elements and outside interactions only closing the top-most floating element. (ignore the craziness of the nested elements :))

https://github.com/melt-ui/melt-ui/assets/53095479/b7a48103-2d51-4f80-92b3-9fc55fd059e7

Also, to prevent issue https://github.com/melt-ui/melt-ui/issues/1167, where interact outside doesn't fire when clicking outside opens a new floating element, which is caused by checking too late whether the element is at the top of the stack. This is because the outside interaction triggers the opening of another floating element, so by the time pointerup fires for the first floating element, the second floating element had already been added to the stack. This is prevented in this PR by checking in pointerup whether the element was at the top of the stack during pointerdown. I added a regression test for this as well.

### Before
https://github.com/melt-ui/melt-ui/assets/53095479/cbd69a3b-bfdf-4379-82e4-52e4f0385664

### After

https://github.com/melt-ui/melt-ui/assets/53095479/c539079d-3d36-433f-a9e7-d1c810a1de56


anatolzak:
And in case you notice the following bug:
In the escape key video example, when I press escape when the select is open, the focus goes back to the popover trigger, which is incorrect. The focus should be trapped inside the popover content. The reason this happens is because we for some reason have `focusTrap: null` on the listbox menu builder. Changing it to `{ returnFocusOnDeactivate: false, clickOutsideDeactivates: true, escapeDeactivates: true }` fixes this issue (and we can add an option to disable focus trapping). I will create a separate issue and PR for this.

huntabyte:
Oh no not the timezone tests failing due to DST again! I'll get that fixed later today on the develop branch so you can merge it in!

huntabyte:
That failing test may just need an `await waitFor()` for the focus. If it still doesn't pass then it may have been introduced by this PR

anatolzak:
@huntabyte not sure why the tests are not passing. I tried to manually test the functionality that's not passing the tests and at first glance, things seem to be working correctly. Will look into it tomorrow

anatolzak:
@huntabyte it seems like merging the following PR will fix the tests for Date Picker https://github.com/melt-ui/melt-ui/pull/1055
it looks like the root issue was with the popover component which the Date Picker uses

huntabyte:
The tests were passing without #1055 before so we need to identify what is breaking here with this PR

anatolzak:
@huntabyte this PR is finally ready to be reviewed! :)

I split up some of the changes in this PR onto different PRs to make reviewing this PR easier. So the following PRs should be merged before merging this one:

https://github.com/melt-ui/melt-ui/pull/1087 - this makes sure that we correctly run cleanup functions returned from actions. We currently don't do this correctly due to race conditions and promises.
https://github.com/melt-ui/melt-ui/pull/1095 - fix focus trap deactivating when `closeOnEscape: false`
https://github.com/melt-ui/melt-ui/pull/1094 fix focus trap deactivating when `closeOnOutsideClick: false`
https://github.com/melt-ui/melt-ui/pull/1088 - this removes the need for unnecessary if checks checking the type of clean up functions returned from actions, which reduces the code significantly.

huntabyte:
Alright, all of the dependent PRs have been merged!

anatolzak:
@huntabyte let me resolve the conflicts

huntabyte:
Will defer to @TGlide to give this the final review/approval 😃

anatolzak:
@huntabyte the tests are failing because I removed the escape-keydown changes to a different PR https://github.com/melt-ui/melt-ui/pull/1142 which is what is failing the tests here. Merging https://github.com/melt-ui/melt-ui/pull/1142 will solve this.

alexbjorlig:
Hi everyone, the work here is amazing 🚀 Just wanted to let you all know that in 21RISK we are also facing this issue, so it will be so good when it's fixed.

If there is anything we can help with let us know.

huntabyte:
Hey @alexbjorlig, thanks for your patience! There are a lot of small (but mighty) internal changes going on simultaneously, so we're doing our best to ensure we don't introduce a regression into the library!

TGlide (approved):
Sorry for the delay! Amazing PR as always @anatolzak 🎉

Issue #519, opened 2023-09-12
🌟Feature request: Better animation support


TGlide:
Sometimes, users may want to implement more context-aware animations, e.g. knowing where a popover menu is placed. As an example, Radix exposes this via data-attributes and CSS vars: https://www.radix-ui.com/primitives/docs/components/popover#origin-aware-animations

iolyd:
Other than not updating elements' data-attributes accordingly, builders currently do not reflect any automatic adjustments in the returned `positioning` store or any returned store. Is this also on the table? It feels a bit counter-intuitive to have an explicitly reactive value be returned but it not being representative of what is applied to the elements' styles.

huntabyte:
> Other than not updating elements' data-attributes accordingly, builders currently do not reflect any automatic adjustments in the returned `positioning` store or any returned store. Is this also on the table? It feels a bit counter-intuitive to have an explicitly reactive value be returned but it not being representative of what is applied to the elements' styles.

This would certainly be cool. I agree it's a bit counter-intuitive that the updates only flow one direction.

I'm trying to imagine what that might look like - maybe observers that watch the element and update the store as it updates.

iolyd:
I had in mind more something like hooking into floating-ui's state through some middleware or already available lifecycle callback. Just starting to look around, but maybe something here: https://github.com/melt-ui/melt-ui/blob/bb148fee4e263603b1679a48d65381018b3615fb/src/lib/internal/actions/floating/action.ts#L35-L147

TGlide:
Allowing more fine-tuned control over floating UI, with middlewares or why not, is definitely something I want to implement :)

iolyd:
> > Other than not updating elements' data-attributes accordingly, builders currently do not reflect any automatic adjustments in the returned `positioning` store or any returned store. Is this also on the table? It feels a bit counter-intuitive to have an explicitly reactive value be returned but it not being representative of what is applied to the elements' styles.
> 
> I'm confused what you mean by that? As the positioning store only reflects the configuration that is passed to floating-ui, rather than any actual values.

Sorry for the lack of clarity. But essentially I meant to say that when I first used builders with floating elements, I was expecting the `options.positioning` store would expose the (re-)computed [placement](https://floating-ui.com/docs/middleware#placement) rather than the initial placement.

sebasmaltese:
https://github.com/melt-ui/melt-ui/pull/1082 just fixed this 🥳

PR #1082, opened 2024-03-10
feat: `data-side` & `data-align` for floating content

huntabyte:
This PR introduces small yet mighty improvements to the "floating" content and arrows, which apply `data-side` and `data-align` attributes to the various "floating" elements containing the _true_ positioning of the element after the floating UI computation.

`data-side` can be one of - `'top' | 'left' | 'bottom' | 'right'`
`data-align` can be one of - `'start' | 'end' | 'center'`

The arrow component just receives the `data-side` attribute, so you can easily adjust the style of the arrow depending on the side it's actually rendered on.

### DIFF
--- a/.changeset/popular-meals-walk.md
+++ b/.changeset/popular-meals-walk.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+Floating: `data-side` & `data-align` for floating content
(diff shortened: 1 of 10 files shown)


sebasmaltese:
Thanks! This closes https://github.com/melt-ui/melt-ui/issues/519

Issue #1086, opened 2024-03-14
[Popover, Context Menu, List Box, Menu, MenuBar] provided `forceVisible: false`, cleanup functions don't run


anatolzak:
### Describe the bug

Currently, any builder that has an effect that is based on `isVisible` and some sort of trigger that is set when the content is opened, the clean up functions returned from actions don't run.

To give a specific example, I added the following code to the `usePopper` action to log when the pointer enters the content. 

```typescript
const usePopper = (node) => {
  const onPointerEnter = () => console.log('pointer entered');
  node.addEventListener('pointerenter', onPointerEnter);

  return { destroy: () => node.removeEventListener('pointerenter', onPointerEnter) };
}
```

Then in the popover content, we have an effect that cleans up the previous popper and creates a new action.

```typescript
const content = makeElement(name('content'), {
  action: (node: HTMLElement) => {
    let unsubPopper = noop;

    const unsubDerived = effect([isVisible, activeTrigger], ([$isVisible, $activeTrigger]) => {
      unsubPopper();
      if (!$isVisible || !$activeTrigger) return;

      tick().then(() => {
        unsubPopper = usePopper(node).destroy;
      });
    });

    return {
      destroy() {
        unsubDerived();
        unsubPopper();
      },
    };
  },
});
```

However, the cleanup doesn't actually run as can be seen in the below video. Each time the pointer enters the popover content in the video, there should only be 1 log.

https://github.com/melt-ui/melt-ui/assets/53095479/50bf4d37-2df6-43d3-994c-c3586773a4af

### Reproduction

-

### System Info

```bash
System:
    OS: macOS 14.3
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 50.02 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    Yarn: 1.22.15 - ~/.yarn/bin/yarn
    npm: 10.1.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
    pnpm: 8.11.0 - /usr/local/bin/pnpm
    bun: 1.0.22 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 122.1.63.161
    Chrome: 120.0.6099.109
    Edge: 114.0.1823.82
    Safari: 17.3
  npmPackages:
[...]

PR #1087, opened 2024-03-14
fix action cleanup function not running due to race condition and promises

anatolzak:
closes https://github.com/melt-ui/melt-ui/issues/1086

See the above issue for the full details. 
The action cleanup functions don't run in the following condition:
- When setting the cleanup function after a promise (i.e. tick())
- When the effect runs twice "concurrently", which can happen when updating more than 1 store that the effect depends on

Let's assess what happens with the following popover code assuming forceVisible is `false` and so `isVisible` matches the open state. Also, for context, the popover `handleOpen` sets the open state to true and sets the active trigger as well, which causes the following action to run twice "concurrently".

```typescript
const content = makeElement(name('content'), {
    let unsubPopper = noop;

    const unsubDerived = effect([isVisible, activeTrigger], ([$isVisible, $activeTrigger]) => {
      unsubPopper();
      if (!$isVisible || !$activeTrigger) return;

      tick().then(() => {
        unsubPopper = usePopper(node).destroy;
      });
    });

    return {
      destroy() {
        unsubDerived();
        unsubPopper();
      },
    };
  },
});
```

In the first run, we call `unsubPopper` which is `noop`. Then we create a promise in `tick()`. Before the promise gets resolved, the action gets rerun a second time. We call `unsubPopper` which is still `noop`. The second rerun creates a promise as well. Now the first promise gets resolved and creates a popper action. And then the second rerun promise gets resolved and creates another popper action. And this is how we ended up not running the previous cleanup.

This is solved by running the cleanup after the promise in `tick()` is resolved.

Before & After
I added the following lines of code to `usePopper` to log when the pointer enters the popper element
```typescript
const onPointerEnter = () => console.log('pointer entered');
popperElement.addEventListener('pointerenter', onPointerEnter);
callbacks.push(() => popperElement.removeEventListener('pointerenter', onPointerEnter));
```

https://github.com/melt-ui/melt-ui/assets/53095479/bb2160e0-ecf6-4726-a298-9bdf01494205

### DIFF
--- a/.changeset/sharp-files-deliver.md
+++ b/.changeset/sharp-files-deliver.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+Fix bug where action doesn't get cleaned up due to race conditions with promises (closes #1086)
--- a/.changeset/sharp-files-deliver.md
+++ b/.changeset/sharp-files-deliver.md
@@ -2,4 +2,4 @@
 '@melt-ui/svelte': patch
 ---
 
-Fix bug where action doesn't get cleaned up due to race conditions with promises (closes #1086)
+Fixed memory leak caused by race conditions for various components (closes #1086)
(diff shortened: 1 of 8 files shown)


huntabyte:
Yep! That's a memory leak alright. Nice catch on this one!

huntabyte (approved):
Thank you!

PR #1101, opened 2024-03-17
docs: fix tooltip always mounted

anatolzak:
We should not keep the tooltip always mounted. 
After this PR https://github.com/melt-ui/melt-ui/pull/1080, this will cause the tooltip to always stay in the stack and so escape key presses will behave unexpectedly. 

So for example, in the docs we have 2 tooltips using this component--the theme switcher and search. Since the tooltips are always mounted, the first tooltip when open will not react to an escape key press since it's not in the top of the stack.

### DIFF
--- a/src/docs/components/tooltip.svelte
+++ b/src/docs/components/tooltip.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
-	import { browser } from '$app/environment';
 	import { createTooltip, melt } from '$lib/index.js';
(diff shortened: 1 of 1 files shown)


huntabyte:
Should this PR be held until #1080 is merged?

anatolzak:
> Should this PR be held until #1080 is merged?

@huntabyte no, we can merge this any time

huntabyte (approved):
Thank you!

PR #1103, opened 2024-03-17
Fix: mounting each floating element into their own portal by default

anatolzak:
We should wait till https://github.com/melt-ui/melt-ui/pull/1080 is merged before this PR.

After https://github.com/melt-ui/melt-ui/pull/1080, we will correctly handle sibling portals and now the default should be to mount each floating element to their own portal by default.

RadixUI and MaterialUI both use separate portals for each floating element. 

Using separate portal fixes clipping and z-index issues. 

If you set `overflow: hidden` on the parent floating element, the child floating element will be clipped.

Also using separate portals fixes z-index issues. For example, with nested dialogs, the overlay of the nested dialog does not work correctly. The nested overlay does not show any visual difference over the main dialog because it is a child of the main dialog content so it does not "cover" it visually. Moving the nested dialog to its own portal fixes this issue since now the nested dialog portal comes after the main dialog portal in the DOM.

### Before 
https://github.com/melt-ui/melt-ui/assets/53095479/0fa00dd6-81dd-4780-a4b4-4fd52c3b3871

### After
https://github.com/melt-ui/melt-ui/assets/53095479/8d51090e-6f67-463b-9e05-b6515d13243f


huntabyte (approved):
Approved for merge once #1080 is merged.

TGlide (approved):

PR #1107, opened 2024-03-19
fix(combobox): menu close within composition session

shana0440:
Fix menu close by Enter event within composition session.

### Before

https://github.com/melt-ui/melt-ui/assets/6259797/1a8c9142-948f-44d6-ae79-f1cda1443111

### After

https://github.com/melt-ui/melt-ui/assets/6259797/aabe1731-feb3-4e42-95c7-101b770f7a13


huntabyte (approved):
Thank you!

Issue #1109, opened 2024-03-20
[Dialog, Popover] manually setting `open` to `false` does not restore focus to trigger


anatolzak:
### Describe the bug

This issue was first mentioned here https://github.com/melt-ui/melt-ui/issues/960, but figured I would open a new issue since https://github.com/melt-ui/melt-ui/issues/960 is about exposing additional stuff.

When using a dialog or popover, if you manually set the `open` store to `false`, the focus does not restore to the trigger.

https://github.com/melt-ui/melt-ui/assets/53095479/7fd6ae9a-4431-4df0-841b-f3ebca9f23b8

### Reproduction

https://stackblitz.com/edit/github-bn3cdm?file=src%2Froutes%2F%2Bpage.svelte

### System Info

```bash
System:
    OS: Linux 5.0 undefined
    CPU: (6) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.3 - /usr/local/bin/pnpm
  npmPackages:
    @melt-ui/svelte: latest => 0.76.0 
    @sveltejs/kit: ^2.0.0 => 2.5.4 
    svelte: ^4.1.2 => 4.2.12
```

### Severity

annoyance

PR #1110, opened 2024-03-20
fix(dialog,popover): restore focus when `open` set to `false`

anatolzak:
closes https://github.com/melt-ui/melt-ui/issues/1109

We now restore focus to the trigger (or using the `closeFocus` prop) in an `effect` that runs when `open` is `false`. In order to not focus the trigger on initial mount when `open` is `false`, I added a parameter `skipFirstRun` to the `effect` function that optionally skips the first run of the effect.

I also added regression tests for the dialog and popover. 

Also, in the menu builder, we are already doing what I implemented with the dialog and popover, which removes the need in the menu builder to manually focus the trigger in multiple places. And I cleaned up the effect in the menu component. 

### Dialog Before
https://github.com/melt-ui/melt-ui/assets/53095479/584dd6ad-bd39-4901-9ff1-2a0e987f1a2e

### Dialog After 
https://github.com/melt-ui/melt-ui/assets/53095479/ac7ac1d1-f02e-4b2a-8f48-040438149a43

### Popover Before

https://github.com/melt-ui/melt-ui/assets/53095479/89b2dc56-e1a8-47e3-9ce3-d4c42fbe41c3

### Popover After

https://github.com/melt-ui/melt-ui/assets/53095479/fa3943a1-35d1-417b-afe9-cb4f5650b69d

### DIFF
--- a/src/lib/internal/helpers/store/effect.ts
+++ b/src/lib/internal/helpers/store/effect.ts
@@ -10,18 +10,25 @@ import { safeOnDestroy } from '../lifecycle.js';
  * @template S - The type of the stores object
  * @param stores - The stores object to derive from
  * @param fn - The function to run when the stores change
+ * @param skipFirstRun - Whether to skip the first run
  * @returns A function that can be used to unsubscribe the effect
  */
 export function effect<S extends Stores>(
 	stores: S,
-	fn: (values: StoresValues<S>) => (() => void) | void
+	fn: (values: StoresValues<S>) => (() => void) | void,
+	skipFirstRun?: boolean
 ): () => void {
 	let cb: (() => void) | void = undefined;
+	let isFirstRun = true;
 
 	// Create a derived store that contains the stores object and an onUnsubscribe function
 	const destroy = derived(stores, (stores) => {
 		cb?.();
-		cb = fn(stores);
+		if (isFirstRun && skipFirstRun) {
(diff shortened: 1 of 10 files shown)


src/lib/internal/helpers/store/effect.ts
```diff
@@ -10,18 +10,25 @@ import { safeOnDestroy } from '../lifecycle.js';
...
  * @param stores - The stores object to derive from
  * @param fn - The function to run when the stores change
+ * @param skipFirstRun - Whether to skip the first run
  * @returns A function that can be used to unsubscribe the effect
  */
 export function effect<S extends Stores>(
 	stores: S,
-	fn: (values: StoresValues<S>) => (() => void) | void
+	fn: (values: StoresValues<S>) => (() => void) | void,
+	skipFirstRun?: boolean
```
huntabyte, line 19: Let's make this an object. @TGlide and I have spoke in the past about some other options we want to add to effects and by making it an object now it'll ease the addition of those.

```ts
type EffectOptions = {
  skipFirstRun?: boolean;
}
```

huntabyte (changes requested):
Minor thing! Thanks man!

anatolzak:
@huntabyte I added an options parameter to `effect` as you requested!

huntabyte (approved):
Thank you!

PR #1112, opened 2024-03-22
fix: overlay portal cleanup - prevent memory leak

anatolzak:
Prevent potential memory leak on popover overlay portal. Should cleanup portal action in the case that the `portal` store updates.


huntabyte (approved):
Thank you!

PR #1113, opened 2024-03-22
test: allow intercepting escape keydown events

anatolzak:
This prevents regression of https://github.com/melt-ui/melt-ui/pull/845

Added tests to allow a child to intercept escape keydown events without the dialog (or any other builder relying on `useEscapeKeydown()`) getting closed.


huntabyte (approved):
Thank you!

Issue #917, opened 2024-01-15
Dialog: closes when 1password icon is clicked


rafales:
### Describe the bug

1Password - and other password managers - typically have a widget that shows up in password fields and which is appended to body.

This means that if I have a password field inside a modal and closeOnClickOutside is enabled - then clicking that widget will close the modal, which is not expected behavior.

This functionality should be probably implemented with "click on overlay" instead of catching clicks outside.

### Reproduction

https://www.sveltelab.dev/b7dn7stqhkjikd0

IT WILL NOT WORK IN SVELTELAB: In SvelteLab 1password shows button in top-level page, while app itself is rendered in an iframe. So contents of an iframe never see the click on 1password button.

In order to reproduce you need 1password installed, fork this locally, then try clicking on 1password's icon next to password button.

### System Info

```bash
-
```

### Severity

annoyance

rafales:
https://github.com/melt-ui/melt-ui/assets/919812/2c007537-1645-4eb6-a354-7edd5f3d81c5

Here's a recording of the problem.

moreorover:
Selecting text inside input field with your mouse and dragging it outside the input field closes the dialog as well. This is related I understand.
[dialog outside.webm](https://github.com/melt-ui/melt-ui/assets/17711314/384af6ff-f88d-4887-9015-66390057704e)

huntabyte:
> Selecting text inside input field with your mouse and dragging it outside the input field closes the dialog as well. This is related I understand. [dialog outside.webm](https://github.com/melt-ui/melt-ui/assets/17711314/384af6ff-f88d-4887-9015-66390057704e)

This is actually unrelated to this specific issue.

PR #1114, opened 2024-03-22
Fix(interact outside): enable interact outside interception

anatolzak:
Closes https://github.com/melt-ui/melt-ui/issues/917, https://github.com/melt-ui/melt-ui/issues/1128

This PR adds the ability for a DOM element to intercept an outside interaction by calling `e.stopPropagation()`.

Let's use an example that demonstrates how the DOM behaves and the challenges it imposes for this PR:

On a mouse device, clicking an element will trigger all of the following events:
1. pointerdown
2. pointerup
3. mousedown
4. mouseup
5. click

If the user intercepts the `pointerdown` event, `pointerup` and `click` will still trigger while the mouse events will not.
If the user intercepts `mousedown` or `mouseup`, all the other events will still be triggered.

Let's use another example: on a touch device, clicking an element will trigger all of the following events:
1. pointerdown
2. touchstart
3. pointerup
4. touchend
5. mousedown
6. mouseup
7. click

If the user intercepts `touchstart` or `touchend`, the mouse and click events will NOT be triggered but the pointer events WILL.

Ok, so ideally for our sake, the user should be able to intercept any of the events and prevent an outside interaction.

Currently, we default to only using pointer events and fallback to mouse and touch events when pointer events are not available.

But in the case of issue https://github.com/melt-ui/melt-ui/issues/917, clicking on the 1Password button inside an input that is inside a dialog closes the dialog. Despite 1Password intercepting the mouse events, our `pointerup` handler is still triggered, leading to the unintended closure of the dialog.

So we should listen to all the appropriate events and trigger an outside interaction only if none of the events were intercepted by the user.

Similar to PR https://github.com/melt-ui/melt-ui/pull/845 that enabled intercepting an escape key down by not using `capture` listeners, in order to allow intercepting outside interactions, we should not trigger an outside interaction using capture listeners.

However the challenge is--how do you determine if any of the events were intercepted when the browser will trigger a bunch of other events nonetheless.

To identify intercepted events during an interaction, we implement a two-step approach for the following events: `pointerdown`, `touchstart`, `pointerup`, `touchend`, `mousedown`, `mouseup`, and `click`. First, we attach an event listener in the capture phase (`capture: true`) to flag the event as intercepted. Subsequently, we add another event listener for the bubbling phase, designed to tag the event as NOT intercepted if it passes through. This setup relies on the principle that the bubbling phase listener is called only if the event was not intercepted by the user. Conversely, if an event is intercepted, only the capture phase listener will trigger, which marks the event as intercepted.

So now we can easily check if any events were intercepted and not trigger an outside interaction in that case.

Since we attach event listeners for multiple events, the browser may trigger multiple events at the same time (i.e. `pointerdown` + `touchstart`, `pointerup` + `touchend`, etc.). In order to make sure that we don't run into race conditions and prematurely check whether any events were intercepted before all other bubbling handlers have potentially marked the events as not intercepted, we debounce the event handlers, which also prevents them from being executed more than once.

After an interaction has finished, we also need to correctly reset the `interceptedEvents` object so that future interactions that are not intercepted will behave as expected. We should only reset `interceptedEvents` at the end of an interaction (`touchend`, `pointerup`, `mouseup`, `click`) in case the user intercepted the beginning of an interaction while still expecting the whole interaction to be intercepted. We debounce resetting `interceptedEvents` to avoid race conditions. We should also call `resetInterceptedEventsDebounced` in the capture phase handlers since the bubbling phase handlers may not be called due to user interception.

I modified the dialog example in the docs and added buttons that intercept different events for demonstration purposes. The modified docs should not be included in the final PR.

Also, I added a lot of tests that make sure all this works correctly so don't be intimidated by the number of lines added by this PR :)

Thanks for reading all this :)

### Before

https://github.com/melt-ui/melt-ui/assets/53095479/21002268-8cff-4ed8-a8af-1f21e4f43a95

### After

https://github.com/melt-ui/melt-ui/assets/53095479/51cb8552-c1e5-4041-addc-b33930d6bb69

### Mouse Device

https://github.com/melt-ui/melt-ui/assets/53095479/c95a68dc-772f-43c8-8167-700309dc4bb2

### Touch Device

https://github.com/melt-ui/melt-ui/assets/53095479/c724abf0-e84f-4ecf-aae6-ed6f4111c367

### DIFF
--- a/.changeset/young-peaches-retire.md
+++ b/.changeset/young-peaches-retire.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+Fixed bug where outside interactions could not be intercepted (closes #917)
--- a/src/docs/previews/dialog/main/tailwind/index.svelte
+++ b/src/docs/previews/dialog/main/tailwind/index.svelte
@@ -4,6 +4,7 @@
 	import { flyAndScale } from '$docs/utils/index.js';
 	import { X } from '$icons/index.js';
 	import { fade } from 'svelte/transition';
+	import { usePortal } from '$lib/internal/actions/portal.js';
 
 	const {
 		elements: {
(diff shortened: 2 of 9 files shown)


huntabyte:
Hey @anatolzak this looks great, incredible investigative work to tackle this problem! Whenever you're ready (after cleaning up the demo) let me know and we'll get this merged!

anatolzak:
@huntabyte I am glad all this makes sense lol :)
I cleaned the demo so this PR is ready! 🎉

PR #1117, opened 2024-03-23
chore: use testKbd in tests

anatolzak:


huntabyte (approved):
You're incredible! Thank you!

Issue #1115, opened 2024-03-23
Webkit Browsers: Dialog overlay doesn't prevent bubbling on mobile


hugotox:
### Describe the bug

Initially discovered in [shadcn Dialog](https://github.com/huntabyte/shadcn-svelte/issues/548), this is still happening. I don't have a direct dependency of melt-ui, but just `"bits-ui": "^0.20.0"` 

> When opening a dialog on mobile devices, the overlay is not preventing interactions with elements behind it. This behavior doesn't happen on desktop.

### Reproduction

Just open a dialog on mobile, and notice how the elements behind the overlay are interactive.
<img width="716" alt="291118255-988171c0-b307-4dae-9b5a-5e02d747d569" src="https://github.com/melt-ui/melt-ui/assets/1758062/d7c3e4c4-f72d-4ee8-959f-548bbe9f7879">

https://stackblitz.com/edit/github-hrerew?file=src%2Froutes%2F%2Bpage.svelte

### System Info

```bash
System:
    OS: macOS 12.0
    CPU: (10) arm64 Apple M1 Pro
    Memory: 96.73 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 18.19.0 - ~/Library/pnpm/node
    npm: 10.2.3 - ~/Library/pnpm/npm
    pnpm: 8.14.2 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 123.0.6312.59
    Edge: 122.0.2365.92
    Safari: 15.0
```

### Severity

annoyance

huntabyte:
Can this be reproduced using an actual mobile device and not the mobile preview in the browser?

hugotox:
Yes, it happens on webkit browsers: safari, duckduckgo. Doesn't happen on chrome

huntabyte:
Of course, it's the webkit browsers 🤦‍♂️

We'll try to determine what the cause of this is and how we can get around it.

anatolzak:
@huntabyte I was able to fix this issue in this PR https://github.com/melt-ui/melt-ui/issues/1115

PR #1118, opened 2024-03-24
Fix(interact outside) touch devices leak click events to dom elements

anatolzak:
closes https://github.com/melt-ui/melt-ui/issues/1115

This closes an issue where clicking outside a dialog on a touch device would trigger a click event on the element underneath the dialog.

This only happens when using pointer events on touch devices. After `pointerup` is called on touch devices, we need to wait for a `click` event before triggering an outside interaction. There is a delay between `pointerup` and `click` during which we used to close the dialog and hence trigger a click event on the element under the dialog.

Technically this issue should also occur when using touch events, which we use when pointer events are not available, however this issue doesn't occur in our case when using touch events because we call `e.preventDefault()` which will prevent a click event occurring after a `touchend`

### Before

https://github.com/melt-ui/melt-ui/assets/53095479/0e1f2964-d5f2-4465-bda9-262f1435a683

https://github.com/melt-ui/melt-ui/assets/53095479/9697e482-26b7-4fc1-95ea-0dcad5f369ae

### After

https://github.com/melt-ui/melt-ui/assets/53095479/fc2187a4-9ecd-429e-bae7-a6c8877bb0ff

https://github.com/melt-ui/melt-ui/assets/53095479/0218571f-82e1-460b-a4ca-5fbc8c4a7fa0

### DIFF
--- a/.changeset/wise-schools-complain.md
+++ b/.changeset/wise-schools-complain.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+Fixed a bug on touch devices where an outside interaction leaked click events to other dom elements (closes #1115)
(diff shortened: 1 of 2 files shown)


huntabyte (approved):
You're insane! Thank you!

PR #1121, opened 2024-03-25
test: dialog touch events when overlay is clicked

anatolzak:
add regression test for https://github.com/melt-ui/melt-ui/pull/1118


anatolzak:
let me know what you think about whether the `test` script should run both vitest and playwright. currently, the DX is really nice with vitest and watch mode and only rerunning tests that depend on the new files changed. not sure if we should include playwright in the default test scripts or just have it run on pull requests in github actions.

huntabyte (approved):
Thank you!

PR #1124, opened 2024-03-25
Feat: add `autoSort` option for sliders

### LINKED ISSUE #1122: 🌟Feature request: Add option to keep order of Slider values
**Is your feature request related to a problem? Please describe.**

Currently when using a slider with multiple values, the value array gets automatically sorted when moving thumbs past each other instead of keeping the original order. For example, for `[10, 20, 30]` when moving the first thumb to 25 it automatically sets the value store to `[20, 25, 30]` instead of `[25, 20, 30]`. If the different indexes hold some important meaning, it could be problematic if they get switched.

**Describe the solution you'd like**

It would be nice to have an extra option to not swap the values automatically. It could be called `enableSwap` / `disableSwap` or something like that.

bennymi:
Closes #1122.

I added a `disableSwap` option to sliders which disables the automatic ordering of the value array. Therefore it's now possible to have arrays like these: `[25, 20, 30]`, whereas before the value arrays were always automatically sorted. This is important because in some instances users might want indexes to hold specific meaning, so if the values get swapped around there's no way of keeping track of the individual values.

I also added an example for it, which demonstrates the different behaviors:
image

### DIFF
--- a/.changeset/new-hairs-vanish.md
+++ b/.changeset/new-hairs-vanish.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+slider: add disable swapping option to avoid sorting of values array
--- a/src/docs/content/builders/slider.md
+++ b/src/docs/content/builders/slider.md
@@ -61,10 +61,10 @@ You can add slider ticks using the `ticks` state and the `tick` element returned
     <svelte:component this={previews.ticks} />
 </Preview>
 
-### Disable Index Swapping
+### Auto Sorting
 
 By default, the `value` array is always automatically sorted. You can disable this behavior by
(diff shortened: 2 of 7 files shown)


abdel-17:
I suggest we name the prop `disableAutoSort`. It makes it clear which functionality is being disabled. `swap` isn’t very clear imo.

huntabyte:
> I suggest we name the prop disableAutoSort. It makes it clear which functionality is being disabled. swap isn’t very clear imo.

Agreed on the clarity of "swap". 

Even simplifying it to `autoSort` with a default of `true` would suffice and be self-explanatory.

TGlide (approved):
Thank you, great PR! 🎉

PR #1127, opened 2024-03-27
chore: refactor dialog tests

anatolzak:
Currently in our dialog tests, every test repeats the assertions for opening a dialog so I refactored it into an `open` function.
Also most tests need `getByTestId('content')` and `getByTestId('overlay')`, so I added them to the object returned from `setup`


huntabyte (approved):
Thank you!

PR #1129, opened 2024-03-27
fix(link-preview): reset activeTrigger on trigger unmount

anatolzak:
reset activeTrigger on trigger unmount. Currently, if you unmount the trigger, the content stays open because we never reset the `activeTrigger`. So on trigger destroy, we set `activeTrigger` to `null`

### Before
https://github.com/melt-ui/melt-ui/assets/53095479/edacd63f-160a-4241-9ef5-c0a1b48507b2

### After

https://github.com/melt-ui/melt-ui/assets/53095479/767ac36f-988b-4483-9041-7f4348b7f689

### DIFF
--- a/.changeset/old-taxis-yell.md
+++ b/.changeset/old-taxis-yell.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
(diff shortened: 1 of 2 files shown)


huntabyte (approved):
Thank you!

Issue #1132, opened 2024-03-29
[Focus trap] focus trap deactivates on escape that was intercepted


anatolzak:
### Describe the bug

When using a dialog for example with an element that intercepts an escape key. On an escape key press, the dialog doesn't close as expected, however the focus trap deactivates which is not the expected behavior.

https://github.com/melt-ui/melt-ui/assets/53095479/c8c65baf-9782-4c78-a45f-61ed92f0e17a

### Reproduction

https://stackblitz.com/edit/github-evusou?file=src%2Froutes%2F%2Bpage.svelte

### System Info

```bash
System:
    OS: Linux 5.0 undefined
    CPU: (6) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.3 - /usr/local/bin/pnpm
  npmPackages:
    @melt-ui/svelte: latest => 0.76.2 
    @sveltejs/kit: ^2.0.0 => 2.5.5 
    svelte: ^4.1.2 => 4.2.12
```

### Severity

annoyance

PR #1133, opened 2024-03-29
fix(focus trap): focus trap deactivates on intercepted escape

anatolzak:
closes https://github.com/melt-ui/melt-ui/issues/1132

We should not let the focus trap package deactivate the focus trap on an escape key press because they use capture events which will fire even if an element intercepted the event. Therefore we should never let the focus trap package deactivate based on an escape key press but rather when the dialog or popover unmount which is already implemented in the destroy method of the focus trap action.

### Before - I added an escape key interceptor on the dialog content

https://github.com/melt-ui/melt-ui/assets/53095479/5dccb797-25fb-420b-afd6-13ad5272fac2

### After

https://github.com/melt-ui/melt-ui/assets/53095479/9520d91e-7ff2-46b5-af8b-f5532460261e

### DIFF
--- a/.changeset/hungry-steaks-whisper.md
+++ b/.changeset/hungry-steaks-whisper.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+Fixed bug where focus trap would deactivate on escape that was intercepted (closes #1132)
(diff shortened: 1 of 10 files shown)


anatolzak:
@huntabyte I resolved the conflicts :)

PR #1136, opened 2024-03-29
refactor: focus-trap

anatolzak:
This PR should be merged after https://github.com/melt-ui/melt-ui/pull/1135 and https://github.com/melt-ui/melt-ui/pull/1133

We currently set the same options for the focus trap action in multiple places (popover, dialog, popper). This PR moves all the same options that we use to be the default ones used in `createFocusTrap()` to avoid redundancy. Also as new issues come up with the focus trap (like the ones that are solved in the prerequisite PRs), there is the possibility of missing all the different places that use the focus trap that should behave the same way.

I also cleaned up `createFocusTrap()` and made it more similar to how our other actions behave.

And I cleaned up the dialog content builder.

### DIFF
--- a/.changeset/tough-squids-rescue.md
+++ b/.changeset/tough-squids-rescue.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+Refactor focus trap internals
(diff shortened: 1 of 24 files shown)


huntabyte (approved):
Thank you!

anatolzak:
@huntabyte sorry about all the merge conflicts... this is ready now as well. thanks so much!

PR #1139, opened 2024-03-30
fix(menu): overlay portal cleanup - prevent memory leak #1112

anatolzak:
Prevent potential memory leak on menu overlay portal. Should cleanup portal action in the case that the `portal` store updates.

### DIFF
--- a/src/lib/builders/menu/create.ts
+++ b/src/lib/builders/menu/create.ts
@@ -350,6 +350,7 @@ export function createMenuBuilder(opts: _MenuBuilderOptions) {
 			} as const;
 		},
 		action: (node: HTMLElement) => {
+			let unsubPortal = noop;
 			let unsubEscapeKeydown = noop;
 
 			if (closeOnEscape.get()) {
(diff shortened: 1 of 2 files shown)


src/lib/builders/menu/create.ts
```diff
@@ -361,17 +362,19 @@ export function createMenuBuilder(opts: _MenuBuilderOptions) {
...
 			}
 
-			const unsubPortal = effect([portal], ([$portal]) => {
-				if ($portal === null) return noop;
+			const unsubDerived = effect([portal], ([$portal]) => {
+				unsubPortal();
+				if ($portal === null) return;
 				const portalDestination = getPortalDestination(node, $portal);
-				if (portalDestination === null) return noop;
-				return usePortal(node, portalDestination).destroy;
```
TGlide, line 368: callbacks that are returned are called before a new effect runs, so I don't think this is necessary

anatolzak: @TGlide I didn't realize that! So yes, this is unnecessary

Issue #1141, opened 2024-03-30
🌟Feature request: Add more options to `closeOnEscape` prop


anatolzak:
**Is your feature request related to a problem? Please describe.**
When dealing with multiple nested floating elements and passing `closeOnEscape: false` on a child element, when the escape key is pressed, both the parent and the child close. I would expect neither the parent nor the child to close since I explicitly "asked" the child not to close on escape

**Describe the solution you'd like**
We should be able to: 
- pass `closeOnEscape: null` to specify that the parent floating element should be responsible when escape is pressed.
- pass `closeOnEscape: false` to specify that the child and any parents should not close when escape is pressed.

**Describe alternatives you've considered**

**Additional context**

huntabyte:
I feel like this is something we should be able to tackle internally so the user shouldn't have to think about it. Does the current stack setup in #1080 not support this behavior?

anatolzak:
> Also, turn on GitHub sponsors or drop a donate link so I can send you something for all the incredible work you've been doing recently 😃

@huntabyte I really appreciate that Hunter! It has been super fun doing all this! I submitted my application for Github sponsors and it's being reviewed now.



> Something along the lines of: "If I'm at the top of the stack and my `closeOnEscape` prop is `false`, don't do anything".

That's what https://github.com/melt-ui/melt-ui/pull/1080 implements. However, I am suggesting an additional option where the component is not responsible for the escape key press. 

For example if you want to have an open tooltip inside a dialog and that when escape is pressed, the dialog closes. Currently with https://github.com/melt-ui/melt-ui/pull/1080, the options are to either only close the tooltip on escape using `closeOnEscape: true` or to not close the tooltip on escape using `closeOnEscape: false`. There is no option for tooltip to ignore the escape and let it "bubble" to the underneath layers. 

Before https://github.com/melt-ui/melt-ui/pull/1080, when a tooltip was open inside a dialog and the tooltip has `closeOnEscape: false`, pressing escape would close the dialog even though the tooltip "asked" to not close when escape is pressed. But then there is no way to disable an escape while the tooltip is open without it cascading to the underneath layers.

Basically, what I am suggesting is that we add 3 options for `closeOnEscape`:
- `true`: on escape, closes the element
- `false`: on escape, doesn't close anything
- `null`, doesn't do anything to the current element and let's the event cascade to previous layers. And so if a previous layer has `closeOnEscape: true`, the previous layer will close which will then also close the top-most element.

PR #1142, opened 2024-03-30
Feat: use stacks in `useEscapeKeydown` & introduce `escapeBehavior` prop

anatolzak:
closes https://github.com/melt-ui/melt-ui/issues/1141, https://github.com/melt-ui/melt-ui/issues/1143

This PR fixes how we handle escape presses when dealing with multiple nested floating elements by maintaining a stack of the open elements. This was part of https://github.com/melt-ui/melt-ui/pull/1080 but I realized that it changed too many things and would make it unnecessarily difficult to review. 

Here is a video showing nested floating elements and outside escape presses only closing the top-most floating element.

https://github.com/melt-ui/melt-ui/assets/53095479/d9c2e50c-2062-4a0e-a515-37f12b602cc0

Previously we added `closeOnEscape` to the dependencies in the effect of every builder that was interested in the escape behavior. This breaks as soon as we use stacks, since updating `closeOnEscape` while an element is open removes the element from stack and then adds it to the top of the stack. To prevent that, this PR allows passing the `closeOnEscape` store to `useEscapeKeydown` which therefore allows updating `closeOnEscape` without reseting the position of the element in the stack. I added tests for this as well.

Now that we use a stack to keep track of open elements that are listening to the escape presses, I had to remove a lot of extra keydown event listeners that we had in many builders that would interfere when using multiple nested elements.

This PR also replaces the `closeOnEscape` builder prop with `escapeBehavior` which provides even further fine-grained control over the escape behavior in builders.

Previously when dealing with multiple nested floating elements and passing `closeOnEscape: false` to a child element, when the escape key was pressed, both the parent and the child would close. I would expect neither the parent nor the child to close since I explicitly "asked" the child not to close on escape. So previously there was no way to force an element to stay open when escape is pressed without triggering the parent to close as well.

To allow more fine-grained behaviors, we now introduce the `escapeBehavior` prop:
- `close`: Closes the element immediately.
- `ignore`: Prevents the element from closing and also blocks the parent element from closing in response to the Escape key.
- `defer-otherwise-close`: Delegates the action to the parent element. If no parent is found, it closes the element.
- `defer-otherwise-ignore`: Delegates the action to the parent element. If no parent is found, nothing is done.

Here are videos showing the behavior of adding different `escapeBehavior` values to a popover that is nested inside a dialog. The dialog's `escapeBehavior` is `close`.

### Popover `escapeBehavior: 'close'`

https://github.com/melt-ui/melt-ui/assets/53095479/8e670d2a-370a-4b6a-ba93-8f4d95c40236

### Popover `escapeBehavior: 'ignore'`

https://github.com/melt-ui/melt-ui/assets/53095479/38b7996b-5b1a-4b60-84ec-38fe25280ef1

### Popover `escapeBehavior: 'defer-otherwise-ignore'`

https://github.com/melt-ui/melt-ui/assets/53095479/d19ed8f4-2949-42cf-b749-4fa89f1c45e2

### Popover `escapeBehavior: 'defer-otherwise-close'`

https://github.com/melt-ui/melt-ui/assets/53095479/ec36f7e3-e3ad-403d-9ee9-23f7abc1c027

### Popover `escapeBehavior: 'defer-otherwise-close'` mounted as a single element

https://github.com/melt-ui/melt-ui/assets/53095479/62aa47a2-6b46-40b8-b5b3-e628b3d5b194


TGlide (approved):
Dude. this is amazing! Such a comprehensive PR. Thank you!

PR #1145, opened 2024-03-30
fix: mount portal only when isVisible is true

anatolzak:
Currently when using `forceVisible: false` and using a dialog, menu overlay, popover overlay, we mount the portal before the element is actually open. This can create z-index conflicts.

This is easily solved by only mounting the portal when `isVisible` is `true`

### DIFF
--- a/.changeset/breezy-buses-promise.md
+++ b/.changeset/breezy-buses-promise.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+Only mount the portal for dialog, menu overlay, and popover overlay when the component is set to open. This applies when `forceVisible: false` is used.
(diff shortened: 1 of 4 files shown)


huntabyte:
Does this conflict with transitions in a project like bits-ui?

anatolzak:
> Does this conflict with transitions in a project like bits-ui?

@huntabyte no, it does not. Bits UI uses `forceVisible: true` so this change doesn't affect bits ui at all since `isVisible` will always be `true`. Bits UI always keeps the portal mounted (because it doesn't mount the portal conditionally when `$open` is `true`)

TGlide:
This is unneeded I'd say, it'd just be more expensive really. There should be no z-index conflicts since the elements should be hidden most of the time anyways.

anatolzak:
> This is unneeded I'd say, it'd just be more expensive really. There should be no z-index conflicts since the elements should be hidden most of the time anyways.

@TGlide Imagine you have a popover nested within a dialog and both should be rendered onto separate portals. If you have have the popover earlier in the svelte markup than the dialog, what ends up happing is that popover's portal gets mounted first and then the dialog's portal gets rendered second. Then when you open the dialog and then the popover, the popover is going to appear behind the dialog even if they both have the same z-index because the dialog appears after the popover in the HTML.

PR #1148, opened 2024-03-31
test(combobox): add interact outside tests

anatolzak:
add regression tests to prevent issue reported in https://github.com/melt-ui/melt-ui/pull/1038#issuecomment-2016526702 
when combobox is open and clicking outside on a different input, the combobox closes as expected, however the click on the input is prevented and the it does not get focused. You should be able to click onto an input while the combobox is open and have the input be focused after the combobox closes.


huntabyte (approved):
Nice!

PR #1149, opened 2024-03-31
chore: refactor tags input

anatolzak:
Add `isEditing` helper to avoid repeating logic for checking if a tag is being edited. 

This refactor is also essential to https://github.com/melt-ui/melt-ui/pull/1080/ where we add `useInteractOutside` to a tag only when it's being edited to prevent closing parent floating elements when clicking outside of a tag that's being edited.

### DIFF
--- a/.changeset/real-colts-run.md
+++ b/.changeset/real-colts-run.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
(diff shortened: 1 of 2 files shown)


anatolzak:
Closing this PR as these changes were included in https://github.com/melt-ui/melt-ui/pull/1080

Issue #1150, opened 2024-03-31
🌟Feature request: add more options to `closeOnOutsideClick` prop


anatolzak:
When dealing with multiple nested floating elements and passing `closeOnOutsideClick: false` on a child element, clicking outside is disabled.

It would be nice to be able to have an element not be counted as a floating element that should react to outside clicks. For example, the following is currently not possible: Having an open popover inside a dialog and on outside click, the dialog should close. Currently we can either pass `closeOnOutsideClick: true` to the popover, which will only close the popover on outside click. And if you pass `closeOnOutsideClick: false` to the popover, on outside click nothing will close.

**Describe the solution you'd like**
I suggest that we allow passing `closeOnOutsideClick: null` to not count an element in the interact-outside stack and let the parent element react to the outside interaction.

To recap, the new options for `closeOnOutsideClick` should be:
- `true`: on outside interaction, the element should close as long as it's the top-most floating element.
- `false`: on outside interaction, the element should not close.
- `null`: on outside interaction, the parent element should handle the interaction.

In https://github.com/melt-ui/melt-ui/pull/1080 we use a stack to maintain the top-most floating elements interested in outside interactions. the PR handles both `closeOnOutsideClick: true | false`. The only change that needs to be made is to simply not add an element to the stack provided `closeOnOutsideClick: null`.

anatolzak:
See https://github.com/melt-ui/melt-ui/pull/1152 for videos showing how each option behaves

PR #1152, opened 2024-03-31
feat(interact-outside): introduce `interactOutsideBehavior` prop

anatolzak:
closes https://github.com/melt-ui/melt-ui/issues/1150

Previously we added the `closeOnOutsideClick` store to the dependencies of the effect of every builder that was interested in the outside-click behavior. This breaks as soon as we use stacks, since updating the `closeOnOutsideClick` option while an element is open removes the element from stack and then adds it to the top of the stack. To prevent that, this PR allows passing the `closeOnOutsideClick` store to `useInteractOutside` which therefore allows updating `closeOnOutsideClick` without reseting the position of the element in the stack. I added tests for this as well.

This PR replaces the `closeOnOutsideClick` builder prop with `interactOutsideBehavior` which provides even further fine-grained control over the outside-interaction behavior in builders.

The `interactOutsideBehavior` prop enables the following behaviors:
- `close`: Closes the element immediately.
- `ignore`: Prevents the element from closing and also blocks the parent element from closing in response to an outside interaction.
- `defer-otherwise-close`: Delegates the action to the parent element. If no parent is found, it closes the element.
- `defer-otherwise-ignore`: Delegates the action to the parent element. If no parent is found, nothing is done.

Here are videos showing the behavior of adding different `interactOutsideBehavior` values to a popover that is nested inside a dialog. The popover and the dialog are mounted in separate portals.

### `interactOutsideBehavior: 'close'`

https://github.com/melt-ui/melt-ui/assets/53095479/65a9745f-33b8-48c9-adb9-dc8ca7cf4b9b

### `interactOutsideBehavior: 'ignore'`

https://github.com/melt-ui/melt-ui/assets/53095479/f0306e63-2c8b-4982-975c-b5a9024f0ec7

### `interactOutsideBehavior: 'defer-otherwise-ignore'`

https://github.com/melt-ui/melt-ui/assets/53095479/111a914b-0728-4e91-bb80-c531178baa40

### Popover `interactOutsideBehavior: 'defer-otherwise-close'`

https://github.com/melt-ui/melt-ui/assets/53095479/3d965692-54d3-44f3-803c-5c31881aceaa

### Popover `interactOutsideBehavior: 'defer-otherwise-close'` mounted as a single element

https://github.com/melt-ui/melt-ui/assets/53095479/4e250ee7-23be-4225-85b0-9397a7b18e74


huntabyte:
In the event that a component has this prop set to `"defer"` and it doesn't have a part, will it fall back to closing or remain open?

anatolzak:
@huntabyte remain open

anatolzak:
@huntabyte but I do see what you are getting at. I think that `defer` should late a parent decide its response but if there is only 1 layer with `defer`, I would think it should close. what do you think?

anatolzak:
> @huntabyte but I do see what you are getting at. I think that `defer` should let the parent decide its response but if there is only 1 layer with `defer`, I would think it should close. what do you think?

and if you agree about what I said here, what would you expect to happen if there are 2 elements with `defer`?

huntabyte:
What I'm thinking is, let's say I have a component that I know I always want to close on outside click if it's the highest in the stack, but defer to the parent if one exists. How would I accomplish this?

In my mind right now, it seems like `'defer'` should equal defer to parent but fallback to the default if no parent exists.

anatolzak:
> let's say I have a component that I know I always want to close on outside click if it's the highest in the stack, but defer to the parent if one exists. How would I accomplish this?

With what I suggested you would have to pass a different prop to this component based on where you mount it. 

But I see that it might not be ideal. I do like what you suggested about `defer` defaulting to current element if no parent exists. 

What about if you had 2 elements nested with each other both with `defer`? Would you expect the child to close or the parent to close?

anatolzak:
> let's say I have a component that I know I always want to close on outside click if it's the highest in the stack, but defer to the parent if one exists. How would I accomplish this?

@huntabyte You said that you want to close on outside click, but if the parent has `ignore` and the child has `defer`, would you still want it to close with the example you provided? Currently, with the example I provided, no element would close.

TGlide:
This is a *huge* PR with quite a lot of considerations to take into account. As Melt is going through a refactor for its next iteration, I'll close this. But I'll certainly take inspiration and have this in mind, thank you *so* much for the hard work here!

Issue #886, opened 2024-01-06
🐛Bug: Tooltip with `disableHoverableContent: true` is glitchy


Xyzrr:
**Describe the bug**
When a tooltip has `disableHoverableContent: true`, when your cursor approaches the trigger slowly from the left or the top, the tooltip flashes and disappears.

**To Reproduce**
1. Create a tooltip with these settings (the settings can vary but it's most evident like this):
```
createTooltip({
    positioning: {
      placement: `bottom`
    },
    openDelay: 0,
    closeDelay: 0,
    disableHoverableContent: true
  });
```
2. Move your cursor slowly onto the trigger, from the left edge or the top edge. The trigger can be any element; I've tested on both block and inline, `<a>` and `<div>`.
3. Observe the tooltip flicker, then remain closed even when the cursor is over the trigger.

**Expected behavior**
The tooltip should be open when the mouse is over the trigger, and closed when the mouse is not over the trigger.

**Screenshots**
https://github.com/melt-ui/melt-ui/assets/7032597/fb6209f0-204a-4e32-9a7d-6ec96edc8c73

**Desktop (please complete the following information):**
 - OS: MacOS Ventura 13.5
 - Browser: Tested on both Chrome and Safari, latest

huntabyte:
I see the behavior you're describing, but I'm not able to reproduce.

Xyzrr:
Sure, I made a repro here:

https://www.sveltelab.dev/5d0ac2llmli1ccf?files=.%2Fsrc%2Froutes%2F%2Bpage.svelte

A video of me triggering the flicker in that demo:

https://github.com/melt-ui/melt-ui/assets/7032597/8f44c85b-4966-4d7f-997b-92f54dda1aa7

huntabyte:
I feel like one of your PRs will fix this issue @anatolzak, is that true?

anatolzak:
@huntabyte my PRs don't fix this but I can confirm I am getting the same issue. I assume it's the `pointInPolygon` logic. I will look into this tomorrow

PR #1158, opened 2024-04-06
Fix: tooltip closes while pointer is still inside trigger

anatolzak:
closes https://github.com/melt-ui/melt-ui/issues/886

The reason issue https://github.com/melt-ui/melt-ui/issues/886 occurs is because `clientX` is smaller by 1 than `triggerRect.left`, which causes us to inaccurately "think" that the pointer is outside the trigger which causes the tooltip to close. This is due to sub-pixel rendering and rounding issues by the browser. 

To make sure this doesn't happen, we now keep track of 2 extra variables `isPointerInsideTrigger` and `isPointerInsideContent` that are set on `pointerenter` and `pointerleave` for the trigger and content respectively. And `isPointerInsideTrigger` and `isPointerInsideContent` are taken into account in addition to `pointInPolygon()` when deciding whether the pointer is inside the tooltip area.

### Before

https://github.com/melt-ui/melt-ui/assets/53095479/85d98203-f883-4996-8f8e-d39087c38c94

### After

https://github.com/melt-ui/melt-ui/assets/53095479/0d996b7f-4093-4667-b35d-c933fe7bf697

### DIFF
--- a/.changeset/moody-jokes-peel.md
+++ b/.changeset/moody-jokes-peel.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
+Fixed bug where tooltip would close while the pointer is still inside trigger (closes #886)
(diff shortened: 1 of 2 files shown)


huntabyte (approved):
You're an animal (in a good way)!

TGlide (approved):

Issue #596, opened 2023-10-05
⚙Tech Debt: Unify Link Preview & Tooltip open/close behaviours


TGlide:
Both Link Preview & Tooltips are REALLY similar in the opening and closing behaviours. Tooltip is a bit more advanced currently, with features such as a grace area, and only closing by the same method it was opened, while Link Preview is currently behind.

While Link Preview has some extra features of its own, a lot of the logic could be unified to avoid code duplication and behaviour mismatch.

huntabyte:
If anyone would like to assist with unifying the behaviors of the tooltip and link preview to reduce some duplication that would be awesome!

PR #1159, opened 2024-04-06
chore: refactor isPointerInGraceArea

anatolzak:
We currently have duplicate code in the tooltip and menu builders for checking if a pointer is inside a polygon of points. This PR aims to refactor this.

### DIFF
--- a/src/lib/builders/context-menu/create.ts
+++ b/src/lib/builders/context-menu/create.ts
@@ -20,6 +20,7 @@ import {
 	withGet,
 	type WithGet,
 	portalAttr,
+	type Point,
 } from '$lib/internal/helpers/index.js';
 import type { MeltActionReturn } from '$lib/internal/types.js';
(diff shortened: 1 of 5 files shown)


huntabyte:
Hey @anatolzak! Awesome job! Do you think we could include the `LinkPreview` in this as well to close out this issue? https://github.com/melt-ui/melt-ui/issues/596

anatolzak:
@huntabyte will do!

anatolzak:
@huntabyte I think we should not include the proposed changes in https://github.com/melt-ui/melt-ui/issues/596 in this PR because from what I understand @TGlide suggested to combine a lot of the logic of the tooltip and link preview which includes quite a bit more than just the grace area. I do want to create a separate PR implementing what Thomas suggested. So if you don't disagree, I think this PR should be merged independently of https://github.com/melt-ui/melt-ui/issues/596

PR #1162, opened 2024-04-07
chore: use `executeCallbacks` for cleaning up multiple functions returned from an `effect`

anatolzak:
In many builder's effects, there are multiple cleanup functions that should be executed when the effect reruns and when the effect is destroyed. Instead, we can just return all the cleanup functions from an effect inside `executeCallbacks` and avoid the need to call the cleanup functions manually at the start of each effect and in the action's `destroy` method.

However, some builders call `tick()` before assigning the cleanup functions, which creates a promise. In order to allow returning cleanup functions that are created after a tick, PR https://github.com/melt-ui/melt-ui/pull/1140 should be merged first. That PR introduced the `runAfterTick` prop to the `effect` function. Thanks to that PR, we will be able to run the entire effect callback after a tick, which allowed this PR to return all the cleanup functions in one go after a tick with `executeCallbacks`.


TGlide:
Will close for now, as effect may be refactored later on

Issue #1168, opened 2024-04-08
Text selection overflows floating element


anatolzak:
### Describe the bug

Currently, when starting text selection from within a floating element and dragging the pointer outside, the text outside the floating element gets selected. It would be nice to have a prop like `preventTextSelectionOverflow` that controls whether text-selection overflow is allowed.

https://github.com/melt-ui/melt-ui/assets/53095479/b3ec235e-8106-4715-aa1d-bd9db2d2e32e

Currently, the link preview builder is the only builder that implements this feature.

https://github.com/melt-ui/melt-ui/assets/53095479/9bac1dd6-a591-40a1-a2a6-ad8ad964fb44

### Reproduction

-

### System Info

```bash
System:
    OS: Linux 5.0 undefined
    CPU: (6) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.3 - /usr/local/bin/pnpm
  npmPackages:
    @melt-ui/svelte: latest => 0.76.3 
    @sveltejs/kit: ^2.0.0 => 2.5.5 
    svelte: ^4.1.2 => 4.2.12
```

### Severity

annoyance

PR #1170, opened 2024-04-08
Feat: add new `preventTextSelectionOverflow` prop

anatolzak:
closes https://github.com/melt-ui/melt-ui/issues/1168

Currently, when initiating text selection from within a floating element and dragging the pointer outside, the text outside the floating element gets selected.

This PR introduces a new prop `preventTextSelectionOverflow` that controls whether text-selection overflow is allowed. By default `preventTextSelectionOverflow` will be `true`.

The affected builders are:
- listbox (combobox, select)
- menu (context menu, dropdown menu, menubar)
- popover

Link preview already implemented preventing text-selection overflow, so now the link preview builder gets this behavior from `usePopper` like the other builders mentioned above.

Tooltip, however, is not added in this PR because the tooltip closes when the pointer leaves the grace zone. The link preview has logic to prevent closing when text selection occurred. In a later PR, I will merge a lot of the behaviors of the tooltip and link preview, at which point tooltip will inherit the prevention of text-selection overflow and not close itself after the pointer leaves after text selection.

Also, the dialog is not added in this PR because the overlay already takes care of not allowing user selection because the overlay spans the entire viewport.

I haven't added tests yet because I am not sure how you would test this behavior without bending backwards

Here are videos showing all the affected builders before and after.

### Context menu

https://github.com/melt-ui/melt-ui/assets/53095479/60c62f3d-550f-4c96-9015-5d34a0bfa775

### Dropdown menu

https://github.com/melt-ui/melt-ui/assets/53095479/fb1c0930-6069-4a1e-9f09-6dba1060ad1b

### Link preview - same behavior as before

https://github.com/melt-ui/melt-ui/assets/53095479/48f201d1-c345-46ed-a1e3-dabbc9e3a4d6

### Combobox

https://github.com/melt-ui/melt-ui/assets/53095479/950f9d22-df83-4f4e-9312-072d6a6a3a7e

### Select

https://github.com/melt-ui/melt-ui/assets/53095479/12984228-2ba3-4cf1-b266-9ac7cb972ff2

### Menubar

https://github.com/melt-ui/melt-ui/assets/53095479/2b913d67-b5ea-4c7e-beb5-11017dc0a27a

### Popover

https://github.com/melt-ui/melt-ui/assets/53095479/7120e666-aa09-4c16-9c5d-5e53e32ec392


huntabyte:
Part of me wonders when this would be desired if you're using the floating element in a "modal" fashion (close on outside click, etc.). I suppose since we don't have a source of truth of whether the box is acting as a modal or not having another prop makes sense.

I do feel like the default for this should align with the default of the other "close on escape" / "close on outside click" where if we default the builder to act as a modal, this should default to `true`

anatolzak:
> Part of me wonders when this would be desired if you're using the floating element in a "modal" fashion (close on outside click, etc.). I suppose since we don't have a source of truth of whether the box is acting as a modal or not having another prop makes sense.
> 
> I do feel like the default for this should align with the default of the other "close on escape" / "close on outside click" where if we default the builder to act as a modal, this should default to `true`

For anyone reading, Hunter thought that I was proposing setting `preventTextSelectionOverflow ` as `false` by default, which is not the case. So disregard his comment :)

TGlide (approved):

PR #1180, opened 2024-04-20
docs(combobox): change item to option

JoBurgard:
In the docs for combobox it mentions 'item' in the API Reference area. It seems that it changed to 'option' at some point but was forgotten to be updated in the docs accordingly. This changes 'item' to 'option'.


huntabyte (approved):
Thank you!

Issue #1108, opened 2024-03-20
Action cleanup functions don't run on builder children when unmounting builder parent provided `forceVisible: false`


anatolzak:
### Describe the bug

This issue relates to the fact that in svelte, action cleanup functions don't run when you remove an element from the dom using js `el.remove()` as opposed to an element being removed by svelte either during component unmount or unmounting an element inside a condition block.

This has particular significance in melt when using builders with `forceVisible: false` where melt is responsible for "unmounting" elements and can only do so using vanilla javascript.

This has multiple consequences. All the points below assume using `forceVisible: false`

### 1. Visual
If we have a popover inside a popover using separate portals, and we click on a button inside the nested popover that closes the parent popover, the second popover stays open.

https://github.com/melt-ui/melt-ui/assets/53095479/096c709d-d59a-4a36-9fe5-f4fd6be11716

This specific visual issue only happens when using sibling portals. However the actual reason this occurs is because when we click on the button to close the parent popover, internally we "unmount" the first portal using `el.remove()`, however that doesn't cause all children to call their action's cleanup functions, which is why the second portal doesn't get removed.

The reason I said that this issue doesn't occur when using only 1 portal is because in that case, both popovers are inside the same portal that gets removed. But this doesn't change the fact that the cleanup functions of all builder children don't run.

### 2. Functional
We are going to have potential memory leaks. Imagine we are listening for events on the `body`. Those event handlers will continue running even though the builder was actually removed from the dom.

And just to make sure this is understood, **portals don't have to be involved for this issue to occur. In general when unmounting any parent builder, all its builder children are not going to cleanup their svelte actions.**

----

I am not sure what we can do about this issue... 

The good news is that when using `forceVisible: true`, this whole issue is gone because svelte is responsible for removing elements from the DOM in which case, svelte will take care of calling cleanup functions returned from actions.

### Reproduction

https://stackblitz.com/edit/github-nxujbd?file=src%2Froutes%2F%2Bpage.svelte,src%2Flib%2Fcomponents%2FPopover.svelte

### System Info

```bash
System:
    OS: Linux 5.0 undefined
    CPU: (6) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
[...]

PR #1183, opened 2024-04-21
docs: fix context menu submenu not reopening after escape due to #1108

anatolzak:
When opening context menu, and opening a submenu, when pressing escape, the whole menu closes as expected. However, when reopening the context menu, and hovering over the submenu trigger, the submenu doesn't open. This issue only happens when using `forceVisible: false`, due to action cleanup functions not running on children as documented in https://github.com/melt-ui/melt-ui/issues/1108

https://github.com/melt-ui/melt-ui/assets/53095479/c47c45a7-1f63-4e5f-b0b8-f15ad070ebf1


TGlide:
Closing since it no longer happens

PR #1188, opened 2024-04-30
fix(pagination): do not add page-0 to pagesToShow

Spioune:
Hi,

This is my first contribution to melt-ui.
I am submitting this PR to fix a bug on the Pagination component when there is no element.

image

```js
const {
  elements: { root, pageTrigger, prevButton, nextButton },
  states: { pages, range }
} = createPagination({
  count: 0,
  perPage: 20,
  defaultPage: 1,
  siblingCount: 1
});
$: console.log($pages); // [{"type": "page","value": 0,"key": "page-0"}, {"type": "page", "value": 1,"key": "page-1" }]
```

As you can see, when count = 0, the pagination component creates "page-0" and "page-1" but I think it should only create "page-1"
When count = 1, the component behave correctly and renders only "page-1"

### DIFF
--- a/src/lib/builders/pagination/helpers.ts
+++ b/src/lib/builders/pagination/helpers.ts
@@ -6,9 +6,13 @@ export function getPageItems({
 	siblingCount = 1,
 }: GetPageItemsArgs): Array<PageItem> {
 	const pageItems: Array<PageItem> = [];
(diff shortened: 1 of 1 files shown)


TGlide:
Thank you! Fixed this in a similar way :)

PR #1203, opened 2024-05-15
feat: allow passing Floating UI's `FlipOptions` to `positioning.flip` property 

msigwart:
### Motivation 
Currently when using builders like `createDropdownMenu`, `positioning.flip` is a boolean property. Meaning flipping of the menu in case of overflow can either be turned on or off.

**With `flip: false`:**

https://github.com/melt-ui/melt-ui/assets/9678787/33425aae-505b-4204-9b0a-df110dfb7a30

**With `flip: true`:**

https://github.com/melt-ui/melt-ui/assets/9678787/692c567d-96d5-41cb-b04a-839d3ef8e109

Notice how with `flip: true`  the menu is flipped on both axes in case of overflow.

It would be great if it was possible to customize the flip behaviour further than just turning it on or off.

### Changes
This PR extends the flip options to optionally take an object with [flip customization options offered by Floating UI](https://floating-ui.com/docs/flip#options). This allows more fine-grained control over the flip behaviour of menus. 

Screenshot 2024-05-15 at 15 57 01@2x

For example, it now becomes possible to turn flipping _on_ for the main axis but _off_ for the cross axis. 

```svelte
const {
    elements: { subMenu, subTrigger },
    states: { subOpen },
} = createSubmenu({
    positioning: {
	placement: 'right-start',
        flip: {
	    mainAxis: true,
	    crossAxis: false,
        },
    },
});
```

https://github.com/melt-ui/melt-ui/assets/9678787/d7cedd11-1a4e-4cbf-a9a9-6008e9a7d8a7

Notice, how in the video the submenu is opened on the left instead of the right (because of the right-hand overflow on the "main" axis) but only shifted upwards instead of flipped upward (because of the bottom overflow on the "cross" axis).

### Testing
I've temporarily adjusted [the main preview of the dropdown menu component](http://localhost:5173/docs/builders/dropdown-menu#menu) to showcase the new behaviour. 

You can adjust the browser window so that the opening the dropdown menu would cause an overflow and see the new behaviour in action. 

---
We've recently started using Melt UI and really enjoy it so far so curious to hear your feedback on this PR!

### DIFF
--- a/src/lib/internal/actions/floating/action.ts
+++ b/src/lib/internal/actions/floating/action.ts
@@ -13,7 +13,7 @@ import {
 	autoUpdate,
 } from '@floating-ui/dom';
 import type { FloatingConfig } from './types.js';
-import { isHTMLElement, noop } from '$lib/internal/helpers/index.js';
+import { isHTMLElement, isObject, noop } from '$lib/internal/helpers/index.js';
 import type { Placement, VirtualElement, FlipOptions } from '@floating-ui/core';
 
 const defaultConfig = {
@@ -38,8 +38,6 @@ const ALIGN_OPTIONS = ['start', 'center', 'end'] as const;
 type Side = (typeof SIDE_OPTIONS)[number];
 type Align = (typeof ALIGN_OPTIONS)[number];
 
-const isFlipOptions = (value: any): value is FlipOptions => typeof value === 'object';
-
 export function useFloating(
 	reference: HTMLElement | VirtualElement | undefined,
 	floating: HTMLElement | undefined,
@@ -60,7 +58,7 @@ export function useFloating(
 			flip({
 				boundary: options.boundary,
 				padding: options.overflowPadding,
-				...(isFlipOptions(options.flip) && options.flip),
+				...(isObject(options.flip) && options.flip),
 			})
 		);
 	}
--- a/src/lib/internal/actions/floating/action.ts
+++ b/src/lib/internal/actions/floating/action.ts
@@ -14,7 +14,7 @@ import {
 } from '@floating-ui/dom';
 import type { FloatingConfig } from './types.js';
 import { isHTMLElement, noop } from '$lib/internal/helpers/index.js';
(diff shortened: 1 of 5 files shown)


TGlide:
Hey, this is great stuff! Thank you!

src/lib/internal/actions/floating/action.ts
```diff
@@ -38,6 +38,8 @@ const ALIGN_OPTIONS = ['start', 'center', 'end'] as const;
 type Side = (typeof SIDE_OPTIONS)[number];
 type Align = (typeof ALIGN_OPTIONS)[number];
 
+const isFlipOptions = (value: any): value is FlipOptions => typeof value === 'object';
```
TGlide, line 41: This type-guard is misleading, and can lead to errors in the future. Wouldn't an `isObject` type guard suffice? IIRC we have one already.

src/docs/components/preview-wrapper.svelte
```diff
@@ -34,7 +34,7 @@
 </script>
 
 <div class={cn(previewVariants({ variant, size }))} data-variant={variant} {id}>
-	<div class={cn('z-10 mx-auto inline-block px-4')}>
+	<div class={cn('z-10 mb-4 ml-auto inline-block self-end px-4')}>
```
TGlide, line 37: Please change this before merging :)

TGlide:
Also, it seems like I can't edit the PR directly. Could you add a changeset with `npx changeset`?

src/lib/internal/actions/floating/action.ts
msigwart: Good idea, replaced with `isObject`.

src/docs/components/preview-wrapper.svelte
msigwart: Done!

msigwart:
Thanks for the review, I've added a changeset and I think otherwise the PR should be ready.

TGlide (approved):
Thank you!

PR #1210, opened 2024-05-27
feat: add rootMargin to TableOfContents builder

bolstycjw:
## Motivation
TableOfContents interaction observer allows passing in scroll offset, which can be used to account for fixed/sticky headers when scrolling, but there is no way to cater for that in interaction observer.

## Changes
This PR adds a rootMargin argument for TableOfContents builder, which is then passed to the InteractionObserver constructor, thus catering for viewport customisation and use cases like fixed headers/footers.

### DIFF
--- a/.changeset/rude-pugs-tease.md
+++ b/.changeset/rude-pugs-tease.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": minor
+---
+
+feat: add rootMargin to TableOfContents builder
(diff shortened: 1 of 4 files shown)


TGlide:
Awesome, would you mind adding info to the docs as well?

bolstycjw:
have updated the docs :)

TGlide (approved):
Awesome, thank you!

PR #1212, opened 2024-05-28
feat(slider): add `onValueCommitted` callback

mikededo:
### Motivation

While developing a filter-like interface, a range slider is used. The state of
the filters is kept in the URL, meaning that on every change for the slider, the
URL would be updated. Since the `onValueChange` function is called a lot of
times, the browser would end up being blocked for a while, as the URL is updated
on every change.

### Solution

Add a new `onValueCommitted` callback which is called whenever the user has
already finished changing the value.


TGlide (approved):
Thank you so much, this is great!

PR #1217, opened 2024-05-31
fix: show toasts with `closeDelay: undefined`

### LINKED ISSUE #1176: `Toast` doesn't show up at all when using `closeDelay: undefined`
### Describe the bug

`closeDelay` config options are specified as `closeDelay?: number | undefined` and I would expect `undefined` to make it fallback to the default defined by melt-ui, which is currently 5s.

Unfortunately this is not the case and the toast doesn't show up at all, which is related to https://github.com/melt-ui/melt-ui/blob/3c5277a77df9fa197afed262535d2719ce036500/src/lib/builders/toast/create.ts#L47-L52

### Reproduction

https://stackblitz.com/edit/github-jjlpgz?file=src%2Flib%2Fcomponents%2FToast.svelte

### Logs

_No response_

### System Info

```bash
System:
    OS: Linux 6.5 Ubuntu 23.10 23.10 (Mantic Minotaur)
    CPU: (16) x64 AMD Ryzen 7 5800H with Radeon Graphics
    Memory: 11.44 GB / 30.71 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 20.11.0 - ~/.volta/tools/image/node/20.11.0/bin/node
    Yarn: 1.22.22 - ~/.volta/tools/image/yarn/1.22.22/bin/yarn
    npm: 10.2.4 - ~/.volta/tools/image/node/20.11.0/bin/npm
    pnpm: 8.14.1 - ~/.volta/tools/image/pnpm/8.14.1/bin/pnpm
    bun: 1.1.3 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 123.1.64.122
    Chrome: 123.0.6312.122
  npmPackages:
    @melt-ui/svelte: ^0.76.3 => 0.76.3 
    @sveltejs/kit: ^2.5.5 => 2.5.5 
    svelte: ^4.2.9 => 4.2.12
```

### Severity

annoyance

DerTimonius:
## Description

This PR fixes a small issue with toasts not showing up when passing `closeDelay: undefined`. The expected behaviour would be that the toasts are shown for the default 5 seconds.

I tried to simply move the spreading of the props in the creation of the `propsWithDefaults` variable, which does work - kind of. The problem then would be that `closeDelay` could not be changed then...

So I checked if the closeDelay is undefined and use the default instead. I'm not sure if the solution is what you had in mind and if you wanted a better or more verbose approach, but this ensures the correct bahaviour.

Closes #1176

### DIFF
--- a/src/lib/builders/toast/create.ts
+++ b/src/lib/builders/toast/create.ts
@@ -33,9 +33,9 @@ export function createToaster<T = object>(props?: CreateToasterProps) {
 
 	const addToast = (props: AddToastProps<T>) => {
 		const propsWithDefaults = {
-			...props,
 			closeDelay: closeDelay.get(),
 			type: type.get(),
+			...props,
 		} satisfies AddToastProps<T>;
 
 		const ids = {
@@ -49,7 +49,7 @@ export function createToaster<T = object>(props?: CreateToasterProps) {
 				? null
 				: window.setTimeout(() => {
 						removeToast(ids.content);
-				  }, propsWithDefaults.closeDelay);
+				  }, propsWithDefaults.closeDelay ?? defaults.closeDelay);
 
 		const getPercentage = () => {
 			const { createdAt, pauseDuration, closeDelay, pausedAt } = toast;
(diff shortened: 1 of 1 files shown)


TGlide:
This would be an error with everything that undefined is passed, throughout the entirety of Melt. A helper needs to be created to deal with this, and used in every builder

PR #1218, opened 2024-06-01
fix(slider): incorrect `onValueCommitted` calls

mikededo:
### Motivation

This is a small bug/edge case that was introduced with the addition of
`onValueCommitted` in #1212. I basically discovered the problem when I was using
a `select` within a `Slider`, that the `select` component would not open since
the state was being updated due to incorrect `onValueCommitted` calls.
With this fix, we ensure that `onValueCommitted` is only called when the
`Slider` was actually active, meaning that the `onmousedown` was called in one
of the thumbs of the `Slider`.

### DIFF
--- a/.changeset/fifty-suits-reflect.md
+++ b/.changeset/fifty-suits-reflect.md
@@ -1,0 +1,5 @@
+---
+'@melt-ui/svelte': patch
+---
+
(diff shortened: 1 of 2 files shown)


TGlide (approved):
Thank you!

PR #1235, opened 2024-06-20
Resolves #687 Add skip to content button

### LINKED ISSUE #687: 📁Docs: Add 'Skip to Main Content' Button

evelinabe:
Skip to content button added. It does not work so well on the first page - it is the first item in the navigation order, but the focus is set to the popover button. Let me know what you think, and if there's something in the code that needs updating.

Resolves issue [#687](https://github.com/melt-ui/melt-ui/issues/687)

### DIFF
--- a/src/routes/(landing-ui)/popover.svelte
+++ b/src/routes/(landing-ui)/popover.svelte
@@ -25,8 +25,13 @@
 
 	export let contentClass = '';
 
+	let popoverButton: HTMLButtonElement;
+
 	onMount(() => {
 		open.set(true);
+		setTimeout(() => {
+			popoverButton.blur();
+		}, 100);
 	});
 </script>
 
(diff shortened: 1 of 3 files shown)


TGlide:
Hey, thank you for the PR!

It seems the button isn't visible when focusing it. I can see in the console that it is focused, but nonetheless there is no visual indication. At least in the cloudflare preview!

evelinabe:
> It seems the button isn't visible when focusing it. I can see in the console that it is focused, but nonetheless there is no visual indication. At least in the cloudflare preview!

I noticed that, too, took a look at the preview. It needed a higher z-index, I already pushed that change :)

TGlide:
Hmm, the new preview still fails. I don't think z-1000 is a valid tailwind class, you may need to do `z-[1000]`

evelinabe:
Strange, it works on my machine 🤔 
Giving it another go 🤞

TGlide:
Awesome, it works now! Thank you so much for the contribution, and for putting up with the back and forth 😄

evelinabe:
No problem, happy to help!

Issue #1213, opened 2024-05-30
Calendar with a locale with Monday as firstDay skips first Monday (like 1 July)


cbasje:
### Describe the bug

~When a Calendar is created with a locale that has Mondays as firstDay, such as en-GB~, it skips the first day of the month when it is a Monday (when viewing the previous month). You only see this when `fixedWeek: true` is set as true, otherwise the last row is not rendered, of course.

It is difficult to explain well but I have a reproduction: [stackblitz.com/edit/github-qbikvw](https://stackblitz.com/edit/github-qbikvw?file=src%2Flib%2Fcomponents%2FCalendar.svelte). If you view June 2024, for example, you will see that the bottom row skips 1 July and shows 2-8 July (even though it should only go until 7 July). Or you can go to August 2025 where you see that the bottom row again skips 1 September.

Edit: this is still a problem with locales that have Sundays as firstDay (see August 2024 with the locale set to 'en-US')

### Reproduction

[https://stackblitz.com/edit/github-qbikvw?file=src%2Flib%2Fcomponents%2FCalendar.svelte](https://stackblitz.com/edit/github-qbikvw?file=src%2Flib%2Fcomponents%2FCalendar.svelte)

### System Info

```bash
System:
    OS: macOS 14.4.1
    CPU: (10) arm64 Apple M2 Pro
    Memory: 120.80 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.0.0 - ~/.nvm/versions/node/v21.0.0/bin/node
    npm: 10.2.0 - ~/.nvm/versions/node/v21.0.0/bin/npm
    bun: 1.1.7 - /opt/homebrew/bin/bun
  Browsers:
    Safari: 17.4.1
  npmPackages:
    @melt-ui/svelte: ^0.80.0 => 0.80.0
    @sveltejs/kit: ^2.5.2 => 2.5.2
    svelte: ^4.2.12 => 4.2.12
```

### Severity

blocking an upgrade

huntabyte:
Closed by #1236

PR #1236, opened 2024-06-23
fix(Calendar): sometimes first day of next month is missing

xollaborator:
Solves https://github.com/melt-ui/melt-ui/issues/1213.

This was solved by me in https://github.com/radix-vue/radix-vue/pull/1029 but @epr3 suggest you might also need this fix. 

Root cause: when 6th week will start in new month then `startFrom` will be `undefined`. Currently `if (!startFrom)` was using first day of next month, but then in creation of extraDaysArray all dates were incremented which resulted in first day of new month being missed.

I am not familliar with Svelte so I do not have any test case for this. But you might want to cross check with how it **might be** tested in radix-vue: https://github.com/radix-vue/radix-vue/pull/1029/files#diff-a067dd33cc3ddceb2d881b1d1984509071d713079564a97d9ea182084284e2f8

At least using August 2024 instead of January 1980 might help if you are into this, but my conclusion is it most probably will not be worth it, especially from performance perspective.

### DIFF
--- a/.changeset/curly-mails-jump.md
+++ b/.changeset/curly-mails-jump.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": patch
+---
+
+fix(Calendar): sometimes first day of next month is missing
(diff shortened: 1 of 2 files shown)


TGlide:
Thank you!!

PR #1237, opened 2024-06-24
feat(radio group): add `name` prop to radio group builder

ksmoore17:
hidden input element on radio group builder always had undefined `name` attribute

added a test and updated docs


TGlide (approved):
Thank you!

PR #1239, opened 2024-06-26
Feat: add a pushState function as parameter to the table of content builder

Manuel-DaSilva:
Hello! This is a first attempt to provide the pushState function from SvelteKit to the Table of Contents builder. Using history.pushState not only throws a warning but also breaks some of SvelteKit's $page store state. I'd love to hear your comments and explore other possible implementation approaches.

### DIFF
--- a/.changeset/purple-eggs-flow.md
+++ b/.changeset/purple-eggs-flow.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": minor
+---
+
+feat: add a pushState prop to createTableOfContents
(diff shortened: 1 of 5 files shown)


TGlide:
This is a great idea! Would you mind adding some documentation for it?

Manuel-DaSilva:
Awesome! Sure I will add the documentation!

Manuel-DaSilva:
I've added some documentation for the pushStateFn. However, I'm unsure whether we should emphasize SvelteKit's function or use it as an example in the builder preview.

This expected to work with the 'svelte-routing' library too. However, I haven't tested it yet.

TGlide:
@Manuel-DaSilva I'd say we can emphasize and use it as an example.

Manuel-DaSilva:
@TGlide Awesome, in that case the documentation that I added has that emphasis and the example included!

Manuel-DaSilva:
Hello @TGlide! Anything more I can do for this PR? Did you find the docs ok?

TGlide:
@Manuel-DaSilva sorry for the wait! this is great, merging 🎉 Thanks again!

PR #1258, opened 2024-08-03
feat: Add an option to ListBox to prevent close on input click

codercms:
It makes possible to create ListBox like on Google search page (if you click on input listbox won't be closed)


TGlide:

PR #1270, opened 2024-09-24
feat(listbox): add `highlightOnHover` prop to listboxes

ksmoore17:
Adds a prop to select and combobox that allows users to disable highlighting the selected (or first selected) option by default when opening the listbox menu. Useful as otherwise the listbox menu will take keyboard inputs starting from the highlighted instead of from the top of the list.


TGlide:

PR #1271, opened 2024-09-24
feat(popover): add `group` prop

ksmoore17:
Same idea/implementation as the tooltip `group` prop


TGlide:

PR #1278, opened 2024-10-23
feat: expose active state on slider

bryanmylee:
It is sometimes desirable to customize the behaviour of the slider based on whether it is active / being interacted with.

For example, I have a slider that snaps to certain predefined values when the user drags the slider, but can be set with a text input to non-snapped values on my project.

Due to `steps`, the slider calls `value.update` whenever `$value` changes, even if the change comes from an external interaction. To prevent this, I've patched `@melt-ui/svelte` to expose the `isActive` store as `active`, enabling the following behaviour:

```ts
export let valueStore: Writable<number>;

const {
	elements: { root, range, thumbs },
	states: { active }
} = createSlider({
	value: {
		...derived(valueStore, ($value) => [$value]),
		set(newValue: number[]) {
			if (!$active) return;
			valueStore.set(newValue[0]);
		},
		update(updater: Updater<number>) {
			if (!$active) return;
			valueStore.update((prev) => updater([prev])[0]);
		}
	},
	min,
	max
});
```

I use it in my [project](https://github.com/bryanmylee/natural-light/blob/main/src/routes/TemperatureSlider.svelte) via a patch.

### DIFF
--- a/.changeset/flat-books-jam.md
+++ b/.changeset/flat-books-jam.md
@@ -1,0 +1,5 @@
+---
+"@melt-ui/svelte": minor
+---
+
+feat: expose active state on slider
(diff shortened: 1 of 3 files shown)


TGlide (approved):
Thank you!

PR #1302, opened 2025-03-06
Fix main/module/exports for "."

jycouet:
Using `vitest 3`, I get this error when I run my tests: 

```
Error: Failed to resolve entry for package "@melt-ui/svelte". The package may have incorrect main/module/exports specified in its package.json: No known conditions for "." specifier in "@melt-ui/svelte" package
```

Adding this default export is fixing the issue. _(I have it locally in my pnpm paches atm)_

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
 	"exports": {
 		".": {
 			"types": "./dist/index.d.ts",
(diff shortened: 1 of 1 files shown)


TGlide (approved):
Thank you!

Issue #1307, opened 2025-03-26
DateField does not work in Shadow Dom


eugenefm:
### Describe the bug

The Date Field component does not seem to work inside a shadow dom. 

It just defaults to the current date when you attempt to enter a number and is not editable.

At an initial glance, I believe it is because of the underlying date utilities calling `document.getElementById`. This component builder needs to accept a rootElement prop like some of the builders and the date utilities need to be updated to use the existing `getElementById` utility with the rootElement passed in.

### Reproduction

Here's a link to a stackblitz that reproduces the error:
https://stackblitz.com/edit/github-8xa9to-y5azhbdw?file=src%2Froutes%2F%2Blayout.svelte

https://github.com/user-attachments/assets/2d63f006-3b42-4ddc-8c5d-f1f346627851

### Logs

```bash

```

### System Info

```bash
System:
    OS: Linux 6.10 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (12) arm64 unknown
    Memory: 2.68 GB / 7.65 GB
    Container: Yes
    Shell: 5.1.4 - /bin/bash
Binaries:
    Node: 18.18.0 - /opt/nodejs/bin/node
    Yarn: 1.22.22 - /opt/nodejs/bin/yarn
    npm: 9.8.1 - /opt/nodejs/bin/npm
npmPackages:
    @melt-ui/svelte: ^0.83.0 => 0.83.0
    svelte: ^4.2.8 => 4.2.12
```

### Severity

annoyance

moiz-p:
This is affecting us too, pretty badly actually.

moiz-p:
The DatePicker seems to have the same issue?

PR #1312, opened 2025-04-15
fix: added support to define root for date-field component

peopledrivemecrazy:
Essentially allows date field component to work inside shadow dom.

I followed #1276 and #1279, no tests were added, but I got the example on the date field docs, I tried to get the styles align as close as possible.

This PR addresses #1307.


moiz-p:
Thanks for making this PR @peopledrivemecrazy. Any estimate on when it's merged? Thanks.

